PR #5, opened 2015-02-04
[feature] file system provider

icereval:


waterbutler/providers/filesystem/provider.py
```diff
@@ -0,0 +1,155 @@
...
+        self.folder = self.settings['folder']
+        os.makedirs(self.folder, exist_ok=True)
+
+    @asyncio.coroutine
+    def intra_copy(self, dest_provider, source_options, dest_options):
+        source_path = FileSystemPath(self.folder, source_options['path'])
+        dest_path = FileSystemPath(self.folder, dest_options['path'])
+
+        shutil.copy(source_path.full_path, dest_path.full_path)
+        return self.metadata(str(dest_path))
```
jmcarp: `metadata` is a coroutine, so should this be `return (yield from self.metadata(str(dest_path)))`?

waterbutler/providers/filesystem/provider.py
```diff
@@ -0,0 +1,155 @@
...
+
+    @asyncio.coroutine
+    def intra_move(self, dest_provider, source_options, dest_options):
+        source_path = FileSystemPath(self.folder, source_options['path'])
+        dest_path = FileSystemPath(self.folder, dest_options['path'])
+
+        shutil.move(source_path.full_path, dest_path.full_path)
+        return self.metadata(str(dest_path))
```
jmcarp: Same question as above.

waterbutler/providers/filesystem/provider.py
```diff
@@ -0,0 +1,155 @@
...
+
+        self._folder = folder
+
+        full_path = os.path.join(folder, path.lstrip('/'))
+        self._full_path = self._format_path(full_path)
+
+    def __repr__(self):
+        return "{}({!r}, {!r})".format(self.__class__.__name__, self._folder, self._orig_path)
+
+    @property
```
jmcarp: Do we need this to be a property? What about setting `self.folder = folder` in the constructor and deleting this property method?

waterbutler/providers/filesystem/provider.py
```diff
@@ -0,0 +1,155 @@
...
+        path = FileSystemPath(self.folder, path)
+
+        try:
+            yield from self.metadata(str(path))
+        except exceptions.MetadataError:
+            created = True
+        else:
+            created = False
+
+        with open(path.full_path, 'wb') as file_pointer:
```
jmcarp: Does this support nested paths? Do we care?

waterbutler/providers/filesystem/provider.py
icereval: done

waterbutler/providers/filesystem/provider.py
icereval: done

waterbutler/providers/filesystem/provider.py
icereval: Yes, added specific nested upload tests.

waterbutler/providers/filesystem/provider.py
icereval: It does not need to be exposed, its underscores just to indicate its private.

PR #7, opened 2015-02-05
Feature/revisions

chrisseto:


waterbutler/providers/osfstorage/provider.py
```diff
@@ -82,7 +85,7 @@ def download(self, **kwargs):
         provider = self.make_provider(data['settings'])
         data['data']['path'] = '/' + data['data']['path']
         download_kwargs = {}
-        download_kwargs.update(kwargs)
+        # download_kwargs.update(kwargs)
```
icereval: This was changed to allow for accept_url, are you sure this is still handled here?

https://github.com/CenterForOpenScience/waterbutler/commit/cd0bb17b41abe2e3f4a07c29b8bc55a6877d3563#diff-927878a162eb1a1110dfc88d4cf2373dR85

waterbutler/providers/github/provider.py
```diff
@@ -55,10 +55,12 @@ def build_repo_url(self, *segments, **query):
 
     @asyncio.coroutine
     def download(self, path, ref=None, **kwargs):
-        if self._is_sha(ref):
+        file_sha = kwargs.get('fileSha')
```
icereval: add a comment why camel case fileSha is not in the signature.

PR #9, opened 2015-02-10
Box Provider

mfraezz:
Purpose
=======
Add Box to the supported providers for waterbutler
Accompanies: https://github.com/CenterForOpenScience/osf.io/pull/1758

Changes
=======
* Box support added
* FormData added to utils (from @chrisseto )

Side Effects
=========
* Theoretically possible for a user to intercept outgoing requests, edit the id, and gain access to an authorizer's files that aren't in the shared directory

### DIFF
--- a/waterbutler/providers/box/metadata.py
+++ b/waterbutler/providers/box/metadata.py
@@ -48,6 +48,12 @@ class BoxFileMetadata(BaseBoxMetadata, metadata.BaseFileMetadata):
     def content_type(self):
         return None
 
+    @property
+    def extra(self):
+        return {
+            'etag': self.raw.get('etag')
+        }
+
 
 class BoxRevision(metadata.BaseFileRevisionMetadata):
 
@@ -77,5 +83,7 @@ class BoxRevision(metadata.BaseFileRevisionMetadata):
             return self.raw.get('modified')
 
     @property
-    def revision(self):
-        return self.raw['etag']
+    def extra(self):
+        return {
+            'etag': self.raw.get('etag')
+        }
--- a/waterbutler/providers/box/metadata.py
+++ b/waterbutler/providers/box/metadata.py
@@ -85,5 +85,5 @@ class BoxRevision(metadata.BaseFileRevisionMetadata):
     @property
     def extra(self):
         return {
-            'etag': self.raw.get('etag')
+            'etag': self.raw['etag']
         }
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -30,7 +30,7 @@ class BoxProvider(provider.BaseProvider):
     def __init__(self, auth, credentials, settings):
         super().__init__(auth, credentials, settings)
         self.token = self.credentials['token']
-        self.folder = self.settings['folder_id']
+        self.folder = self.settings['folder']
 
     @property
     def default_headers(self):
@@ -56,7 +56,10 @@ class BoxProvider(provider.BaseProvider):
 
     @asyncio.coroutine
     def upload(self, stream, path, **kwargs):
-        path = BoxPath('/{}{}'.format(self.folder.lstrip('/'), path))
+        if path.split('/')[0] == '':
+            path = '/{}{}'.format(self.folder, path)
+        import ipdb; ipdb.set_trace()
+        path = BoxPath(path)
         try:
             meta = yield from self.metadata(str(path))
         except exceptions.MetadataError:
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -30,7 +30,7 @@ class BoxProvider(provider.BaseProvider):
     def __init__(self, auth, credentials, settings):
         super().__init__(auth, credentials, settings)
         self.token = self.credentials['token']
-        self.folder = self.settings['folder']
+        self.folder = self.settings['folder_id']
 
     @property
     def default_headers(self):


waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,123 @@
...
+        return self.raw['name']
+
+    @property
+    def path(self):
+        return '/{0}/{1}'.format(self.raw['id'], self.raw['name'])
+
+    @property
+    def size(self):
+        try:
+            return self.raw['size']
```
chrisseto:
```python
return self.raw.get('size')
```

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,123 @@
...
+        except KeyError:
+            return None
+
+    @property
+    def modified(self):
+        try:
+            return self.raw['modified_at']
```
chrisseto: See above

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,123 @@
...
+        except KeyError:
+            return None 
+
+    @property
+    def parent(self):
+        try:
+            return self.raw['parent']['id']
```
chrisseto: Shouldn't there always be a parent?

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,123 @@
+import os
```
chrisseto: `os` is unused here.

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,123 @@
...
+
+from waterbutler.core import metadata
+
+
+class BaseBoxMetadata(metadata.BaseMetadata):
+
+    def __init__(self, raw, folder):
+        super().__init__(raw)
+        self._folder = folder
```
chrisseto: Is `self._folder` ever used?

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+from waterbutler.providers.box.metadata import BoxRevision
+from waterbutler.providers.box.metadata import BoxFileMetadata
+from waterbutler.providers.box.metadata import BoxFolderMetadata
+
+from waterbutler.providers.box import utils as box_utils
+
+
+class BoxPath(utils.WaterButlerPath):
+
+    def __init__(self, path, prefix=False, suffix=False):
```
chrisseto: You don't need to include prefix and suffix here. In `super().__init__(...)` you should just specify them as they should never change.

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+        super().__init__(path, prefix=prefix, suffix=suffix)
+        if path != '/':
+            self._id = path.split('/')[1]
+        else:
+            self._id = path
+
+    def __repr__(self):
+        return "{}({!r})".format(self.__class__.__name__, self._orig_path)
```
chrisseto: This method is defined the same way in the superclass so there is no need to define it here.

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+class BoxProvider(provider.BaseProvider):
+
+    BASE_URL = settings.BASE_URL
```
chrisseto: Better to just refer to these as `settings.BASE_URL` throughout the class for clarity of where the value comes from.

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+                'to_path': dest_path.path,
+            },
+            expects=(200, ),
+            throws=exceptions.IntraMoveError,
+        )
+        data = yield from resp.json()
+        return BoxFileMetadata(data, self.folder).serialized()
+
+    @asyncio.coroutine # ✓
+    def download(self, path, revision=None, **kwargs):
```
chrisseto: Doesn't look like revision is an option here, does box allow for asking for versions?

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+            'GET',
+            self.build_url('files', path._id, 'content'),
+            expects=(200, ),
+            throws=exceptions.DownloadError,
+        )
+        return streams.ResponseStreamReader(resp)
+
+    @asyncio.coroutine # ✓
+    def upload(self, stream, path, **kwargs):
+        path = BoxPath('/'+self.folder+path)
```
chrisseto: use `os.path.join` here.

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+            self.build_url('files', path._id),
+            #data={'root': 'auto', 'path': path.path},
+            expects=(204, ),
+            throws=exceptions.DeleteError,
+        )
+
+    @asyncio.coroutine # ✓
+    def metadata(self, path, **kwargs):
+        path = BoxPath(path)
+        uid = self.folder
```
chrisseto: What is the purpose of assigning another variable here?

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+            'GET',
+            self.build_url('files', path._id, 'versions'),
+            expects=(200, ),
+            throws=exceptions.RevisionsError
+        )
+        data = yield from response.json()
+
+        idx = 1
+        ret = []
+        while(data['entries']):
```
chrisseto: Why is this not a list comprehension or just a for loop?

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+
+        #import ipdb; ipdb.set_trace()
+
+        return ret
+#        return [
+#            BoxRevision(item, folder=self.folder).serialized()
+#            for item in data['entries']
+#        ]
+
+    def can_intra_copy(self, dest_provider):
```
chrisseto: You don't need to define these.

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+        return type(self) == type(dest_provider)
+
+    def can_intra_move(self, dest_provider):
```
chrisseto: See above comment.

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,267 @@
...
+            'GET',
+            self.build_url('folders', uid, 'items'),
+            expects=(200, ),
+            throws=exceptions.MetadataError
+        )
+
+        data = yield from resp.json()
+
+        for item in data['entries']:
+            if item['type']=='file':
```
chrisseto: Add white space around operators.

waterbutler/providers/box/utils.py
```diff
@@ -0,0 +1,97 @@
+import http
+import uuid
+import asyncio
+
+from waterbutler.core import exceptions
+
+
+def file_or_error(article, file_id):
```
chrisseto: Is this used anywhere?

waterbutler/providers/box/metadata.py
mfraezz: A file will always be contained within a parent folder, yes. However, when you get the items in a folder (`GET /folders/{id}/items`), file metadata is created but `parent` isn't in the returned json.

waterbutler/providers/box/provider.py
mfraezz: They're not referred to anywhere in the class, but if unassigned aiohttp will `raise ValueError('Host could not be detected.')`

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,201 @@
...
+        return BoxFileMetadata(data['entries'][0], self.folder).serialized(), created
+
+    @asyncio.coroutine 
+    def delete(self, path, **kwargs):
+        #'etag' of the file can be included as an ‘If-Match’ header to prevent race conditions
+
+        path = BoxPath(path)
+        # A metadata call will verify the path specified is actually the
+        # requested file or folder.
+        yield from self.metadata(str(path))
```
chrisseto: ?

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,201 @@
...
+
+        idx = 1
+        ret = []
+        curr = yield from self.metadata(str(path))
+        curr['revision'] = idx
+        ret.append(BoxRevision(curr).serialized())
+
+        for item in data['entries']:
+            idx += 1
+            item['revision'] = idx
```
chrisseto: Is this value used anywhere given that box provides revision ids itself?

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,201 @@
...
+    def _get_file_meta(self, path):
+        resp = yield from self.make_request(
+            'GET',
+            self.build_url('files', path._id),
+            expects=(200, ),
+            throws=exceptions.MetadataError
+        )
+
+        data = yield from resp.json()
+        if data:
```
chrisseto: Does box not throw 404s?

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+    def name(self):
+        return self.raw['name']
+
+    @property
+    def path(self):
+        return '/{}/'.format(self.raw['id'])
+
+    @property
+    def content_type(self):
+        return self.raw['type']
```
chrisseto: This should be a mime or `None`, not file or folder.
screen shot 2015-02-17 at 10 55 55

chrisseto: Actually `content_type` shouldn't be include in folder metadata

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+    def name(self):
+        return self.raw['name']
+
+    @property
+    def path(self):
+        return '/{0}/{1}'.format(self.raw['id'], self.raw['name'])
+
+    @property
+    def size(self):
+        return self.raw.get('size')
```
chrisseto: From the box documentation it looks like size will always be present.

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    def parent(self):
+        return self.raw.get('parent').get('id')
+
+    @property
+    def folder(self):
+        return self.settings['folder']
+
+    @property
+    def content_type(self):
```
chrisseto: See above comment on `content_type`

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class BoxRevision(metadata.BaseFileRevisionMetadata):
+
+    @property
+    def size(self):
+        return self.raw.get('size')
+
+    @property
+    def kind(self):
```
chrisseto: Base file revision contains this property

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    def size(self):
+        return self.raw.get('size')
+
+    @property
+    def modified(self):
+        return self.raw.get('modified_at')
+
+    @property
+    def parent(self):
```
chrisseto: This should go in extra or be removed as its not used anywhere else.

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+    @property
+    def size(self):
+        return self.raw.get('size')
+
+    @property
+    def kind(self):
+        return 'file'
+
+    @property
+    def name(self):
```
chrisseto: This is also not required for revisions.

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+
+    @property
+    def content_type(self):
+        return self.raw['type']
+
+
+class BoxRevision(metadata.BaseFileRevisionMetadata):
+
+    @property
+    def size(self):
```
chrisseto: Also not required for revisions.

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,201 @@
...
+
+        path = BoxPath(path)
+        # A metadata call will verify the path specified is actually the
+        # requested file or folder.
+        yield from self.metadata(str(path))
+
+        yield from self.make_request(
+            'DELETE',
+            self.build_url('files', path._id),
+            #data={'root': 'auto', 'path': path.path},
```
chrisseto: Remove this line.

waterbutler/providers/box/metadata.py
mfraezz: Not for `GET folders/{id}/items`

waterbutler/providers/box/metadata.py
```diff
@@ -0,0 +1,104 @@
...
+            return '/{0}/{1}'.format(self.raw['id'], self.raw['name'])
+        except KeyError:
+            return self.raw['path']
+
+    @property
+    def modified(self):
+        return self.raw.get('modified_at')
+
+    @property
+    def revision(self):
```
chrisseto: This should be in extra

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,201 @@
...
+    @property
+    def default_headers(self):
+        return {
+            'Authorization': 'Bearer {}'.format(self.token),
+        }
+
+    @asyncio.coroutine 
+    def download(self, path, revision=None, **kwargs):
+        path = BoxPath(path)
+        if revision and revision != path._id:
```
chrisseto: Consider just build the url in the if else blocks for less redundancy

waterbutler/providers/box/provider.py
```diff
@@ -0,0 +1,201 @@
...
+
+        ret = []
+        for item in data['entries']:
+            if item['type'] == 'folder':
+                ret.append(BoxFolderMetadata(item, self.folder).serialized())
+            else:
+                ret.append(BoxFileMetadata(item, self.folder).serialized())
+        return ret
+
+    def _upload_create(self, stream, path):
```
chrisseto: Update both upload methods to use the `FormDataStream` class

waterbutler/providers/box/utils.py
```diff
@@ -0,0 +1,84 @@
+import http
```
chrisseto: You should be able to remove this file now.

waterbutler/providers/box/provider.py
mfraezz: Done

waterbutler/providers/box/metadata.py
mfraezz: Updated to return `None`

waterbutler/providers/box/provider.py
mfraezz: Made a note related to what we discussed about this

waterbutler/providers/box/metadata.py
mfraezz: `modified_at` is normally included, but in the case of the current revision, `modified` is included instead (due to the BoxFileMetadata property of that name). Possible to change this, but it might have unintended consequences on the OSF-side

chrisseto: That's an excellent point I didn't realize the keys were different. Both should  be `modified`. I'll let you know when the change it made waterbutler side.

waterbutler/providers/box/metadata.py
```diff
@@ -77,5 +83,7 @@ def modified(self):
             return self.raw.get('modified')
 
     @property
-    def revision(self):
-        return self.raw['etag']
+    def extra(self):
+        return {
+            'etag': self.raw.get('etag')
```
icereval: Do we expect an `etag` every time? If so it should be directly accessed `self.raw['etag']`

mfraezz: Here: yes, above: no. You're right

PR #12, opened 2015-02-16
Feature/form data

chrisseto:


waterbutler/core/streams.py
```diff
@@ -140,3 +141,149 @@ def write(self, data):
...
+    Usage:
+        >>> stream = FormDataStream(key1='value1', file=FileStream(...))
+    Or:
+        >>> stream = FormDataStream()
+        >>> stream.add_field('key1', 'value1')
+        >>> stream.add_file('file', FileStream(...), mime='text/plain')
+    Additional options for files can be passed as a tuple ordered as:
+        >>> FormDataStream(fieldName=(FileStream(...), 'fileName', 'Mime', 'encoding'))
+
+    Auto generates boundarys and properly concatenates them
```
jmcarp: boundarys -> boundaries

waterbutler/core/streams.py
```diff
@@ -140,3 +141,149 @@ def write(self, data):
...
+    Use FormDataStream.headers to get the proper headers to be included with requests
+    Namely Content-Length, Content-Type
+    """
+    FORM_DATA_HEADER = 'Content-Disposition: form-data; name="{}"\r\n\r\n'
+    FILE_HEADER = 'Content-Disposition: file; name="{}"{}\r\nContent-Type: {}\r\nContent-Transfer-Encoding: {}\r\n\r\n'
+
+    @classmethod
+    def make_boundary(self):
```
jmcarp: This method is static. How about decorating with `@staticmethod` or moving out of the class entirely (I prefer the second option)?

waterbutler/core/streams.py
```diff
@@ -140,3 +141,149 @@ def write(self, data):
...
+        >>> stream.add_field('key1', 'value1')
+        >>> stream.add_file('file', FileStream(...), mime='text/plain')
+    Additional options for files can be passed as a tuple ordered as:
+        >>> FormDataStream(fieldName=(FileStream(...), 'fileName', 'Mime', 'encoding'))
+
+    Auto generates boundarys and properly concatenates them
+    Use FormDataStream.headers to get the proper headers to be included with requests
+    Namely Content-Length, Content-Type
+    """
+    FORM_DATA_HEADER = 'Content-Disposition: form-data; name="{}"\r\n\r\n'
```
jmcarp: These are a little brittle. How about something more generic, like a `build_header` method?

waterbutler/core/streams.py
```diff
@@ -140,3 +141,149 @@ def write(self, data):
...
+
+    @asyncio.coroutine
+    def read(self, n=-1):
+        if self.can_add_more:
+            self.finalize()
+        return super().read(n=n)
+
+    def finalize(self):
+        if self.can_add_more:
+            if not self.stream and self.streams:
```
jmcarp: Is this purpose of this line to make sure that `self.stream` is defined so aiohttp can call `read`? Maybe this logic can move to the parent class--or at least we can add some documentation here.

jmcarp:
Added a few questions inline, but LGTM :cake: 
Once we're confident this is working with all providers (@chrisseto mentioned that it wasn't working with Box earlier), let's also add tests.

chrisseto:
@jmcarp There are a few tests for basic functionality

waterbutler/core/streams.py
chrisseto: :+1:

waterbutler/core/streams.py
chrisseto: I doesn't really make sense to be anywhere outside of the class IMO.
It might be better to just dissolve this method and just call `uuid...` in `__init__` as thats the only place it is used.

jmcarp: Sounds good. I just don't want to use a static method that's decorated with `classmethod` but takes `self` ಠ_ಠ

chrisseto: Missed the `self` ლ(ಠ_ಠლ)
If that was changed to `cls` would that be acceptable?
It also makes testing easier as just that method needs to be mocked. (☞ ͡° ͜ʖ ͡°)☞

waterbutler/core/streams.py
```diff
@@ -140,3 +141,187 @@ def write(self, data):
...
+        return {
+            'Content-Length': str(self.size),
+            'Content-Type': 'multipart/form-data; boundary={}'.format(self.boundary)
+        }
+
+    @asyncio.coroutine
+    def read(self, n=-1):
+        if self.can_add_more:
+            self.finalize()
+        return super().read(n=n)
```
jmcarp: The superclass `read` is a coroutine. Should this be `return (yield from super().read(n=n))`?

waterbutler/core/streams.py
```diff
@@ -140,3 +141,187 @@ def write(self, data):
...
+    def add_streams(self, *streams):
+        self._streams.extend(streams)
+
+        if not self.stream:
+            self._cycle()
+
+    @asyncio.coroutine
+    def read(self, n=-1):
+        if not self.stream:
+                return b''
```
jmcarp: Extra indent here.

tests/core/test_streams.py
```diff
@@ -0,0 +1,140 @@
...
+
+    def test_cant_add_after_finalize(self):
+        stream = streams.FormDataStream(valjean='24601')
+        stream.finalize()
+
+        with pytest.raises(AssertionError):
+            stream.add_field('javer', 'thelaw')
+
+    def test_headers_calls_finalize(self):
+        stream = streams.FormDataStream(valjean='24601')
```
jmcarp: Five years for what you did, the rest because you tried to run

waterbutler/core/streams.py
chrisseto: That is an excellent point... but it seems to be working?

PR #14, opened 2015-02-17
Feature/gdrive

kushG:
<h3> Purpose </h3>
This PR adds Google Drive addon specific files into waterbutler.

<h3> Changes </h3>
No changes in front end. All the changes made back-end are pretty much similar to the way dropbox is working. Path handling is relatively different here. Path contains three fields folderId, Name & path. This is done to retrieve metadata from drive using folderId as Google drive API does not support getting metadata using path.

<h3> Side Effects </h3>
As mentioned above, due to the difference in the way of path handling, future changes in path of generic view functions may result in failure of this.


waterbutler/providers/gdrive/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+    def path(self):
+        return '/{0}/{1}/{2}'.format(self.raw['id'], self.raw['title'], self.raw['path'])
+
+    @property
+    def size(self):
+        # Google docs(Docs,sheets, slides, etc)  don't have file size before they are exported
+        try:
+            return self.raw['fileSize']
+        except KeyError:
+            return '0'
```
jmcarp: Size is expected to be an int or `None`. I think `None` makes more sense here--the file doesn't actually contain zero bytes, we just don't know how big it is.

waterbutler/providers/gdrive/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+
+    @property
+    def modified(self):
+        #return self.raw['modified']
```
jmcarp: Do we get modification times from gdrive?

waterbutler/providers/gdrive/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+        pass
+    @property
+    def content_type(self):
+        return self.raw['mimeType']
+
+
+# TODO dates!
```
jmcarp: Is this comment relevant? Looks like you have modified date implemented here.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+        return "{}({!r}, {!r})".format(self.__class__.__name__, self._folderId, self._orig_path)
+
+    @property
+    def full_path(self):
+        return self._full_path
+
+
+class GoogleDriveProvider(provider.BaseProvider):
+
+    BASE_URL = settings.BASE_URL
```
jmcarp: Why copy these config variables to class variables? Are the class variables used?

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+        self.folder = self.settings['folder']
+
+    @property
+    def default_headers(self):
+        return {
+            'authorization': 'Bearer {}'.format(self.token),
+        }
+
+    @asyncio.coroutine
+    def download(self, path, revision=None, **kwargs):
```
jmcarp: We don't seem to use the `revision` parameter here. Does gdrive allow us to download files at previous revisions?

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+    @asyncio.coroutine
+    def delete(self, path, **kwargs):
+        # A metadata call will verify the path specified is actually the
+        # requested file or folder.
+        yield from self.metadata(str(path))
+        path = GoogleDrivePath(path, self.folder)
+        yield from self.make_request(
+            'DELETE',
+            self.build_url('files', path._folderId),
+            # data={'root': 'auto', 'path': path.full_path},
```
jmcarp: Remove commented-out code if we don't need it.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+            expects=(200, 204),
```
jmcarp: Does gdrive return a 200 or 204 for successful delete requests, or both? If they return a consistent status code (which hopefully they do), we shouldn't expect both.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+                expects=(200, ),
+                throws=exceptions.MetadataError
+            )
+            data = yield from resp.json()
+            data['path'] = os.path.join(path.full_path, data['title'])
+
+        else:
+            ret = []
+            for item in data['items']:
+                item['path'] = os.path.join(path.full_path, item['title'])  # custom add, not obtained from API
```
jmcarp: Could you clarify what this comment means?

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+                'GET',
+                self.build_url('files', path._folderId),
+                expects=(200, ),
+                throws=exceptions.MetadataError
+            )
+            data = yield from resp.json()
+            data['path'] = os.path.join(path.full_path, data['title'])
+
+        else:
+            ret = []
```
jmcarp: Prefer list comprehension to explicit for loop.

waterbutler/providers/gdrive/settings.py
```diff
@@ -0,0 +1,10 @@
+try:
+    from waterbutler import settings
+except ImportError:
+    settings = {}
+
+config = settings.get('GOOGLEDRIVE_PROVIDER_CONFIG', {})
+
+
+BASE_URL = config.get('BASE_URL', 'https://www.googleapis.com/drive/v2')
+BASE_CONTENT_URL = config.get('BASE_CONTENT_URL', 'https://www.googleapis.com/upload/drive/v2')
```
jmcarp: Could this be named something like `BASE_UPLOAD_URL` for clarity?

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+
+        resp = yield from self.make_request(
+            'GET',
+            self.build_url('files', q="'%s' in parents and trashed = false" % path._folderId, alt="json"),
+            expects=(200, ),
+            throws=exceptions.MetadataError
+        )
+        data = yield from resp.json()
+
+        # Check to see if returned data is a file or folder
```
jmcarp: How about splitting this logic into a few different helper functions? For example,
```python
if not data['items']:
    return (yield from self._file_metadata(path))
return self._folder_metadata(items)
```

Then `_file_metadata` and `_folder_metadata` will be small, readable functions that will be easy to test.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+        #todo:make this proper
+        upload_id = query_params.split("=")[2]
+
+        #Step 3 - Upload the file
+
+        resp = yield from self.make_request(
+            'PUT',
+            self._build_content_url("files", uploadType='resumable', upload_id=upload_id),
+            headers={'Content-Length': str(stream.size),
+                     #'Content-Type': 'image/jpeg',#todo: hardcoded in for testing
```
jmcarp: Remove commented-out code.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+    def upload(self, stream, path, **kwargs):
+
+        try:
+            yield from self.metadata(str(path))
+        except exceptions.MetadataError:
+            created = True
+        else:
+            created = False
+        path = GoogleDrivePath(path, self.folder, isUpload=True)
+        content = yield from stream.read()
```
jmcarp: Do we need to call `stream.read` here? We can normally pass the stream directly to aiohttp (i.e., change line 153 from `data=content` to `data=stream`).

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+        path = GoogleDrivePath(path, self.folder)
+        resp = yield from self.make_request(
+            'GET',
+            self.build_url('files', path._folderId),
+            expects=(200, ),
+            throws=exceptions.DownloadError,
+        )
+        data = yield from resp.json()
+        download_url = data['downloadUrl']
+        downloadResp = yield from self.make_request(
```
jmcarp: In Python, prefer underscore_cased variable names to camelCased.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+                    "kind": "drive#parentReference",
+                    "id": path._folderId
+                },
+            ],
+
+            "title": path._uploadFileName,
+            #"Content-Type": "image/png",
+
+        }
+        #Step 1 - Start a resumable session
```
jmcarp: How about breaking these steps out into separate helper functions?

waterbutler/providers/gdrive/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+    def id(self):
+        return self.raw['id']
+
+    @property
+    def name(self):
+        return self.raw['title']
+
+    @property
+    def path(self):
+        return'/{0}/{1}/{2}'.format(self.raw['id'], self.raw['title'], self.raw['path'])
```
jmcarp: Prefer using `os.path.join` over string formatting to build paths.

waterbutler/providers/gdrive/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, raw, folder):
+        super().__init__(raw)
+        self._folder = folder
+
+    @property
+    def provider(self):
+        return 'gdrive'
+
+    def build_path(self, path):
```
jmcarp: Is this used?

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+                    else:
+                        tempPath = tempPath + '/' + parts[i]
+                self._path = tempPath
+                full_path = tempPath
+            self._full_path = self._format_path(full_path)
+        self._uploadFileName = ""
+        if isUpload:
+
+            #this is VERY HACKISH. MUST be fixed once other code is fixed.
+            folder_plus_name = parts[-1:][0]
```
jmcarp: FWIW, `array[-1:][0]` is equivalent to `array[-1]`

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+        parts = path.strip('/').split('/')
+        if len(parts) > 1:
+            name = parts[1]  # TODO : Remove this later if of no use
+            folderId = parts[0]
+            self._folderId = folderId
+            self._name = name
+            if folderId == folder['id']:
+                full_path = folder['name']
+            else:
+                tempPath = ''
```
jmcarp: How about replacing this loop with something like
```python
full_path = os.path.join(*[parts[2:]])
```
Then we drop the loop and the `tempFile` variable.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+from waterbutler.providers.gdrive.metadata import GoogleDriveFolderMetadata
+
+
+class GoogleDrivePath(utils.WaterButlerPath):
+
+    def __init__(self, path, folder, isUpload=False, prefix=True, suffix=False):
+        super().__init__(path, prefix=prefix, suffix=suffix)
+
+        parts = path.strip('/').split('/')
+        if len(parts) > 1:
```
jmcarp: What happens if parts has length 1? Looks like we don't set `self._path` or `self._full_path` in that case. Is that expected?

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+            name = parts[1]  # TODO : Remove this later if of no use
```
jmcarp: Please remove if unnecessary.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+                for i in range(2, len(parts)):
+                    if tempPath == '':
+                        tempPath = parts[i]
+                    else:
+                        tempPath = tempPath + '/' + parts[i]
+                self._path = tempPath
+                full_path = tempPath
+            self._full_path = self._format_path(full_path)
+        self._uploadFileName = ""
+        if isUpload:
```
jmcarp: Instead of passing `isUpload` to the constructor, how about making `upload_name` a property or helper method of `GoogleDrivePath`? Also, looks like you flagged this as needing a rewrite--please make that change when you get a chance.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+
+        #Step 2 - Save the resumable session URI
+
+        yield from resp.json()
+
+        location = resp.headers['LOCATION']
+
+        query_params = urlparse(location).query
+        #todo:make this proper
+        upload_id = query_params.split("=")[2]
```
jmcarp: Parsing query strings manually is brittle and easy to mess up. Prefer to use a helper like furl:
```python
parsed_url = furl.furl(resp.headers['LOCATION'])
upload_id = parsed_url.args[QUERY_KEY]
```
where `QUERY_KEY` is the name of the query parameter you want to extract.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+            throws=exceptions.DeleteError,
+        )
+
+    @asyncio.coroutine
+    def metadata(self, path, **kwargs):
+        path = GoogleDrivePath(path, self.folder)
+
+        resp = yield from self.make_request(
+            'GET',
+            self.build_url('files', q="'%s' in parents and trashed = false" % path._folderId, alt="json"),
```
jmcarp: For consistency, prefer `str.format` to percent-formatting. Also, this query syntax looks kind of weird--could you include a comment linking to the relevant docs?

waterbutler/providers/gdrive/metadata.py
kushG: We get modified date which includes time as well.

waterbutler/providers/gdrive/metadata.py
kushG: Removed the comment. Haven't tried date yet, it was just copied from dropbox code.

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
...
+        self.token = self.credentials['token']
+        # self.refresh_token = self.credentials['refresh_token']
+        # self.client_id = self.credentials['client_id']
+        # self.client_secret = self.credentials['client_secret']
+        self.folder = self.settings['folder']
+
+    @property
+    def default_headers(self):
+
+        # # Refesh access_token before making any calls
```
jmcarp: Delete unused code. Has this been moved to the OSF side?

waterbutler/providers/gdrive/provider.py
```diff
@@ -0,0 +1,227 @@
+import os
+import asyncio
+import json
+import requests
```
jmcarp: Delete unused import.

icereval:
Merge GoogleDrive :+1:

PR #17, opened 2015-02-25
Handle docs extensions

jmcarp:
Detect file extensions for Google Docs files. Append appropriate pseudo-extension to file names; download using appropriate export links.

### DIFF
--- a/waterbutler/providers/googledrive/metadata.py
+++ b/waterbutler/providers/googledrive/metadata.py
@@ -1,5 +1,8 @@
 import os
+
 from waterbutler.core import metadata
+
+from waterbutler.providers.googledrive import utils
 
 
 class BaseGoogleDriveMetadata(metadata.BaseMetadata):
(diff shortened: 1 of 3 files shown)


waterbutler/providers/googledrive/metadata.py
```diff
@@ -42,7 +43,12 @@ def id(self):
 
     @property
     def name(self):
-        return self.raw['title']
+        title = self.raw['title']
+        name, ext = os.path.splitext(title)
+        if utils.is_docs_file(self.raw) and not ext:
+            ext = utils.get_extension(self.raw['exportLinks'])
+            title += '.' + ext
```
chrisseto: Chaining with `+=` looks a bit odd IMO, @icereval?

```python
title = '{}.{}'.format(title, ext)
```
or maybe some parens?
```python
title += ('.' + ext)
```

PR #24, opened 2015-03-10
Prevent deletion of public files.

jmcarp:
# Purpose
The Figshare API allows deleting public files, which shouldn't be deletable. On delete, send a metadata request, and bail out if file is public. We wrote to the Figshare team about this, but for now, disabling deletion on these files is probably safest.

@chrisseto


chrisseto:
+1
@icereval Thoughts?

icereval:
LGTM

PR #26, opened 2015-03-12
Add canDelete field to figshare extra metadata.

jmcarp:
Second attempt to handle figshare deletes. Based on correspondence with the figshare lead, it is actually possible to delete public / published files, but only when the file being deleted is not the last file in the fileset. This patch adds a `canDelete` field to the figshare file metadata indicating whether the file can be deleted.


jmcarp:
@chrisseto can you review?

PR #31, opened 2015-03-30
Feature/move n copy

chrisseto:


waterbutler/providers/dataverse/provider.py
```diff
@@ -153,12 +171,15 @@ def metadata(self, path, version=None, **kwargs):
         # Get appropriate metadata
         dataset_metadata = yield from self._get_data(version)
```
rliebz: To properly take advantage of caching:

```
# Get appropriate metadata
dataset_metadata = yield from self._maybe_fetch_metadata(version=version, refresh=True)

if path.is_root
    return dataset_metadata

File: waterbutler/providers/dataverse/provider.py [hunk 2]
```diff
@@ -153,12 +171,15 @@ def metadata(self, path, version=None, **kwargs):
...
-            return dataset_metadata
+        if path.is_root:
+            return (yield from self._maybe_fetch_metadata(version=version))
 
         try:
             return next(
-                item for item in dataset_metadata if item['path'] == path
+                item
+                for item in
+                (yield from self._maybe_fetch_metadata(version=version))
```
rliebz: Since we're storing this in `dataset_metadata`, you can just use that.

If you'd prefer to just use the caching in `_maybe_fetch_metadata`, then don't use the variable `dataset_metadata` and make sure to pass `refresh=True` when relevant.

waterbutler/providers/googledrive/provider.py
```diff
@@ -62,10 +36,87 @@ def __init__(self, auth, credentials, settings):
...
+    def validate_path(self, path, file_id=None, **kwargs):
+        if path == '/':
+            return GoogleDrivePath('/', _ids=[self.folder['id']], folder=True)
+
+        if file_id:
+            parts = yield from self._resolve_id_to_parts(file_id)
+        elif path:
+            parts = yield from self._resolve_path_to_ids(path)
+        else:
+            raise Exception  # TODO
```
rliebz: Did you mean to leave this here?

waterbutler/server/handlers/core.py
```diff
@@ -104,25 +99,99 @@ def write_error(self, status_code, exc_info):
...
+    def json(self):
+        try:
+            return self._json
+        except AttributeError:
+            pass
+        try:
+            self._json = json.loads(self.request.body.decode('utf-8'))
+        except ValueError:
+            if self.JSON_REQUIRED:
+                raise Exception  # TODO
```
rliebz: You'll probably want to raise a more specific exception

PR #32, opened 2015-04-09
Feature/dataverse

rliebz:


tests/providers/dataverse/test_provider.py
```diff
@@ -0,0 +1,450 @@
...
+        <category term="isMinorUpdate" scheme="http://purl.org/net/sword/terms/state" label="State">true</category>
+        <category term="locked" scheme="http://purl.org/net/sword/terms/state" label="State">false</category>
+        <category term="latestVersionState" scheme="http://purl.org/net/sword/terms/state" label="State">DRAFT</category>
+    </feed>'''
+
+
+@pytest.fixture
+def native_dataset_metadata():
+    return {'data': {u'createTime': u'2015-04-02T13:21:59Z',
+ u'distributionDate': u'Distribution Date',
```
chrisseto: You dont need any of the `u`s here

waterbutler/providers/dataverse/metadata.py
```diff
@@ -0,0 +1,124 @@
...
+        super().__init__(raw)
+        self._name = name
+        self.doi = doi
+
+        if native:
+            files = self.raw['files']
+            self._entries = [DataverseNativeFileMetadata(f['datafile']) for f in files]
+
+        else:
+            entry_feed = raw['feed'].get('entry') or []
```
chrisseto: `.get('entry', [])`

waterbutler/providers/dataverse/provider.py
```diff
@@ -0,0 +1,169 @@
...
+        self.token = self.credentials['token']
+        self.doi = self.settings['doi']
+        self.id = self.settings['id']
+        self.name = self.settings['name']
+
+    @asyncio.coroutine
+    def download(self, path, **kwargs):
+        resp = yield from self.make_request(
+            'GET',
+            provider.build_url(self.DOWN_BASE_URL, path),
```
chrisseto: `self.build_url`

waterbutler/providers/dataverse/provider.py
```diff
@@ -0,0 +1,169 @@
...
+            expects=(200, ),
+            throws=exceptions.DownloadError,
+            params={'key': self.token},
```
chrisseto: This can be passed into `build_url`

waterbutler/providers/dataverse/provider.py
chrisseto: There should be some about of path validation here.

waterbutler/providers/dataverse/provider.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        return streams.ResponseStreamReader(resp)
+
+    @asyncio.coroutine
+    def upload(self, stream, path, **kwargs):
+
+        filename = path.strip('/')
```
chrisseto: Re: path validation.

waterbutler/providers/dataverse/provider.py
chrisseto: Consider creating a `DataversePath` class

rliebz: `self.build_url` joins `self.BASE_URL`, as opposed to `provider.build_url` which does not. Since Dataverse does not have a `BASE_URL`, it cannot take advantage of the inherited `build_url`. I can override `self.build_url` if that's preferable, but the behavior I want is from `provider.build_url`.

waterbutler/providers/dataverse/settings.py
```diff
@@ -0,0 +1,13 @@
+try:
+    from waterbutler import settings
+except ImportError:
+    settings = {}
+
+config = settings.get('DATAVERSE_PROVIDER_CONFIG', {})
+
+HOSTNAME = config.get('HOSTNAME', 'apitest.dataverse.org')
```
chrisseto: In `DataverseProvider`
```
class ...:
   BASE_URL = 'https://{}'.format(HOSTNAME)

def download(...):
    download_url = self.build_url('api/access/datafile/', query='param')
    ...
```

tests/providers/dataverse/test_provider.py
rliebz: Done.

waterbutler/providers/dataverse/metadata.py
rliebz: Done.

waterbutler/providers/dataverse/provider.py
rliebz: Done.

waterbutler/providers/dataverse/settings.py
rliebz: Done.

waterbutler/providers/dataverse/provider.py
rliebz: Done.

waterbutler/providers/dataverse/provider.py
rliebz: Validation done using a helper function, due to the nature of Dataverse path validation.

rliebz:
Ready for a second pass :baby_chick:

waterbutler/providers/dataverse/provider.py
rliebz: By which I mean path validation is done where I thought it was needed, but because there is no namespacing by path in dataverse, I don't think there is any validation to be done per upload. The provider never allows an opportunity to upload files to any dataset other than the one specified in settings.

waterbutler/core/streams.py
```diff
@@ -325,3 +331,173 @@ def add_file(self, field_name, file_stream, file_name=None, mime='application/oc
...
+            chunk += self._flush()
+        self._cycle()
+        nextn = -1 if n == -1 else n - len(chunk)
+        next_chunk = (yield from self.read(nextn))
+        chunk += self._compress(next_chunk) if self.on_data_stream else next_chunk
+        return chunk
+
+    def add_file(self, filename, file_stream):
+
+        self.filename = filename
```
chrisseto: Does this support multiple files?

if not you may just want to remove this method and but the logic in the constructor

waterbutler/core/streams.py
```diff
@@ -325,3 +331,173 @@ def add_file(self, field_name, file_stream, file_name=None, mime='application/oc
...
+            0,
+            count,
+            count,
+            len(output),
+            centdir_offset,
+            0,
+        )
+
+        output += endrec
+        self.footer = output
```
chrisseto: Its confusing to store this on the object if its never used.

waterbutler/providers/dataverse/provider.py
```diff
@@ -0,0 +1,191 @@
...
+        yield from self.make_request(
+            'DELETE',
+            self.build_url(self.EDIT_MEDIA_BASE_URL, 'file', path),
+            auth=(self.token, ),
+            expects=(204, ),
+            throws=exceptions.DeleteError,
+        )
+
+    @asyncio.coroutine
+    def metadata(self, path='/', state=None, **kwargs):
```
chrisseto: Path should not have a default value.

waterbutler/providers/dataverse/metadata.py
```diff
@@ -0,0 +1,124 @@
...
+    @property
+    def provider(self):
+        return 'dataverse'
+
+
+class BaseDataverseFileMetadata(BaseDataverseMetadata, metadata.BaseFileMetadata):
+
+    def __init__(self, raw):
+        super().__init__(raw)
+        original_name, version = dataverse_utils.unpack_filename(self.name)
```
chrisseto: If dataverse doesn't support versioning whats the purpose of returning it?

rliebz: This isn't technically a version, but more analogous to the way a file system will append a number to the end of a duplicate file. Because the Dataverse doesn't do updates, but instead renames duplicate files, this seemed like the best way to keep track of which file was uploaded.

To clarify, if you upload `name.txt` when there is already a `name.txt`, the API will accept it and rename it as `name-1.txt`, but not include that information in the response. What we use the version for is a search through the updated list of files and choose the one with the most recent version number (with a slightly more complex method to choose the correct file version, at https://github.com/rliebz/waterbutler/blob/d414f3ede3e0aab7881b2a33a3e815fa1a56f1fd/waterbutler/providers/dataverse/provider.py#L79-L97)

waterbutler/providers/dataverse/utils.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+
+# For use in determining which version of a file was just uploaded
+def unpack_filename(filename):
```
chrisseto: If I decide to name a file `name-1.txt` and another file `name.txt` what would happen?

rliebz: In the case that `name.txt` was uploaded second, waterbutler would pick the wrong file. There isn't really any logic you could apply to filename checking that would solve that problem, since uploading `name.txt` twice would produce the same list of files but have a different correct resolution.

As for alternatives:
 - I've tried looking at timestamps, but the dataset as a whole is stamped rather than individual files. There is no way to tell which one has been uploaded more recently.
 - We could make a metadata request and diff the files before and after the upload to find which one is new. While this would solve the problem you pointed out, it opens up the issue of what happens when there are multiple people working on a dataset at once.
 - We could repopulate the entire list of files. This one would be accurate in every situation, but mean a forced refresh every time an upload happens.

waterbutler/providers/dataverse/provider.py
rliebz: Done.

waterbutler/core/streams.py
rliebz: Fixed.

waterbutler/core/streams.py
rliebz: Done.

waterbutler/providers/dataverse/utils.py
rliebz: This has been resolved by deleting files with the same name before performing an upload, avoiding Dataverse's versioning system altogether. This is the same behavior that the pre-waterbutler OSF uses on production.

waterbutler/providers/dataverse/metadata.py
rliebz: Resolved. Explanation in https://github.com/CenterForOpenScience/waterbutler/pull/32/files#r28357338

rliebz:
Ready for another pass :hatching_chick:

setup.py
```diff
@@ -43,6 +43,7 @@ def parse_requirements(requirements):
             'github = waterbutler.providers.github:GitHubProvider',
             'osfstorage = waterbutler.providers.osfstorage:OSFStorageProvider',
             's3 = waterbutler.providers.s3:S3Provider',
+            'dataverse = waterbutler.providers.dataverse:DataverseProvider'
```
icereval: missing comma

waterbutler/core/streams.py
```diff
@@ -1,8 +1,14 @@
 import os
 import abc
 import uuid
+import struct
+import time
+import binascii
 import asyncio
 
+import zipfile
```
chrisseto: zipfile and zlib are stdlib there shouldn't be a new line above them.

waterbutler/providers/dataverse/metadata.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def provider(self):
+        return 'dataverse'
+
+
+class DataverseFileMetadata(BaseDataverseMetadata, metadata.BaseFileMetadata):
+
+    @property
+    def id(self):
```
chrisseto: id is a builtin in python, can you make this `file_id`

waterbutler/providers/dataverse/metadata.py
```diff
@@ -0,0 +1,72 @@
...
+    def name(self):
+        return self.raw['name']
+
+    @property
+    def path(self):
+        return self.build_path(self.id)
+
+    @property
+    def size(self):
+        pass
```
chrisseto: prefer to explicitly return `None` here

waterbutler/providers/dataverse/metadata.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @property
+    def content_type(self):
+        return self.raw['contentType']
+
+    @property
+    def modified(self):
```
chrisseto: Re: Explicit `None`

waterbutler/providers/dataverse/metadata.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.build_path(self.doi)
+
+    @property
+    def entries(self):
+        return self._entries
+
+    def serialized(self):
+        if self._entries:
+            return [e.serialized() for e in self._entries]
+        return super(DataverseDatasetMetadata, self).serialized()
```
chrisseto: python3 just requires `super().serialized()`

waterbutler/providers/dataverse/settings.py
```diff
@@ -0,0 +1,13 @@
...
+
+EDIT_MEDIA_BASE_URL = config.get('EDIT_MEDIA_BASE_URL', "/dvn/api/data-deposit/v1.1/swordv2/edit-media/")
```
chrisseto: Is there a reason these cant just be literal strings in the methods in which they are relevant?

waterbutler/providers/dataverse/settings.py
```diff
@@ -0,0 +1,13 @@
...
+DOWN_BASE_URL = config.get('DOWN_BASE_URL', "/api/access/datafile/")
+METADATA_BASE_URL = config.get('METADATA_BASE_URL', "/dvn/api/data-deposit/v1.1/swordv2/statement/study/")
+JSON_BASE_URL = config.get('JSON_BASE_URL', "/api/datasets/{0}/versions/:{1}")
```
chrisseto: Is the colon here intentional?

waterbutler/providers/dataverse/provider.py
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        try:
+            # Delete old file if it exists
+            metadata = yield from self.get_data('latest')
+            files = metadata if isinstance(metadata, list) else []
+            old_file = next(file for file in files if file['name'] == filename)
+            yield from self.delete(old_file['path'])
+            created = False
+        except StopIteration:
```
chrisseto: It may make the logic a bit messier but can you tighten up the try except?
Its rather large.

waterbutler/providers/dataverse/provider.py
```diff
@@ -0,0 +1,176 @@
...
+                "Could not retrieve file '{}'".format(path),
+                code=http.client.NOT_FOUND,
+            )
+
+    @asyncio.coroutine
+    def revisions(self, path, **kwargs):
+        raise exceptions.ProviderError({'message': 'Dataverse does not support file revisions.'}, code=405)
+
+    @asyncio.coroutine
+    def get_data(self, version):
```
chrisseto: If version will only ever be `latest` or `latest-published` could you make this a boolean

waterbutler/providers/dataverse/provider.py
```diff
@@ -0,0 +1,176 @@
...
+
+        data = yield from resp.json()
+        data = data['data']
+
+        return DataverseDatasetMetadata(
+            data, self.name, self.doi,
+        ).serialized()
+
+    @asyncio.coroutine
+    def get_all_data(self):
```
chrisseto: Its waterbutler convention to not "expose" and methods that a reserved for internal use.
So any non-public method, not download, upload, etc,  should be prefixed with `_`

waterbutler/providers/dataverse/settings.py
rliebz: There are two main benefits I saw:
 - Referencing these in testing makes it a lot easier to write tests
 - In the likely scenario that Dataverse introduces a new route that better suits our needs or we bump the API version in the URL, this reduces the number of places we need to change the code. 

If there's a reason to use the plain strings instead, we can do that, but they are lengthy (e.g., `/dvn/api/data-deposit/v1.1/swordv2/statement/study/`)

waterbutler/providers/dataverse/settings.py
rliebz: Yes

waterbutler/providers/dataverse/provider.py
rliebz: We are currently only supporting `latest` and `latest-published` because that is what their old API supported, but in the scenario that we decide to support versioning, the infrastructure is there on their end.

chrisseto: :+1:

chrisseto:
lgtm :+1: 

I'll make some of the suggested changes during the merge.

PR #37, opened 2015-04-21
Add support for dataverse file revisions

rliebz:


waterbutler/providers/dataverse/provider.py
```diff
@@ -129,7 +133,17 @@ def metadata(self, path, state=None, **kwargs):
 
     @asyncio.coroutine
     def revisions(self, path, **kwargs):
-        raise exceptions.ProviderError({'message': 'Dataverse does not support file revisions.'}, code=405)
+        versions = ['latest', 'latest-published']
```
chrisseto: Can this be defined as a class level tuple?

rliebz: If we ever do decide to allow numbered versions (I'm not sure yet how viable that would be through the API), each dataset would have a different list of versions. That being said, for the immediate future, these will be the only two versions. I think this way is a bit more conducive to future changes.

rliebz:
Requires https://github.com/CenterForOpenScience/osf.io/pull/2570

waterbutler/providers/dataverse/provider.py
```diff
@@ -167,9 +229,17 @@ def _get_all_data(self):
         published_files = published_data if isinstance(published_data, list) else []
```
chrisseto: What could `published_data` be other than a list?

waterbutler/providers/dataverse/provider.py
```diff
@@ -159,7 +221,7 @@ def _get_data(self, version):
 
     @asyncio.coroutine
     def _get_all_data(self):
-        # Unspecified (file view page), check both sets for metadata
+        """Get list of file metadata for all dataset versions"""
         try:
             published_data = yield from self._get_data('latest-published')
         except exceptions.MetadataError:
```
chrisseto: You may want to do some checks on the status code here, if dataverse is down and returns a 503 that should be reraised.

waterbutler/providers/dataverse/provider.py
```diff
@@ -129,16 +168,39 @@ def metadata(self, path, state=None, **kwargs):
...
         :param str version:
-            'latest' for draft files
-            'latest-published' for published files
+
+            - 'latest' for draft files
+            - 'latest-published' for published files
+            - None for all data
         """
 
+        if not version:
```
chrisseto: Should version default to `None`?

waterbutler/providers/dataverse/provider.py
```diff
@@ -129,16 +168,39 @@ def metadata(self, path, state=None, **kwargs):
...
+            data = yield from self._get_all_data()
```
chrisseto: You can also just `return (yield from self._get_all_data())`

waterbutler/providers/dataverse/provider.py
```diff
@@ -129,16 +168,39 @@ def metadata(self, path, state=None, **kwargs):
...
+        """Get past versions of the request file. Currently only checks
+        'latest' and 'latest-published' versions.
+
+        :param str path: The path to a key
+        :rtype list:
+        """
+        versions = ['latest', 'latest-published']
+        revisions = []
+        for version in versions:
+            metadata = yield from self._get_data(version)
```
chrisseto: Any reason not to us get_all_data here and then iterate over that?

waterbutler/providers/dataverse/metadata.py
```diff
@@ -77,3 +77,18 @@ def serialized(self):
...
+    def version_identifier(self):
+        return 'version'
+
+    @property
+    def version(self):
+        return self.raw
+
+    @property
+    def modified(self):
+        return None
```
chrisseto: Please add a new line at the end of this file, you may want to make pycharm add an extra line by default.

chrisseto:
first pass done :racehorse:

waterbutler/providers/dataverse/provider.py
rliebz: Fixed. This was an issue with the way dataset metadata was being serialized.

waterbutler/providers/dataverse/provider.py
rliebz: Done.

waterbutler/providers/dataverse/provider.py
rliebz: Done.

waterbutler/providers/dataverse/provider.py
rliebz: Neat syntax. Done.

waterbutler/providers/dataverse/provider.py
rliebz: The idea was to specifically iterate over possible versions, but now that I look at it, that should be done in `_get_all_data` anyway. Done.

waterbutler/providers/dataverse/metadata.py
rliebz: Done and done.

chrisseto:
If the corresponding osf PR is a hotfix this will need to be upgraded to a hotfix as well.

rliebz:
Moving now.

PR #40, opened 2015-04-23
Hotfix/revisions

rliebz:
Changes
---
- Add Dataverse documentation
- Add file revisions
- Improve version handling

Side effects
--- 
- Requires https://github.com/CenterForOpenScience/osf.io/pull/2572

### DIFF
--- a/waterbutler/providers/dataverse/provider.py
+++ b/waterbutler/providers/dataverse/provider.py
@@ -168,23 +168,18 @@ class DataverseProvider(provider.BaseProvider):
 
     @asyncio.coroutine
     def revisions(self, path, **kwargs):
-        """Get past versions of the request file. Currently only checks
-        'latest' and 'latest-published' versions.
+        """Get past versions of the request file. Orders versions based on
+        `_get_all_data()`
 
         :param str path: The path to a key
         :rtype list:
         """
-        versions = ['latest', 'latest-published']
-        revisions = []
-        for version in versions:
-            metadata = yield from self._get_data(version)
-            revision = next(
-                (item for item in metadata if item['path'] == path), None
-            )
-            if revision:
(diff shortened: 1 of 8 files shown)


waterbutler/providers/dataverse/provider.py
```diff
@@ -7,24 +7,48 @@
...
+        :param dict credentials: Contains `token`
+        :param dict settings: Contains `host`, `doi`, `id`, and `name` of a dataset. Hosts:
+
+            - 'apitest.dataverse.org': Api Test Server
+            - 'dataverse-demo.iq.harvard.edu': Harvard Demo Server
+            - 'dataverse.harvard.edu': Dataverse Production Server **(NO TEST DATA)**
+            - Other
+        """
         super().__init__(auth, credentials, settings)
+        self.BASE_URL = 'https://{0}'.format(self.settings['host'])
```
chrisseto: Can we assume that https will always be supported?

waterbutler/providers/dataverse/provider.py
```diff
@@ -7,24 +7,48 @@
...
-    BASE_URL = 'https://{0}'.format(settings.HOSTNAME)
+    """Provider for Dataverse"""
 
     def __init__(self, auth, credentials, settings):
+        """
+        :param dict auth: Not used
+        :param dict credentials: Contains `token`
+        :param dict settings: Contains `host`, `doi`, `id`, and `name` of a dataset. Hosts:
+
+            - 'apitest.dataverse.org': Api Test Server
```
chrisseto: :+1:

waterbutler/providers/dataverse/provider.py
rliebz: They will redirect from http to https, so I would say that's a safe assumption.

PR #42, opened 2015-05-04
[feature] Download multiple files/directories as Zip

lyndsysimon:


lyndsysimon:
@chrisseto - Since I spoke with you about refactoring `waterbutler/core/streams.py` into a module, would you mind taking a look at this (209239d) and giving your thoughts on how I broke it up?

I'm happy to make any changes you'd like to see in that, just let me know.

I didn't end up using  `__all__` in `waterbutler/core/streams/__init__.py`, since the base classes ended up being in their own file - the only things imported into the module's namespace are the stream classes themselves.

PR #44, opened 2015-05-18
[feature] modular authentication

icereval:


waterbutler/auth/__init__.py
```diff
@@ -0,0 +1 @@
+__import__("pkg_resources").declare_namespace(__name__)
```
chrisseto: New line at the end of file.

PR #46, opened 2015-05-20
Update README.md

ticklemepierce:
Added a few steps to help with the setup of the virtualenv for waterbutler

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,14 +22,17 @@ In order to run waterbutler, you must have a virtualenv created for waterbutler 
 ```bash
 brew install python3 (installs the latest version of python3)
 which python3 (prints the path to where python3 is installed)
+pip install virtualenv (if you don't already have this)
+pip install virtualenvwrapper (if you don't already have this)
 mkvirtualenv --python=(path to python3) waterbutler
-```
-
-
-```bash
-# Make sure that you are using >= python3.3
 pip install -U -r requirements.txt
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,6 +18,14 @@ invoke server
 
 ### startup commands
 
+In order to run waterbutler, you must have a virtualenv created for waterbutler running python3.3 or higher.  This can be created with the following commands:
```
sloria: Might want to note that the below commands requires `virtualenvwrapper`.

sloria: Also, include the command for activating the virtualenv.

lbanner:
Nice @ticklemepierce

README.md
```diff
@@ -18,13 +18,24 @@ invoke server
...
 ```bash
-# Make sure that you are using >= python3.3
+brew install python3 (installs the latest version of python3)
+which python3 (prints the path to where python3 is installed)
+pip install virtualenv (if you don't already have this)
+pip install virtualenvwrapper (if you don't already have this)
+mkvirtualenv --python=(path to python3) waterbutler
```
sloria: Installing python3 with homebrew will symlink python3 to the correct python3 binary, so you should just be able to do:

```python
mkvirtualenv --python=python3 waterbutler
```

sloria: I'll make this change when I merge

PR #52, opened 2015-06-22
feature/file-view-urls

caseyrollins:
#### Purpose
- Adds source_url to the extra field of the file metadata.
- Addresses [#2926](https://github.com/CenterForOpenScience/osf.io/issues/2926)
- Addons:
 - [X] GitHub: file external view link built from current metadata
 - [X] Google Drive: file external view link retrived from Google Drive API
 - [X] Figshare: file external view link built from current metadata
 - [X] Dropbox: file share link generate from Dropbox API
 - [X] Box: file share link generated from Box API
 - [X] S3: unsupported
 - [X] Dataverse: unsupported

### DIFF
--- a/waterbutler/providers/box/metadata.py
+++ b/waterbutler/providers/box/metadata.py
@@ -29,6 +29,10 @@ class BoxFolderMetadata(BaseBoxMetadata, metadata.BaseFolderMetadata):
 
 class BoxFileMetadata(BaseBoxMetadata, metadata.BaseFileMetadata):
 
+    def __init__(self, raw, path_obj, view_url=None):
+        super().__init__(raw, path_obj)
+        self.view_url = view_url
+
     @property
     def name(self):
         return self.raw['name']
@@ -53,6 +57,7 @@ class BoxFileMetadata(BaseBoxMetadata, metadata.BaseFileMetadata):
     def extra(self):
         return {
             'etag': self.raw.get('etag'),
+            'viewUrl': self.view_url
         }
 
     @property
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -301,6 +301,18 @@ class BoxProvider(provider.BaseProvider):
             path
         ).serialized()
 
+    @asyncio.coroutine
+    def get_shared_link(self, path):
+        resp = yield from self.make_request(
+            'PUT',
+            self.build_url('files', path.identifier),
+            data='{"shared_link": {}}',
+            expects=(200, ),
+            throws=exceptions.MetadataError,
+        )
+        data = yield from resp.json()
+        return data['shared_link']['url']
+
     def _assert_child(self, paths, target=None):
         if self.folder == 0:
             return True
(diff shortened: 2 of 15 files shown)


caseyrollins:
@chrisseto Refactored source_url to view_url. Let me know if there are any other changes I should make. :octopus:

chrisseto:
@caseyrollins Looks good to go, I'll do one last review today and then merge it :+1:

waterbutler/providers/github/provider.py
```diff
@@ -511,12 +516,14 @@ def _metadata_file(self, path, ref=None, **kwargs):
...
 
         tree = yield from self._fetch_tree(latest, recursive=True)
+        view_url = self.build_view_url(path.identifier[0], path.path)
 
         try:
             data = next(
                 x for x in tree['tree']
                 if x['path'] == path.path
             )
+
```
chrisseto: Remove this line, there's no reason for the additional whitespace.

waterbutler/providers/box/metadata.py
```diff
@@ -3,9 +3,10 @@
 
 class BaseBoxMetadata(metadata.BaseMetadata):
 
-    def __init__(self, raw, path_obj):
+    def __init__(self, raw, path_obj, view_url=None):
```
chrisseto: This should only be a part of BoxFileMetadata, not the base.
Same for Dropbox metadata.

caseyrollins: Should the same also happen for GitHub and figshare?

waterbutler/providers/box/provider.py
```diff
@@ -336,7 +348,12 @@ def _get_file_meta(self, path, raw=False):
         if not data:
             raise exceptions.NotFoundError(str(path))
 
-        return data if raw else BoxFileMetadata(data, path).serialized()
+        shared_link = data['shared_link']
```
chrisseto: This logic seems a bit confusing, it looks like `view_url` could end up as `None`?

caseyrollins: This actually prevents ```view_url``` from being set to ```None```.  

The ```shared_link``` key is always available in the returned data, but it will be set to ```None``` if a shared link doesn't yet exist for the file. If a shared_link does exist for the file (so if ```shared_link``` is not ```None```), then the ```url``` key will be available and the ```view_url``` can be set to that. Otherwise, the ```view_url``` needs to be created using ```get_shared_link()```.

caseyrollins: Ah, I think the confusing part is that I used ```shared_link.get('url')```, when I can use ```shared_link['url']``` in this case.

chrisseto: In that case don't use `.get` for the url key if it's expected to be there.
It might looks a bit better as
```python
if data['shared_link']:
    # Maybe a comment about the structure of data in the situation
    view_url = data['shared_link']['url']
    ...
```

caseyrollins: Yep making that change now.

PR #54, opened 2015-06-26
[Feature] tornado 4.2

chrisseto:


waterbutler/server/utils.py
```diff
@@ -1,41 +1,2 @@
...
-import tornado.concurrent
-
-
 def make_disposition(filename):
     return 'attachment;filename="{}"'.format(filename.replace('"', '\\"'))
-
-
-# Running Tornado on asyncio's event loop, including 'yield from' support in request handlers
-# https://gist.github.com/BeholdMyGlory/11067131
-def coroutine(func):
```
icereval: Looking at the changes it appears tornado 4.2 now natively supports asyncio coroutines, thus the removal... is this a correct assumption?

chrisseto: Yes, its documented in the change logs of tornado.
`yield from` asyncio coroutines and `yield` tornado coroutines in a method decorated with `tornado.gen.coroutine`. Commit message contains details.

waterbutler/server/settings.py
```diff
@@ -20,7 +20,7 @@
 CORS_ALLOW_ORIGIN = config.get('CORS_ALLOW_ORIGIN', '*')
 
 CHUNK_SIZE = config.get('CHUNK_SIZE', 65536)  # 64KB
-MAX_BUFFER_SIZE = config.get('MAX_BUFFER_SIZE', 1024 * 1024 * 100)  # 100MB
+MAX_BUFFER_SIZE = config.get('MAX_BUFFER_SIZE', int(4.9 * (1024 ** 3)))  # 4.9 GB
```
icereval: This might need to change to `max_body_size`, e.g. https://github.com/tornadoweb/tornado/blob/fdfaf3dffa49479c7461050eacca07bc5ee8d207/tornado/http1connection.py#L107

PR #55, opened 2015-06-26
[Hotfix] Fixed file upload and display issues from un-inited github repos

bdyetton:
Github repos that have not been initialized (have no files in them) do not have lastest sha values, as a fix, if we try to commit to an un-init-ed repo, we first init it by pushing a .gitkeep using the github api to force initialization.
Treebeard also displays an error when displaying files for an un-init-ed repo, this fix makes sure to return an empty file tree from waterbutler to the osf to play nice with treebeards display
This closes OSF issue #3122 and #2976

### DIFF
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -131,11 +131,30 @@ class GitHubProvider(provider.BaseProvider):
         assert self.name is not None
         assert self.email is not None
 
-        exists = yield from self.exists(path)
-        latest_sha = yield from self._get_latest_sha(ref=path.identifier[0])
+        try:
+            exists = yield from self.exists(path)
+        except exceptions.ProviderError as e:
+            if e.data.get('message') == 'Git Repository is empty.':
+                exists = False
+                resp = yield from self.make_request(
+                    'PUT',
+                    self.build_repo_url('contents', '.gitkeep'),
+                    data=json.dumps({
+                        'content': '',
+                        'path': '.gitkeep',
+                        'committer': self.committer,
+                        'branch': path.identifier[0],
+                        'message': 'Initial commit'
+                    }),
+                    expects=(201,),
+                    throws=exceptions.CreateFolderError
+                )
+                data = yield from resp.json()
+                latest_sha = data['commit']['sha']
+        else:
+            latest_sha = yield from self._get_latest_sha(ref=path.identifier[0])
 
         blob = yield from self._create_blob(stream)
-
         tree = yield from self._create_tree({
             'base_tree': latest_sha,
             'tree': [{
@@ -465,7 +484,13 @@ class GitHubProvider(provider.BaseProvider):
         # the operation using the git/trees api which requires a sha.
 
         if not (self._is_sha(path.identifier[0]) or recursive):
-            data = yield from self._fetch_contents(path, ref=path.identifier[0])
+            try:
+                data = yield from self._fetch_contents(path, ref=path.identifier[0])
+            except exceptions.MetadataError as e:
+                if e.data.get('message') == 'This repository is empty.':
+                    data = []
+                else:
+                    raise
 
             ret = []
             for item in data:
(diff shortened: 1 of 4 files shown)


waterbutler/providers/github/provider.py
```diff
@@ -465,7 +480,13 @@ def _metadata_folder(self, path, recursive=False, **kwargs):
         # the operation using the git/trees api which requires a sha.
 
         if not (self._is_sha(path.identifier[0]) or recursive):
-            data = yield from self._fetch_contents(path, ref=path.identifier[0])
+            try:
+                data = yield from self._fetch_contents(path, ref=path.identifier[0])
+            except exceptions.MetadataError as e:
+                if e.data.get('message', None) == "This repository is empty.":
```
chrisseto: No need for the `None` here

waterbutler/providers/github/provider.py
```diff
@@ -16,7 +16,6 @@
 from waterbutler.providers.github.metadata import GitHubFileTreeMetadata
 from waterbutler.providers.github.metadata import GitHubFolderTreeMetadata
 
-
```
chrisseto: No need to remove this line

waterbutler/providers/github/provider.py
```diff
@@ -131,11 +130,27 @@ def upload(self, stream, path, message=None, branch=None, **kwargs):
         assert self.name is not None
         assert self.email is not None
 
-        exists = yield from self.exists(path)
-        latest_sha = yield from self._get_latest_sha(ref=path.identifier[0])
+        try:
+            exists = yield from self.exists(path)
+        except exceptions.ProviderError as e:
+            if e.data.get('message', None) == "Git Repository is empty.":
```
chrisseto: Again, no need for the `None`

waterbutler/providers/github/provider.py
```diff
@@ -131,11 +130,27 @@ def upload(self, stream, path, message=None, branch=None, **kwargs):
...
+                exists = False
+                yield from self.make_request(
```
chrisseto: This should return the `latest_sha`

bdyetton: Sorry what do you mean here? Line 153 returns the latest sha? i just put the try/except around the exists call only

chrisseto: The return from `yield from self.make_request(....)` should contain the value that calling `latest_sha` will return, so there's no need for the additional request.

bdyetton: Except that if there is no exception we still want to call latest sha, i.e. if there is the repo make_request will not be called.

I see your point that there its a double up on requests when the exception is thrown, but by having the _lastest_sha function called 'again' in that rare case, i feel its a little more transparent.

I can move the lastest sha up into the try statement and get the latest sha from the return from the make request if you would rather though.

chrisseto: As github is rather strict about rate limiting we want to make as few requests as possible.
preferably something like:
```python
try:
    exists = yield from self.exists(...)
except ... as e:
   if e.get(message) != '...':
       raise
    data = yield from self.make_request(...)
    latest_sha = data[commit][sha]
else:
    latest_sha = yield from self.get_latest_sha()

Comment (bdyetton) [inline waterbutler/providers/github/provider.py]: Right on!

**Comment by bdyetton** [comment]
Chris, im getting errors from Travis CI in the amazon s3 tests. I havent touched those? should i be worried?

**Comment by chrisseto** [comment]
Nope, they are time dependant and fail on occasion

**Comment by bdyetton** [comment]
Oh good, i got worried for a sec there. 
Thanks for your help mate, hopefully this one is all done now...

File: waterbutler/providers/github/provider.py [hunk 5]
```diff
@@ -131,11 +131,30 @@ def upload(self, stream, path, message=None, branch=None, **kwargs):
         assert self.name is not None
         assert self.email is not None
 
-        exists = yield from self.exists(path)
-        latest_sha = yield from self._get_latest_sha(ref=path.identifier[0])
+        try:
+            exists = yield from self.exists(path)
+        except exceptions.ProviderError as e:
+            if e.data.get('message') == "Git Repository is empty.":
```
icereval: single quotes

bdyetton:
@icereval Ready for merge

chrisseto:
Merged

PR #56, opened 2015-07-08
[feature] Web View Handler

caseyrollins:
#### Purpose
- Adds ```WebViewHandler``` to retrieve/create file external links.
- Allows external links to be created from the file tree page on the OSF.

### DIFF
--- a/tests/providers/box/test_provider.py
+++ b/tests/providers/box/test_provider.py
@@ -466,7 +466,7 @@ class TestWebView:
 
     @async
     @pytest.mark.aiohttpretty
-    def test_get_web_view_link(self, provider, file_metadata):
+    def test_web_view(self, provider, file_metadata):
         item = file_metadata['entries'][0]
         path = WaterButlerPath('/name.txt', _ids=(provider, item['id']))
 
@@ -474,7 +474,7 @@ class TestWebView:
         aiohttpretty.register_json_uri('GET', file_url, body=item)
         aiohttpretty.register_json_uri('PUT', file_url, body=item)
 
-        result = yield from provider.web_view_link(path)
+        result = yield from provider.web_view(path)
         expected = item['shared_link']['url']
         assert result == expected
 
--- a/tests/providers/box/test_provider.py
+++ b/tests/providers/box/test_provider.py
@@ -320,7 +320,7 @@ class TestUpload:
         path.parts[-1]._id = file_metadata['entries'][0]['id']
         expected = BoxFileMetadata(file_metadata['entries'][0], path).serialized()
 
-        assert metadata == expected
+        assert metadata.serialized() == expected
         assert created is True
         assert aiohttpretty.has_call(method='POST', uri=upload_url)
 
@@ -336,7 +336,7 @@ class TestUpload:
 
         expected = BoxFileMetadata(file_metadata['entries'][0], path).serialized()
 
-        assert metadata == expected
+        assert metadata.serialized() == expected
         assert created is False
         assert aiohttpretty.has_call(method='POST', uri=upload_url)
 
@@ -397,10 +397,8 @@ class TestMetadata:
     def test_metadata(self, provider, folder_object_metadata, folder_list_metadata):
         path = WaterButlerPath('/', _ids=(provider.folder, ))
 
-        # object_url = provider.build_url('folders', provider.folder)
         list_url = provider.build_url('folders', provider.folder, 'items', fields='id,name,size,modified_at,etag')
 
-        # aiohttpretty.register_json_uri('GET', object_url, body=folder_object_metadata)
         aiohttpretty.register_json_uri('GET', list_url, body=folder_list_metadata)
 
         result = yield from provider.metadata(path)
@@ -409,9 +407,9 @@ class TestMetadata:
 
         for x in folder_list_metadata['entries']:
             if x['type'] == 'file':
-                expected.append(BoxFileMetadata(x, path.child(x['name'])).serialized())
+                expected.append(BoxFileMetadata(x, path.child(x['name'])))
             else:
-                expected.append(BoxFolderMetadata(x, path.child(x['name'])).serialized())
+                expected.append(BoxFolderMetadata(x, path.child(x['name'])))
 
         assert result == expected
 
@@ -451,7 +449,7 @@ class TestMetadata:
         aiohttpretty.register_json_uri('PUT', file_url, body=item)
 
         result = yield from provider.metadata(path)
-        expected = BoxFileMetadata(item, path).serialized()
+        expected = BoxFileMetadata(item, path)
         assert result == expected
         assert aiohttpretty.has_call(method='GET', uri=file_url)
 
@@ -500,7 +498,7 @@ class TestRevisions:
         result = yield from provider.revisions(path)
 
         expected = [
-            BoxRevision(each).serialized()
+            BoxRevision(each)
             for each in [item] + revisions_list_metadata['entries']
         ]
 
@@ -522,7 +520,7 @@ class TestRevisions:
         aiohttpretty.register_json_uri('GET', revisions_url, body={}, status=403)
 
         result = yield from provider.revisions(path)
-        expected = [BoxRevision(item).serialized()]
+        expected = [BoxRevision(item)]
         assert result == expected
         assert aiohttpretty.has_call(method='GET', uri=file_url)
         assert aiohttpretty.has_call(method='GET', uri=revisions_url)
@@ -588,6 +586,7 @@ class TestCreateFolder:
 
         resp = yield from provider.create_folder(path)
 
-        assert resp['kind'] == 'folder'
-        assert resp['name'] == '50 shades of nope'
-        assert resp['path'] == '/{}/'.format(folder_object_metadata['id'])
+        assert resp.kind == 'folder'
+        assert resp.name == '50 shades of nope'
+        assert resp.path == '/{}/'.format(folder_object_metadata['id'])
+        assert isinstance(resp, BoxFolderMetadata)
--- a/tests/providers/box/test_provider.py
+++ b/tests/providers/box/test_provider.py
@@ -449,10 +449,9 @@ class TestMetadata:
         file_url = provider.build_url('files', path.identifier)
         aiohttpretty.register_json_uri('GET', file_url, body=item)
         aiohttpretty.register_json_uri('PUT', file_url, body=item)
-        view_url = yield from provider.get_shared_link(path)
 
         result = yield from provider.metadata(path)
-        expected = BoxFileMetadata(item, path, view_url=view_url).serialized()
+        expected = BoxFileMetadata(item, path).serialized()
         assert result == expected
         assert aiohttpretty.has_call(method='GET', uri=file_url)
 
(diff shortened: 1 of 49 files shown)


caseyrollins:
@chrisseto Ready for review! :elephant:

waterbutler/server/app.py
```diff
@@ -24,6 +25,7 @@ def make_app(debug):
             (r'/status', status.StatusHandler),
             (r'/ops/copy', copy.CopyHandler),
             (r'/ops/move', move.MoveHandler),
+            (r'/externalView', external.ExternalViewHandler),
```
chrisseto: I haven't decided on a different name but the end point should be all lowercase.

waterbutler/core/provider.py
```diff
@@ -399,6 +399,9 @@ def validate_path(self, path, **kwargs):
     def revisions(self, **kwargs):
         return []  # TODO Raise 405 by default h/t @rliebz
 
+    def external_link(self, **kwargs):
```
chrisseto: This method should be abstract or throw a 409 error

waterbutler/providers/dropbox/provider.py
```diff
@@ -227,10 +216,21 @@ def metadata(self, path, **kwargs):
                     ret.append(DropboxFileMetadata(item, self.folder).serialized())
             return ret
 
-        view_url = yield from self.get_share_url(path)
+        view_url = yield from self.external_link(path)
```
chrisseto: No need to include view_url with metadata if it has its own endpoint

tests/providers/box/test_provider.py
```diff
@@ -449,7 +449,7 @@ def test_metadata_nested(self, provider, file_metadata):
         file_url = provider.build_url('files', path.identifier)
         aiohttpretty.register_json_uri('GET', file_url, body=item)
         aiohttpretty.register_json_uri('PUT', file_url, body=item)
-        view_url = yield from provider.get_shared_link(path)
+        view_url = yield from provider.external_link(path)
```
chrisseto: Write new tests for each provider's `external_link` function.

waterbutler/providers/box/provider.py
```diff
@@ -352,7 +352,7 @@ def _get_file_meta(self, path, raw=False):
             # 'shared_link' key can be None if a shared link for the file does not already exist
             view_url = data['shared_link']['url']
         else:
-            view_url = yield from self.get_shared_link(path)
+            view_url = yield from self.external_link(path)
```
chrisseto: No need for this to be included in metadata.

waterbutler/core/provider.py
chrisseto: Also please write out a docstring

waterbutler/server/handlers/external.py
```diff
@@ -0,0 +1,16 @@
...
+
+    ACTION_MAP = {
+        'GET': 'externalview',
+    }
+
+    @tornado.gen.coroutine
+    def get(self):
+        """Get or create external link for a file"""
+        result = yield from self.provider.external_link(**self.arguments)
+        self.write({'data': result})
```
chrisseto: I would consider making this a redirect rather than return the string.
I'm not 100% sold on either way.

waterbutler/server/app.py
chrisseto: Stealing from google drives API, which is  `webViewLink`, how would you feel about `webview`?

waterbutler/server/handlers/external.py
chrisseto: You could also do both by checking the accept header.
If the client accepts HTML do a redirect otherwise return it in JSON.

waterbutler/server/app.py
caseyrollins: I like it :+1:

waterbutler/providers/figshare/provider.py
```diff
@@ -9,7 +9,7 @@
 from waterbutler.core import provider
 from waterbutler.core import exceptions
 from waterbutler.core.path import WaterButlerPath
-
+from waterbutler.core.provider import build_url
```
chrisseto: this method is already imported. Use `provider.build_url`.

tests/providers/osfstorage/test_provider.py
```diff
@@ -303,3 +304,12 @@ def test_upload_and_tasks(self, monkeypatch, provider_and_mock, file_stream, cre
...
         inner_provider.move.assert_called_once_with(inner_provider, WaterButlerPath('/uniquepath'), WaterButlerPath('/' + file_stream.writers['sha256'].hexdigest))
+
+
+class TestWebView:
+
+    @async
+    def test_get_web_view_link(self, provider, mock_path):
+        path = mock_path
+        with pytest.raises(exceptions.WebViewError):
+            yield from provider.web_view_link(path)
```
chrisseto: Add a newline at the end of this file.

waterbutler/providers/box/provider.py
```diff
@@ -253,6 +253,18 @@ def metadata(self, path, raw=False, folder=False, **kwargs):
         return (yield from self._get_folder_meta(path, raw=raw, folder=folder))
 
     @asyncio.coroutine
+    def web_view_link(self, path, **kwargs):
+        resp = yield from self.make_request(
```
chrisseto: Could you check the behavior, for implemented providers, of asking for the `web_view_link` of a folder

waterbutler/providers/filesystem/provider.py
```diff
@@ -102,6 +102,10 @@ def metadata(self, path, **kwargs):
             metadata = self._metadata_file(path)
             return FileSystemFileMetadata(metadata, self.folder)
 
+    @asyncio.coroutine
+    def web_view_link(self, path, **kwargs):
+        raise exceptions.WebViewError(str(path))
```
chrisseto: Rather than `str(path)` this should be a informative error message.
You may want to create a subclass of `WebViewError` for not supported.
Alternatively you could create a generic `UnsupportedError(ProviderError):`.

Either way the response code needs to be `409` and the response should state something about not supported.
The message itself can be genric.

waterbutler/server/app.py
```diff
@@ -8,6 +8,7 @@
 from waterbutler.core.utils import AioSentryClient
 from waterbutler.server.handlers import crud
 from waterbutler.server.handlers import copy
+from waterbutler.server.handlers import web_view
```
chrisseto: I'm not a huge fan of underscores in module names, maybe just `view.WebViewHandler`?
Follow your :headphones:

waterbutler/core/provider.py
```diff
@@ -427,6 +427,11 @@ def validate_path(self, path, **kwargs):
     def revisions(self, **kwargs):
         return []  # TODO Raise 405 by default h/t @rliebz
 
+    @abc.abstractmethod
+    def web_view_link(self, **kwargs):
```
chrisseto: I'm not a huge fan of `web_view_link`, maybe just `web_view`?
Follow your :8ball:

waterbutler/core/provider.py
```diff
@@ -427,6 +427,11 @@ def validate_path(self, path, **kwargs):
...
+        """Returns a link to the provider's view of a file."""
+        return None
```
chrisseto: `raise NotImplementedError` here

chrisseto:
2nd pass done, looking good :monkey_face:.
Couple small changes need to be made, otherwise I'm just nitpicky about naming.

waterbutler/providers/box/provider.py
caseyrollins: Dropbox, Google Drive, and GitHub will all correctly link to an appropriate folder view. For Box and figshare, should I add a check in ```web_view_link``` that the given path is not a folder?

chrisseto: Do they show an appropriate error page?

caseyrollins: Figshare completely breaks since I'm directly building the URL. I'm looking into it now to see if there is a persistent URL I can build for a fileset, or if I'll just need to raise an exception if the given path isn't a file. 

For Box, I can change ```web_view``` to make the appropriate request, like so:
```    @asyncio.coroutine
    def web_view(self, path, **kwargs):
        resp = yield from self.make_request(
                'PUT',
                self.build_url('files' if path.is_file else 'folders', path.identifier),
                data='{"shared_link": {}}',
                expects=(200, ),
                throws=exceptions.WebViewError,
        )
        data = yield from resp.json()
        return data['shared_link']['url']

**Comment by caseyrollins** [comment]
@chrisseto Made all of those changes, so this is ready for a final pass :tropical_fish:

File: waterbutler/core/exceptions.py [hunk 14]
```diff
@@ -81,6 +81,14 @@ class RevisionsError(ProviderError):
     pass
 
 
+class WebViewError(ProviderError):
+    pass
+
+
+class UnsupportedError(ProviderError):
+    pass
```
chrisseto: This should force the status code to be 405. See `NotFoundError` for an example.

waterbutler/core/provider.py
```diff
@@ -427,6 +427,11 @@ def validate_path(self, path, **kwargs):
     def revisions(self, **kwargs):
         return []  # TODO Raise 405 by default h/t @rliebz
 
+    @asyncio.coroutine
+    def web_view(self, path):
```
chrisseto: This should just take `**kwargs`

waterbutler/core/provider.py
```diff
@@ -427,6 +427,11 @@ def validate_path(self, path, **kwargs):
...
+        """Returns a link to the provider's view of a file."""
```
chrisseto: Can you add `:returns`, `:rtype`, and `:raises` to the docstring?
Take a look the other BaseProvider methods for examples.

waterbutler/providers/box/provider.py
```diff
@@ -253,6 +253,18 @@ def metadata(self, path, raw=False, folder=False, **kwargs):
         return (yield from self._get_folder_meta(path, raw=raw, folder=folder))
 
     @asyncio.coroutine
+    def web_view(self, path, **kwargs):
+        resp = yield from self.make_request(
+            'PUT',
+            self.build_url('files' if path.is_file else 'folders', path.identifier),
+            data='{"shared_link": {}}',
```
chrisseto: As per [the Box API docs](https://box-content.readme.io/#create-a-shared-link-for-a-file), you may want set `access` to open as the default value could be restrictive. And it may be worth check ahead of time if a link exists as I believe that this will override existing links.

waterbutler/providers/googledrive/provider.py
```diff
@@ -212,6 +212,18 @@ def metadata(self, path, raw=False, **kwargs):
...
+    def web_view(self, path, **kwargs):
+        resp = yield from self.make_request(
+            'GET',
+            self.build_url('files', path.identifier),
+            expects=(200, ),
+            throws=exceptions.MetadataError,
+        )
+
+        data = yield from resp.json()
+        return data['alternateLink']
```
chrisseto: Will this link always be public?

chrisseto:
Last request, could you 2 tests for the handler? 1 for returning the string and one for testing the return of the url.

chrisseto:
Merged https://github.com/CenterForOpenScience/waterbutler/commit/55edda6f475ef6a72c64a0c50d4030e466529f89

PR #60, opened 2015-07-13
[feature] Allow new S3 keys to be encrypted

HarryRybacki:
## Purpose:
Allow users to encrypt (at rest) new keys uploaded to S3.

## Changes:
- Pulls encrypt_uploads bool from settings and modifies PUT header accordingly

## Side Effects:
Files will optionally be encrypted on Amazon's cloud servers.

## Notes:
- Depends on [osf.io sister PR](https://github.com/CenterForOpenScience/osf.io/pull/3533).
- boto's ability to tell if a key is encrypted (on the server) is [currently broken](https://github.com/boto/boto/issues/3269). I am looking into the bug as well as a way to check whether a key is encrypted (on the server) via the rest API.
  - Not an immediate issue now as everything is handled on the AWS side of things.

### DIFF
--- a/waterbutler/providers/s3/metadata.py
+++ b/waterbutler/providers/s3/metadata.py
@@ -46,7 +46,7 @@ class S3FileMetadataHeaders(S3Metadata, metadata.BaseFileMetadata):
     def extra(self):
         return {
             'md5': self.raw['ETAG'].replace('"', ''),
-            'encryption': self.raw['X-AMZ-SERVER-SIDE-ENCRYPTION']
+            'encryption': self.raw.get('X-AMZ-SERVER-SIDE-ENCRYPTION', '')
         }
 
 
--- a/waterbutler/providers/s3/metadata.py
+++ b/waterbutler/providers/s3/metadata.py
@@ -45,7 +45,8 @@ class S3FileMetadataHeaders(S3Metadata, metadata.BaseFileMetadata):
     @property
     def extra(self):
         return {
-            'md5': self.raw['ETAG'].replace('"', '')
+            'md5': self.raw['ETAG'].replace('"', ''),
+            'encryption': self.raw['X-AMZ-SERVER-SIDE-ENCRYPTION']
         }
 
 
--- a/waterbutler/providers/s3/provider.py
+++ b/waterbutler/providers/s3/provider.py
@@ -145,7 +145,11 @@ class S3Provider(provider.BaseProvider):
 
         resp = yield from self.make_request(
             'PUT',
-            self.bucket.new_key(path.path).generate_url(settings.TEMP_URL_SECS, 'PUT'),
+            self.bucket.new_key(path.name).generate_url(
+                settings.TEMP_URL_SECS,
+                'PUT',
+                encrypt_key=self.settings.get('encrypt_uploads', False)
+            ),
             data=stream,
             headers={'Content-Length': str(stream.size)},
             expects=(200, 201, ),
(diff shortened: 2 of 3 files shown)


chrisseto:
Ready for review?

HarryRybacki:
@chrisseto ready for review -- not sure how to test this though

chrisseto:
Take a look at `aiohttpretty` there should be some examples in the code.

HarryRybacki:
@chrisseto Will do, thanks!

waterbutler/providers/s3/provider.py
```diff
@@ -145,7 +145,11 @@ def upload(self, stream, path, conflict='replace', **kwargs):
...
         resp = yield from self.make_request(
             'PUT',
-            self.bucket.new_key(path.path).generate_url(settings.TEMP_URL_SECS, 'PUT'),
+            self.bucket.new_key(path.name).generate_url(
+                settings.TEMP_URL_SECS,
+                'PUT',
+                encrypt_key=self.settings.get('encrypt_uploads', False)
+            ),
             data=stream,
             headers={'Content-Length': str(stream.size)},
```
chrisseto: Don't you need to provide the actual header? Or can it be embedded in the url?

waterbutler/providers/s3/provider.py
```diff
@@ -145,7 +145,11 @@ def upload(self, stream, path, conflict='replace', **kwargs):
 
         resp = yield from self.make_request(
             'PUT',
-            self.bucket.new_key(path.path).generate_url(settings.TEMP_URL_SECS, 'PUT'),
+            self.bucket.new_key(path.name).generate_url(
+                settings.TEMP_URL_SECS,
+                'PUT',
+                encrypt_key=self.settings.get('encrypt_uploads', False)
```
chrisseto: In `__init__` add `self.encrypt = self.setting.get(...)`.
That way you can tell the attributes of the class just be checking `__init__` rather than hunting through the code.

waterbutler/providers/s3/provider.py
HarryRybacki: [Per AWS docs] it has to be a header, (http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html), "You can't enforce whether or not objects are encrypted with SSE-S3 when they are uploaded using pre-signed URLs. This is because the only way you can specify server-side encryption is through the AWS Management Console or through an HTTP request header."

chrisseto: In that case you will have to change this method quite a bit, it only uses signed URLs.

HarryRybacki:
@chrisseto ready for review.

waterbutler/providers/s3/metadata.py
```diff
@@ -45,7 +45,8 @@ def etag(self):
     @property
     def extra(self):
         return {
-            'md5': self.raw['ETAG'].replace('"', '')
+            'md5': self.raw['ETAG'].replace('"', ''),
+            'encryption': self.raw['X-AMZ-SERVER-SIDE-ENCRYPTION']
```
chrisseto: Should this be a `.get`?

HarryRybacki: I'm not sure what you mean

chrisseto: The header `X-AMZ-SERVER-SIDE-ENCRYPTION` may not always be included in the response.
So rather than using direct access `d[...]`  use `.get`
But I defer to you on whether it will always be there or not.

HarryRybacki: Ah, yes it should be wrapped in .get. Will modify

chrisseto:
merged https://github.com/CenterForOpenScience/waterbutler/commit/e5c415cfdbcfd47e252b5747fb4ca9d69dae91f7

PR #65, opened 2015-07-16
[feature] file web views 

caseyrollins:


chrisseto:
It looks like for all providers, that we want to build external links for, their respective metadata objects have all the adequate data to build these urls.
Go ahead and just build the urls in the metadata objects and then the rest of the `web_view` methods can be removed.

waterbutler/providers/github/provider.py
```diff
@@ -92,6 +92,10 @@ def build_repo_url(self, *segments, **query):
         segments = ('repos', self.owner, self.repo) + segments
         return self.build_url(*segments, **query)
 
+    def web_view(self, path):
```
chrisseto: Could you make this a private method (prefix with `_` ) as it is reserved for internal use only.

caseyrollins:
@chrisseto Ready for a final check :octopus:

PR #74, opened 2015-09-23
Fixing bucket name

rafaeldelucena:


chrisseto:
This is actually unrelated to the failing test.
It fails due to the url signature not matching properly as it is a time based signature.
Sometimes the test takes a bit too long and the actual signature will not match the expected one.

PR #76, opened 2015-09-25
Use jwts for OsfAuthHandler

chrisseto:
OSF-4733

### DIFF
--- a/waterbutler/auth/osf/handler.py
+++ b/waterbutler/auth/osf/handler.py
@@ -57,7 +57,7 @@ class OsfAuthHandler(auth.BaseAuthHandler):
             raise exceptions.AuthError(data, code=response.status)
 
         try:
-            return jwt.decode((yield from response.json()), settings.JWT_SECRET, algorithm=settings.JWT_ALGORITHM)['data']
+            return jwt.decode((yield from response.json()), settings.JWT_SECRET, algorithm=settings.JWT_ALGORITHM, options={'require_exp': True})['data']
         except KeyError:
             raise exceptions.AuthError(data, code=response.status)
 
@@ -107,6 +107,6 @@ class OsfAuthHandler(auth.BaseAuthHandler):
             raise exceptions.AuthError(data, code=response.status)
 
         try:
-            return jwt.decode((yield from response.json()), settings.JWT_SECRET, algorithm=settings.JWT_ALGORITHM)['data']
+            return jwt.decode((yield from response.json()), settings.JWT_SECRET, algorithm=settings.JWT_ALGORITHM, option={'require_exp': True})['data']
         except KeyError:
             raise exceptions.AuthError(data, code=response.status)
--- a/waterbutler/auth/osf/handler.py
+++ b/waterbutler/auth/osf/handler.py
@@ -36,12 +36,15 @@ class OsfAuthHandler(auth.BaseAuthHandler):
         if view_only:
             bundle['view_only'] = view_only[0].decode()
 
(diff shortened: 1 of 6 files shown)


waterbutler/auth/osf/handler.py
```diff
@@ -36,7 +38,7 @@ def fetch(self, request, bundle):
         response = yield from aiohttp.request(
             'get',
             settings.API_URL,
-            params=bundle,
+            params={'payload': jwt.encode(bundle, settings.JWT_SECRET, algorithm=settings.JWT_ALGORITHM)},
```
icereval: For added security we might as well set expiration on both sides.

https://github.com/jpadilla/pyjwt#expiration-time-claim

icereval: Should use jwt.encode on the entire json object.

waterbutler/auth/osf/handler.py
```diff
@@ -76,7 +78,7 @@ def get(self, resource, provider, request):
             response = yield from aiohttp.request(
                 'get',
                 settings.API_URL,
-                params=params,
+                params={'payload': jwt.encode(params, settings.JWT_SECRET, algorithm=settings.JWT_ALGORITHM)},
```
icereval: Should use `jwt.encode` on the entire json object.

icereval: For added security set expiration to 30s (https://github.com/jpadilla/pyjwt#expiration-time-claim)

waterbutler/auth/osf/handler.py
```diff
@@ -48,7 +53,7 @@ def fetch(self, request, bundle):
                 data = yield from response.read()
             raise exceptions.AuthError(data, code=response.status)
 
-        return (yield from response.json())
+        return jwt.decode((yield from response.json()), settings.JWT_SECRET, algorithm=settings.JWT_ALGORITHM)
```
icereval: Require exp - https://github.com/jpadilla/pyjwt#requiring-optional-claims

waterbutler/auth/osf/handler.py
icereval: This is very unlikely but I'll at least pose the question. Any concern for key collision, should we nest our data under data? https://github.com/jpadilla/pyjwt#support-of-registered-claim-names

icereval:
First review done :pill:

waterbutler/auth/osf/settings.py
```diff
@@ -5,4 +5,7 @@
 
 config = settings.get('OSF_AUTH_CONFIG', {})
 
+JWT_EXPIRATION = config.get('JWT_EXPIRATION', 15)
+JWT_ALGORITHM = config.get('JWT_ALGORITHM', 'HS256')
+JWT_SECRET = config.get('JWT_SECRET', 'ILiekTrianglesALot')
```
sloria: Should we make this required? Maybe crash if it's not defined? Not sure how this is done elsewhere in WB.

sloria:
Pass finished. Would be good to have some basic tests for this, but this could go in a separate PR.

waterbutler/auth/osf/settings.py
icereval: To simplify local development we've used to a non-crashing default value so user's do not need to have a configuration file.

Thought it certainly wouldn't hurt to issue a warning on important settings.

chrisseto:
Merged via https://github.com/CenterForOpenScience/waterbutler/commit/71341de304a2054fa23708b12205e20a915a9044

PR #77, opened 2015-09-29
Catch rename duplicate

Ghalko:
Returns a 409 error.

Update with a conflict default of warn. This is in the file waterbutler.constants.

### DIFF
--- a/waterbutler/constants.py
+++ b/waterbutler/constants.py
@@ -1,0 +1,3 @@
+"""Constants"""
+
+DEFAULT_CONFLICT = 'warn'
--- a/waterbutler/core/exceptions.py
+++ b/waterbutler/core/exceptions.py
@@ -115,6 +115,15 @@ class FolderNamingConflict(ProviderError):
         )
 
 
+class NamingConflict(ProviderError):
+    def __init__(self, path, name=None):
+        super().__init__(
+            'Cannot complete action: file or folder "{name}" already exists in this location'.format(
+                name=name or path.name
+            ), code=409
+        )
+
+
 class NotFoundError(ProviderError):
     def __init__(self, path):
         super().__init__(
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -324,12 +324,15 @@ class BaseProvider(metaclass=abc.ABCMeta):
         the correct file path to upload to and indicate if that file exists or not
 
         :param WaterbutlerPath path: An object supporting the waterbutler path API
-        :param str conflict: replace or keep
+        :param str conflict: replace, keep, warn
         :rtype: (WaterButlerPath, dict or False)
+        :raises: NamingConflict
         """
         exists = yield from self.exists(path, **kwargs)
-        if not exists or conflict != 'keep':
+        if not exists or conflict == 'replace':
             return path, exists
+        if conflict == 'warn':
+            raise exceptions.NamingConflict(path)
 
         while (yield from self.exists(path.increment_name(), **kwargs)):
             pass
--- a/waterbutler/server/api/v1/provider/movecopy.py
+++ b/waterbutler/server/api/v1/provider/movecopy.py
@@ -7,6 +7,7 @@ from waterbutler.core import exceptions
 from waterbutler.server import settings
 from waterbutler.server.auth import AuthHandler
 from waterbutler.core.utils import make_provider
+from waterbutler.constants import DEFAULT_CONFLICT
 
 auth_handler = AuthHandler(settings.AUTH_HANDLERS)
 
@@ -95,7 +96,7 @@ class MoveCopyMixin:
                     self.path,
                     self.dest_path,
                     rename=self.json.get('rename'),
-                    conflict=self.json.get('conflict', 'replace'),
+                    conflict=self.json.get('conflict', DEFAULT_CONFLICT),
                 )
             )
 


waterbutler/core/provider.py
```diff
@@ -328,9 +328,12 @@ def handle_name_conflict(self, path, conflict='replace', **kwargs):
         :rtype: (WaterButlerPath, dict or False)
         """
         exists = yield from self.exists(path, **kwargs)
-        if not exists or conflict != 'keep':
+        if exists and conflict == 'replace':
+            raise exceptions.NamingConflict(path)
+        elif not exists or conflict != 'keep':
             return path, exists
 
+
```
chrisseto: Remove extra new line

waterbutler/core/provider.py
```diff
@@ -328,9 +328,12 @@ def handle_name_conflict(self, path, conflict='replace', **kwargs):
         :rtype: (WaterButlerPath, dict or False)
         """
         exists = yield from self.exists(path, **kwargs)
-        if not exists or conflict != 'keep':
+        if exists and conflict == 'replace':
```
chrisseto: This case should be for a new value of `conflict`, perhaps `'warn'` or maybe even `None`?

And this would go under the existing if statement

```python
if not exists or conflict != 'keep':
    return ...

if conflict == 'warn':
    raise ...

...
```

Ghalko: @chrisseto The only time that it would get to raise an exception is if something exists and it is keep and warn.

chrisseto:
```python
if not exists or conflict == 'replace':
    return ...

if conflict == 'warn':
    raise ...

...
```

Ghalko:
@chrisseto works on my end too. Tests are passing.

waterbutler/constants.py
```diff
@@ -0,0 +1,3 @@
+"""Constants"""
+
+DEFAULT_CONFLICT = 'warn'
```
abought: +1 for this.

I'd like to see more use of named constants instead of redundant strings... but that's outside scope for this PR.

waterbutler/core/exceptions.py
```diff
@@ -115,6 +115,15 @@ def __init__(self, path, name=None):
         )
 
 
+class NamingConflict(ProviderError):
+    def __init__(self, path, name=None):
+        super().__init__(
+            'Cannot complete action: file or folder "{name}" already exists in this location'.format(
+                name=name or os.path.split(path.path.strip('/'))[1]
```
abought: Could the latter expression be simplified to `os.path.basename`?

Ghalko: "Note that the result of this function is different from the Unix basename program; where basename for '/foo/bar/' returns 'bar', the basename() function returns an empty string ('')."

So a folder name might not show up.

abought: > "Note that the result of this function is different from the Unix basename program; where basename for '/foo/bar/' returns 'bar', the basename() function returns an empty string ('')."

Reasonable point- since this applies to renaming both files and folders, that's ok then.

abought: Be careful inserting user-entered text directly into error messages. Check that this message isn't used in front-end code like Growl or Bootbox, which treat message text as raw HTML. For any usage of this string in `growl`, `bootbox`, or `data-bind=html`  contexts, `$osf.htmlEscape` may be helpful.

abought:
CR1 pass finished: ping me when the error message/ growl box changes are ready and I'll take a look before moving this on. Thanks for your help and diligence on this one. :)

waterbutler/core/exceptions.py
chrisseto: You canjust rely on `name or path.name`, the error that this was copied from contains legacy code.

waterbutler/core/provider.py
```diff
@@ -328,8 +328,10 @@ def handle_name_conflict(self, path, conflict='replace', **kwargs):
         :rtype: (WaterButlerPath, dict or False)
```
chrisseto: Update the docstring to reflect your changes

waterbutler/core/provider.py
```diff
@@ -328,8 +328,10 @@ def handle_name_conflict(self, path, conflict='replace', **kwargs):
...
         """
         exists = yield from self.exists(path, **kwargs)
-        if not exists or conflict != 'keep':
+        if not exists or conflict == 'replace':
             return path, exists
+        if exists and conflict == 'warn':
```
chrisseto: `exists` must be `True` at this point, so there's no need to check

waterbutler/core/exceptions.py
Ghalko: Done.

waterbutler/core/provider.py
Ghalko: Done.

waterbutler/core/provider.py
Ghalko: Done, I think.

chrisseto:
@Ghalko did you push?

waterbutler/core/provider.py
```diff
@@ -324,12 +324,14 @@ def handle_name_conflict(self, path, conflict='replace', **kwargs):
         the correct file path to upload to and indicate if that file exists or not
 
         :param WaterbutlerPath path: An object supporting the waterbutler path API
-        :param str conflict: replace or keep
-        :rtype: (WaterButlerPath, dict or False)
+        :param str conflict: replace, keep, warn
+        :rtype: (WaterButlerPath, dict or False, or raise NamingConflict exception)
```
chrisseto: https://pythonhosted.org/an_example_pypi_project/sphinx.html
> def public_fn_with_sphinxy_docstring(name, state=None):
>    """This function does something.
>
>    :param name: The name to use.
>    :type name: str.
>    :param state: Current state to be in.
>    :type state: bool.
>    :returns:  int -- the return code.
>    :raises: AttributeError, KeyError
>
>    """
>    return 0

Ghalko:
Squashed.

chrisseto:
Cherry picked into develop https://github.com/CenterForOpenScience/waterbutler/commit/b4bfad80b49d6bed5956f097f203fe037c077484

PR #78, opened 2015-10-01
[CR] Update waterbutler to return JSON-API compliant 

felliott:
Via: [#OSF-4667]

WaterButler was sending back responses in its own format, whereas the OSF tries to adhere to JSON-API formatted responses.  This patch updates the v1 endpoints to return JSON-API compliant responses on success.  It does **not** fix the error responses from JSON-API.  

I've also opted to drop the included '?kind=folder' query param from the `new_folder` link.  It's already documented in the BAPI docs as a required parameter, and would probably be more appropriate as a `meta` entry.  If it's still desired, it's a quick patch to put it back.

Cheers,
Fitz

### DIFF
--- a/tests/core/test_metadata.py
+++ b/tests/core/test_metadata.py
@@ -92,3 +92,16 @@ class TestBaseMetadata:
             'modified': 'never',
             'size': 1337,
         }
+
+    def test_file_revision_json_api_serialize(self):
+        file_revision_metadata = utils.MockFileRevisionMetadata()
+        serialized = file_revision_metadata.json_api_serialized()
+
+        assert serialized['id'] == 1
+        assert serialized['type'] == 'file_versions'
+        assert serialized['attributes'] == {
+            'extra': {},
+            'version': 1,
+            'modified': 'never',
+            'versionIdentifier': 'versions',
+        }
--- a/tests/core/test_metadata.py
+++ b/tests/core/test_metadata.py
@@ -1,0 +1,94 @@
+import pytest
+import hashlib
+
+from tests import utils
+from unittest import mock
+from tests.utils import async
+from waterbutler.core import metadata
+from waterbutler.core import exceptions
+
+
+class TestBaseMetadata:
+
+    def test_file_metadata(self):
+        file_metadata = utils.MockFileMetadata()
+
+        assert file_metadata.is_file == True
+        assert file_metadata.is_folder == False
+
+    def test_folder_metadata(self):
+        folder_metadata = utils.MockFolderMetadata()
+
+        assert folder_metadata.is_folder == True
+        assert folder_metadata.is_file == False
+
+    def test_file_json_api_serialize(self):
+        file_metadata = utils.MockFileMetadata()
+        serialized = file_metadata.json_api_serialized('n0d3z')
+        link_suffix = '/v1/resources/n0d3z/providers/mock/Foo.name'
+        etag = hashlib.sha256('{}::{}'.format('mock', 'etag').encode('utf-8')).hexdigest()
+
+        assert serialized['id'] == 'mock/Foo.name'
+        assert serialized['type'] == 'files'
+        assert serialized['attributes'] == {
+            'extra': {},
+            'kind': 'file',
+            'name': 'Foo.name',
+            'path': '/Foo.name',
+            'provider': 'mock',
+            'materialized': '/Foo.name',
+            'etag': etag,
+            'contentType': 'application/octet-stream',
+            'modified': 'never',
+            'size': 1337,
+        }
+        assert serialized['links']['new_folder'] == None
+        assert serialized['links']['move'].endswith(link_suffix)
+        assert serialized['links']['upload'].endswith(link_suffix)
+        assert serialized['links']['download'].endswith(link_suffix)
+        assert serialized['links']['delete'].endswith(link_suffix)
+
+    def test_folder_json_api_serialize(self):
+        folder_metadata = utils.MockFolderMetadata()
+        serialized = folder_metadata.json_api_serialized('n0d3z')
+        link_suffix = '/v1/resources/n0d3z/providers/mock/Bar/'
+        etag = hashlib.sha256('{}::{}'.format('mock', 'etag').encode('utf-8')).hexdigest()
+
+        assert serialized['id'] == 'mock/Bar/'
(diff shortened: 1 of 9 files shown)


waterbutler/core/metadata.py
```diff
@@ -26,6 +31,32 @@ def serialized(self):
...
+            'id': os.path.join(self.provider, self.path),
+            'type': self.kind + 's',
+            'attributes': self.serialized(),
+            'links': self._json_api_links(resource)
+        }
+
+    def _json_api_links(self, resource):
+        url = furl.furl(settings.DOMAIN)
+        segments = ['v1', 'resources', resource, 'providers', self.provider] + self.path.split('/')[1:]
+        url.path.segments.extend([urllib.request.quote(x.encode('utf-8')) for x in segments])
```
chrisseto: `from urllib import quote`

waterbutler/core/metadata.py
```diff
@@ -26,6 +31,32 @@ def serialized(self):
...
+        return {
+            'id': os.path.join(self.provider, self.path),
+            'type': self.kind + 's',
+            'attributes': self.serialized(),
+            'links': self._json_api_links(resource)
+        }
+
+    def _json_api_links(self, resource):
+        url = furl.furl(settings.DOMAIN)
+        segments = ['v1', 'resources', resource, 'providers', self.provider] + self.path.split('/')[1:]
```
chrisseto: `segments = ['v1', 'resources', resource, 'providers', self.provider, self.path.lstrip('/')]`

waterbutler/core/metadata.py
```diff
@@ -26,6 +31,32 @@ def serialized(self):
             'etag': hashlib.sha256('{}::{}'.format(self.provider, self.etag).encode('utf-8')).hexdigest(),
         }
 
+    def json_api_serialized(self, resource):
+        """The JSON API serialization of metadata from WaterButler.
+        .. warning::
+        This method determines the output of API v1
+        """
+        return {
+            'id': os.path.join(self.provider, self.path),
```
chrisseto: `self.path` will always start with a `/`.
Regardless of what you put in front of it in `os.path.join` it will discard everything as the `/` is seen as the "root"
```python
os.path.join(self.provider, self.path) == self.path

os.path.join('test', 'text', '/')  == '/'
```

waterbutler/core/metadata.py
```diff
@@ -26,6 +31,32 @@ def serialized(self):
...
+        }
+
+    def _json_api_links(self, resource):
+        url = furl.furl(settings.DOMAIN)
+        segments = ['v1', 'resources', resource, 'providers', self.provider] + self.path.split('/')[1:]
+        url.path.segments.extend([urllib.request.quote(x.encode('utf-8')) for x in segments])
+        entity_url = url.url
+
+        return {
+            "new_folder": None,
```
chrisseto: Use single quotes

chrisseto:
LGTM, just a couple small things

waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -35,7 +35,7 @@ def get_folder(self):
             return (yield from self.download_folder_as_zip())
 
         data = yield from self.provider.metadata(self.path)
-        return self.write({'data': [x.serialized() for x in data]})
+        return self.write({'data': [x.json_api_serialized(self.resource) for x in data]})
```
chrisseto: `data` elements will be doubly nested here

```JSON
{
    "data": [
        {
             "data" ...
        }
    ]
}
```

waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -98,7 +98,7 @@ def file_metadata(self):
         version = self.get_query_argument('version', default=None) or self.get_query_argument('revision', default=None)
 
         return self.write({
-            'data': (yield from self.provider.metadata(self.path, revision=version)).serialized()
+            'data': (yield from self.provider.metadata(self.path, revision=version)).json_api_serialized(self.resource)
```
chrisseto: See doubly nested `data` elements

waterbutler/core/metadata.py
```diff
@@ -26,6 +29,38 @@ def serialized(self):
             'etag': hashlib.sha256('{}::{}'.format(self.provider, self.etag).encode('utf-8')).hexdigest(),
         }
 
+    def json_api_serialized(self, resource):
+        """The JSON API serialization of metadata from WaterButler.
+        .. warning::
+        This method determines the output of API v1
+        """
+        return {
+            'data': {
```
chrisseto: There's (probably) no need to wrap the value in `data` at this level.

waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -35,11 +35,11 @@ def build_args(self):
         return ({
             'nid': self.resource,  # TODO rename to anything but nid
             'path': self.path,
-            'provider': self.provider.serialized()
+            'provider': self.provider.json_api_serialized(self.resource)
```
chrisseto: `provider.serialized()` is for sending data to celery internally. Please revert this and `self.dest_provider` line below

waterbutler/core/metadata.py
```diff
@@ -26,6 +29,38 @@ def serialized(self):
             'etag': hashlib.sha256('{}::{}'.format(self.provider, self.etag).encode('utf-8')).hexdigest(),
         }
 
+    def json_api_serialized(self, resource):
```
chrisseto: If you don't mind, could you write a couple unit tests for this method?

tests/core/test_metadata.py
```diff
@@ -0,0 +1,107 @@
...
+import hashlib
+
+from tests import utils
+from unittest import mock
+from tests.utils import async
+from waterbutler.core import metadata
+from waterbutler.core import exceptions
+
+
+class TestBaseMetadata:
```
chrisseto: :heart:

waterbutler/core/metadata.py
```diff
@@ -141,6 +175,17 @@ def serialized(self):
             'versionIdentifier': self.version_identifier,
         }
 
+    def json_api_serialized(self):
```
chrisseto: No links here?

PR #79, opened 2015-10-05
[CR] Fix 500s when renaming on S3 and GitHub

felliott:
This branch fixes two separate bugs affecting renaming on GitHub and S3

Fixes: [#OSF-4676]

### DIFF
--- a/waterbutler/server/api/v1/provider/movecopy.py
+++ b/waterbutler/server/api/v1/provider/movecopy.py
@@ -41,7 +41,7 @@ class MoveCopyMixin:
             'nid': self.dest_resource,
             'path': self.dest_path,
             'provider': self.dest_provider.serialized()
-        })
+        }, self.auth['callback_url'], self.auth)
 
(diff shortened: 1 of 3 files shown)


waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -86,7 +86,11 @@ def move_or_copy(self):
             self.dest_path = yield from self.dest_provider.validate_path(self.json['path'])
 
         if not getattr(self.provider, 'can_intra_' + action)(self.dest_provider, self.path):
-            result = yield from getattr(tasks, action).adelay(*self.build_args())
+            result = yield from getattr(tasks, action).adelay(
+                rename=self.json.get('rename'),
```
chrisseto: This looks like invalid syntax?

felliott: How so?  It's working for me, but maybe I'm doing something dumb.

PR #81, opened 2015-10-07
[CR] always 404 on file/folder path mismatch

felliott:
v1 expects folder paths to always end with a slash and file paths to never end with a slash.  However, it wasn't enforcing this and some manually-constructed tests were 500ing as a result.  It was decided
that trying to address folders as file and vice-versa should consistently produce 404s.

[#OSF-4675]

### DIFF
--- a/tests/providers/googledrive/test_provider.py
+++ b/tests/providers/googledrive/test_provider.py
@@ -3,6 +3,7 @@ import pytest
 from tests.utils import async
 
 import io
+from http import client
 
 import aiohttpretty
 
@@ -61,6 +62,101 @@ def settings():
 @pytest.fixture
 def provider(auth, credentials, settings):
     return GoogleDriveProvider(auth, credentials, settings)
+
+
+@pytest.fixture
+def search_for_file_response():
+    return {
+        'items': [
+            {'id': '1234ideclarethumbwar'}
+        ]
+    }
+
+@pytest.fixture
+def actual_file_response():
+    return {
+        'id': '1234ideclarethumbwar',
+        'mimeType': 'text/plain',
+        'title': 'B.txt',
+    }
+
+@pytest.fixture
+def search_for_folder_response():
+    return {
+        'items': [
+            {'id': 'whyis6afraidof7'}
+        ]
+    }
+
+@pytest.fixture
+def actual_folder_response():
+    return {
+        'id': 'whyis6afraidof7',
+        'mimeType': 'application/vnd.google-apps.folder',
+        'title': 'A',
+    }
+
+class TestValidatePath:
+
+    @async
+    @pytest.mark.aiohttpretty
+    def test_validate_v1_path_file(self, provider, search_for_file_response,
+                                           actual_file_response):
+        file_name = 'file.txt'
+        file_id = '1234ideclarethumbwar'
+
+        query_url = provider.build_url(
+            'files', provider.folder['id'], 'children',
+            q="title = '{}'".format(file_name), fields='items(id)'
+        )
+        specific_url = provider.build_url('files', file_id, fields='id,title,mimeType')
+
+        aiohttpretty.register_json_uri('GET', query_url, body=search_for_file_response)
+        aiohttpretty.register_json_uri('GET', specific_url, body=actual_file_response)
+
+        try:
+            wb_path_v1 = yield from provider.validate_v1_path('/' + file_name)
+        except Exception as exc:
+            pytest.fail(str(exc))
+
+        with pytest.raises(exceptions.NotFoundError) as exc:
+            yield from provider.validate_v1_path('/' + file_name + '/')
+
+        assert exc.value.code == client.NOT_FOUND
+
+        wb_path_v0 = yield from provider.validate_path('/' + file_name)
+
+        assert wb_path_v1 == wb_path_v0
+
+    @async
+    @pytest.mark.aiohttpretty
+    def test_validate_v1_path_folder(self, provider, search_for_folder_response,
+                                             actual_folder_response):
+        folder_name = 'foofolder'
+        folder_id = 'whyis6afraidof7'
+
+        query_url = provider.build_url(
+            'files', provider.folder['id'], 'children',
+            q="title = '{}'".format(folder_name), fields='items(id)'
+        )
+        specific_url = provider.build_url('files', folder_id, fields='id,title,mimeType')
+
+        aiohttpretty.register_json_uri('GET', query_url, body=search_for_folder_response)
+        aiohttpretty.register_json_uri('GET', specific_url, body=actual_folder_response)
+
+        try:
+            wb_path_v1 = yield from provider.validate_v1_path('/' + folder_name + '/')
+        except Exception as exc:
+            pytest.fail(str(exc))
+
+        with pytest.raises(exceptions.NotFoundError) as exc:
+            yield from provider.validate_v1_path('/' + folder_name)
+
+        assert exc.value.code == client.NOT_FOUND
+
+        wb_path_v0 = yield from provider.validate_path('/' + folder_name + '/')
+
+        assert wb_path_v1 == wb_path_v0
 
 
 class TestCRUD:
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -24,6 +24,10 @@ class BoxProvider(provider.BaseProvider):
         self.folder = self.settings['folder']
 
     @asyncio.coroutine
+    def validate_v1_path(self, path, **kwargs):
+        return self.validate_path(path, **kwargs)
+
+    @asyncio.coroutine
     def validate_path(self, path, **kwargs):
         if path == '/':
             return WaterButlerPath('/', _ids=[self.folder])
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -25,7 +25,36 @@ class BoxProvider(provider.BaseProvider):
 
     @asyncio.coroutine
     def validate_v1_path(self, path, **kwargs):
-        return self.validate_path(path, **kwargs)
+        if path == '/':
+            return WaterButlerPath('/', _ids=[self.folder])
+
+        obj_id = path.strip('/')
+        files_or_folders = 'folders' if path.endswith('/') else 'files'
+
+        # Box file ids must be a valid base10 number
+        if not obj_id.isdecimal():
+            raise exceptions.NotFoundError(str(path))
+
+        response = yield from self.make_request(
+            'get',
+            self.build_url(files_or_folders, obj_id, fields='id,name,path_collection'),
+            expects=(200, 404,),
+            throws=exceptions.MetadataError,
+        )
+
+        if response.status == 404:
+            raise exceptions.NotFoundError(str(path))
+
+        data = yield from response.json()
+
+        names, ids = zip(*[
+            (x['name'], x['id'])
+            for x in
+            data['path_collection']['entries'] + [data]
+        ])
+        names, ids = ('',) + names[ids.index(self.folder) + 1:], ids[ids.index(self.folder):]
+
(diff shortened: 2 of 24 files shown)


waterbutler/providers/box/provider.py
```diff
@@ -24,6 +24,30 @@ def __init__(self, auth, credentials, settings):
...
+            raise exceptions.NotFoundError(str(path))
+
+        response = yield from self.make_request(
+            'get',
+            self.build_url(files_or_folders, obj_id, fields='id,type,'),
+            expects=(200, 404),
+            throws=exceptions.MetadataError,
+        )
+
+        if response.status in (404, 405):
```
chrisseto: A 405 will never reach here, if the status code is not found in `expects` the exception specified in `throws` is raised.

so you could just do:
```python
        response = yield from self.make_request(
            'get',
            self.build_url(files_or_folders, obj_id, fields='id,type,'),
            expects=(200, 404, 405),
            throws=exceptions.MetadataError,
        )
```
however, throws accepts the status code for the exception constructor so I'm not 100% certain if it will work.

felliott: Herp, derp, I think I meant to delete that 405 but forgot.  Thanks for catching.

Also, I'm adding unit tests for each of these.  I'll update the PR later today.

felliott:
Units tests have been added and this branch is ready for review.

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -48,6 +48,10 @@ def __init__(self, auth, credentials, settings):
         self.use_public = self.settings.get('use_public', True)
 
     @asyncio.coroutine
+    def assert_v1_path_semantics(self, path, **kwargs):
+        return
```
chrisseto: A lot of cloudfiles is not implemented so it's not a huge deal but the semantics are the same as S3's

waterbutler/providers/dataverse/provider.py
```diff
@@ -43,6 +43,10 @@ def build_url(self, path, *segments, **query):
         return super().build_url(*(tuple(path.split('/')) + segments), **query)
 
     @asyncio.coroutine
+    def assert_v1_path_semantics(self, path, **kwargs):
+        return
```
chrisseto: This should probably just ensure that path is `/` or doesn't end with a `/` as dataverse has no concept of folders

felliott:
Ready for code review.  BrianG requested that it return 409 Conflict instead of 404, so I had to add some followup requests to box, s3, and filesystem to figure out if it's bad path semantics or just a typo.

waterbutler/providers/dropbox/provider.py
```diff
@@ -23,6 +23,32 @@ def __init__(self, auth, credentials, settings):
...
+        implicit_folder = path.endswith('/')
+
+        resp = yield from self.make_request(
+            'GET', self.build_url('metadata', 'auto', path),
+            expects=(200, 404),
+            throws=exceptions.MetadataError
+        )
+
+        if resp.status == 404:
+            raise exceptions.NotFoundError(str(path))
```
chrisseto: if `expects` doesn't include `404` it will pass back the 404 to the client.
This would produce a nicer/more consistent error message.

felliott:
Updated.

waterbutler/providers/googledrive/provider.py
```diff
@@ -276,7 +290,7 @@ def _build_upload_url(self, *segments, **query):
     def _serialize_item(self, path, item, raw=False):
         if raw:
             return item
-        if item['mimeType'] == 'application/vnd.google-apps.folder':
+        if item['mimeType'] == self.FOLDER_MIME_TYPE:
```
chrisseto: :+1:

waterbutler/server/api/v1/provider/__init__.py
```diff
@@ -43,6 +43,7 @@ def prepare(self, *args, **kwargs):
 
         self.auth = yield from auth_handler.get(self.resource, provider, self.request)
         self.provider = utils.make_provider(provider, self.auth['auth'], self.auth['credentials'], self.auth['settings'])
+        yield from self.provider.assert_v1_path_semantics(self.path)
```
chrisseto: is there any reason this couldn't be done with:
```python
assert path.endswith('/') == self.path.is_dir, '...'
```

felliott: Do you mean after the call to validate_path()?  IIRC, some of providers were 500ing when given an invalid path.  It was either that or `validate_path` sometimes lied about the dirness of things to appease v0.

chrisseto: `assert_v1_path_semantics` makes a lot of duplicate requests, which I'd like to avoid. The fewer requests the better as it can be very easy to max out a client's API key.

felliott: I agree, but it since `validate_path()` currently has to handle v0 and v1, it will be tough to integrate these.  I took this approach with the idea that once v0 is purged, `assert_v1_path_semantics` can be rolled into validate_path(), and the duplicate requests can be gotten rid of.  Alternatively, we could cache the response for now.  How much work do you think it will be to fully ditch v0?

felliott: Unfortunately, I didn't document a specific example, but I do mention this in my `assert_v1_path_semantics` stub docstring:

```python
       """API v1 requires that requests against folder endpoints always end with a slash, and requests
       against files never end with a slash.  This method checks the provider's metadata for the
       given id and returns a 404 Not Found if the implicit and explicit types don't match.  This
       method will probably duplicate some of the logic in the provider's validate_path method, but
       validate_path must currently accomodate v0 AND v1 semantics.  After v0's retirement, this
       method could be merged back into validate_path.

       :param str path: user-supplied path to validate
       :rtype: :class:`None`
       :raises: :class:`waterbutler.core.exceptions.NotFoundError`
       """
```

felliott:
I've reimplemented this as discussed and added tests to make sure `validate_v1_path` and `validate_path` are returning the same thing.

waterbutler/providers/dropbox/provider.py
```diff
@@ -23,6 +23,26 @@ def __init__(self, auth, credentials, settings):
...
+        resp = yield from self.make_request(
+            'GET', self.build_url('metadata', 'auto', self.folder + path),
+            expects=(200,),
+            throws=exceptions.MetadataError
+        )
+
+        data = yield from resp.json()
+        explicit_folder = data['is_dir']
+        if explicit_folder != implicit_folder:
+            raise exceptions.NotFoundError(str(self.folder + path))
```
chrisseto: You don't want to return `self.folder` here.
`self.folder` is the folder the the owning user has linked to their project, which make this a potential privacy concern.
IE: I link my `Work` folder but the path to it is actually `/Trump/For/Pres/Work/`

waterbutler/providers/filesystem/provider.py
```diff
@@ -23,6 +23,19 @@ def __init__(self, auth, credentials, settings):
...
     @asyncio.coroutine
+    def validate_v1_path(self, path, **kwargs):
+        full_path = self.folder + path
+        if not os.path.exists(full_path):
+            raise exceptions.NotFoundError(str(full_path))
+
+        implicit_folder = path.endswith('/')
+        explicit_folder = os.path.isdir(full_path)
+        if implicit_folder != explicit_folder:
+            raise exceptions.NotFoundError(str(full_path))
```
chrisseto: Not as much of an issue here because this is mostly a debugging class but this is the same concern as with dropbox.

waterbutler/providers/github/provider.py
```diff
@@ -58,6 +58,40 @@ def __init__(self, auth, credentials, settings):
         self.repo = self.settings['repo']
 
     @asyncio.coroutine
+    def validate_v1_path(self, path, **kwargs):
+        if not getattr(self, '_repo', None):
+            self._repo = yield from self._fetch_repo()
+            self.default_branch = self._repo['default_branch']
+
+        branch_ref = kwargs.get('branch') or kwargs.get('ref') or self.default_branch
```
chrisseto: `branch` and `ref` should probably be switched here as `ref` is a more specific resource than `branch`.
Sorry about that :8ball:

felliott: NP.  Fixed for both v1 and v0.

felliott:
Updated.

waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -32,6 +38,11 @@ def validate_post(self):
         except (KeyError, ValueError):
             raise exceptions.InvalidParameters('Content-Length is required', code=411)
 
+    def postvalidate_post(self):
```
chrisseto: Any reason for this method to exist?

waterbutler/providers/s3/provider.py
```diff
@@ -56,6 +56,34 @@ def __init__(self, auth, credentials, settings):
...
+                'GET',
+                self.bucket.generate_url(settings.TEMP_URL_SECS, 'GET'),
+                params={'prefix': path, 'delimiter': '/'},
+                expects=(200, 404),
+                throws=exceptions.MetadataError,
+            )
+        else:
+            resp = yield from self.make_request(
+                'HEAD',
+                self.bucket.new_key(path).generate_url(settings.TEMP_URL_SECS, 'HEAD',),
```
chrisseto: Extra comma after `'HEAD'`

tests/providers/googledrive/test_provider.py
```diff
@@ -63,6 +64,101 @@ def provider(auth, credentials, settings):
...
+        wb_path_v0 = yield from provider.validate_path('/' + file_name)
+
+        assert wb_path_v1 == wb_path_v0
+
+    @async
+    @pytest.mark.aiohttpretty
+    def test_validate_v1_path_folder(self, provider, search_for_folder_response,
+                                             actual_folder_response):
+        folder_name = 'foofolder'
+        folder_id = 'whyis6afraidof7'
```
chrisseto: 111 1000 1001

chrisseto:
Looks good, just a couple small things :honey_pot:

waterbutler/server/api/v1/provider/movecopy.py
felliott: Nah, I just like stub functions.  I'll take it out.

tests/providers/googledrive/test_provider.py
felliott: Nope, it's because getting older is the most intrinsic human fear.  Age is the monster waiting in the closet, and every breath draws us unavoidably nearer, until we must reach out and open the door knowing full well the awful fate that awaits within.  It's a primitive terror, whose unctuous tentacles constrict our rational mind and drive us to acts of madness and self-destruction that only accelerate our descent into its grasp.  And yet, it is our salvation, purging the chaff of mankind from the vibrant, rich core.  And one day, perhaps all too soon, we will be the chaff, the dross, the detritus awaiting the scouring that must claim us to save those who come after.

felliott: oops my bad that was supposed to be part of my prereg cr.

felliott:
Issues addressed, commited, rebased, and pushed.  Thanks, @chrisseto!

tests/providers/googledrive/test_provider.py
chrisseto: sad-thumbs

waterbutler/server/api/v1/provider/__init__.py
```diff
@@ -157,12 +169,13 @@ def _send_hook(self, action):
         else:
             # This is adequate for everything but github
             # If extra can be included it will link to the given sha
+            payload_path = self.target_path if self.target_path is not None else self.path
```
chrisseto: `self.target_path or self.path`

chrisseto:
One final nitpick, otherwise LGTM :tangerine:

felliott:
Nitpick fixed!

PR #83, opened 2015-10-12
[CR] github: don't leave behind source file when moving

felliott:
When moving files from a non-root folder in github, the source file was  being left behind.  This was because the tree entry for the source directory was left in the modified master tree, where it still contained references to the old blob/tree.  The solution was to remove all tree entries from the master tree and to allow github to infer the new tree structure from the path names of the remaining blobs.

I also included a fix for folder move/copy actions returning the wrong status code, since the changes are very closely related.

Fixes: [#OSF-4792], [#OSF-4870]

### DIFF
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -584,7 +584,7 @@ class GitHubProvider(provider.BaseProvider):
         old_commit_tree_sha = branch_data['commit']['commit']['tree']['sha']
 
         tree = yield from self._fetch_tree(old_commit_tree_sha, recursive=True)
-        exists = any(x['path'] == dest_path.path for x in tree['tree'])
+        exists = any(x['path'] == dest_path.path.rstrip('/') for x in tree['tree'])
 
         target, keep = None, []
 
@@ -653,4 +653,4 @@ class GitHubProvider(provider.BaseProvider):
             else:
                 folder.children.append(GitHubFileTreeMetadata(item))
 
-        return folder, exists
+        return folder, not exists
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -577,7 +577,16 @@ class GitHubProvider(provider.BaseProvider):
 
     @asyncio.coroutine
     def _do_intra_move_or_copy(self, src_path, dest_path, is_copy):
-        target, branch = None, src_path.identifier[0]
+
+        # ON PATHS:
+        #   WB and GH use slightly different default conventions for their paths, so we often
+        #   have to munge our WB paths before comparison. Here is a quick overview:
+        #     WB (dirs):  wb_dir.path == 'foo/bar/'     str(wb_dir) == '/foo/bar/'
+        #     WB (file):  wb_file.path = 'foo/bar.txt'  str(wb_file) == '/foo/bar.txt'
+        #     GH (dir):   'foo/bar'
+        #     GH (file):  'foo/bar.txt'
+
+        branch = src_path.identifier[0]
         branch_data = yield from self._fetch_branch(branch)
 
         old_commit_sha = branch_data['commit']['sha']
@@ -586,26 +595,28 @@ class GitHubProvider(provider.BaseProvider):
         tree = yield from self._fetch_tree(old_commit_tree_sha, recursive=True)
         exists = any(x['path'] == dest_path.path.rstrip('/') for x in tree['tree'])
 
-        target, keep = None, []
+        blobs = [
+            item
+            for item in tree['tree']
+            if src_path.is_dir and item['path'].startswith(src_path.path) or
+            src_path.is_file and item['path'] == src_path.path
+        ]
 
-        for item in tree['tree']:
-            if item['path'] == str(src_path).strip('/'):
-                assert target is None, 'Found multiple targets'
-                target = item
-            elif item['path'].startswith(src_path.path):
-                keep.append(item)
-
-        if target is None or (src_path.is_dir and target['type'] != 'tree'):
+        if len(blobs) == 0:
             raise exceptions.NotFoundError(str(src_path))
 
+        if src_path.is_file:
+            assert len(blobs) == 1, 'Found multiple targets'
+
         if is_copy:
-            tree['tree'].append(copy.deepcopy(target))
-        elif src_path.is_dir:
-            for item in keep:
-                tree['tree'].remove(item)
+            tree['tree'].extend([copy.deepcopy(blob) for blob in blobs])
 
-        target['path'] = target['path'].replace(src_path.path.strip('/'), dest_path.path.strip('/'), 1)
+        for blob in blobs:
+            blob['path'] = blob['path'].replace(src_path.path, dest_path.path, 1)
 
+        # github infers tree contents from blob paths
+        # see: http://www.levibotelho.com/development/commit-a-file-with-the-github-api/
+        tree['tree'] = [item for item in tree['tree'] if item['type'] != 'tree']
         new_tree_data = yield from self._create_tree({'tree': tree['tree']})
         new_tree_sha = new_tree_data['sha']
 
@@ -638,7 +649,8 @@ class GitHubProvider(provider.BaseProvider):
         )
 
         if dest_path.is_file:
-            return GitHubFileTreeMetadata(target, commit=commit), not exists
+            assert len(blobs) == 1, 'Destination file should have exactly one candidate'
+            return GitHubFileTreeMetadata(blobs[0], commit=commit), not exists
 
         folder = GitHubFolderTreeMetadata({
             'path': dest_path.path.strip('/')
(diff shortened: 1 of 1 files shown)


felliott:
I don't have proper unit tests for this, b/c it's a lot to mock up, but I do have a Paw session I can run for you to verify.  I am also planning on adding a Runscope regression test for this.

waterbutler/providers/github/provider.py
```diff
@@ -567,34 +567,48 @@ def _update_ref(self, sha, ref='master'):
...
+            elif src_path.is_file and item['path'] == src_path.path:
+                assert len(blobs) == 0, 'Found multiple targets'
+                blobs.append(item)
             elif item['path'].startswith(src_path.path):
-                keep.append(item)
+                blobs.append(item)
 
-        if target is None or (src_path.is_dir and target['type'] != 'tree'):
+        if len(blobs) == 0:
             raise exceptions.NotFoundError(str(src_path))
```
chrisseto: This would mean it is impossible to have an empty folder in git, correct?
Which is true, just wanted to make sure everyone is on the same page.

waterbutler/providers/github/provider.py
```diff
@@ -567,34 +567,48 @@ def _update_ref(self, sha, ref='master'):
...
 
+        # github infers tree contents from blob paths
+        # see: http://www.levibotelho.com/development/commit-a-file-with-the-github-api/
+        tree['tree'][:] = [item for item in tree['tree'] if item['type'] != 'tree']
```
chrisseto: What is the purpose of the `[:]` here? I've never seen that syntax before

waterbutler/providers/github/provider.py
```diff
@@ -628,7 +642,7 @@ def _do_intra_move_or_copy(self, src_path, dest_path, is_copy):
         )
 
         if dest_path.is_file:
-            return GitHubFileTreeMetadata(target, commit=commit), not exists
+            return GitHubFileTreeMetadata(blobs[0], commit=commit), not exists
```
chrisseto: `blobs` should always have a length of 1 here?
If so could you add an assertion?

waterbutler/providers/github/provider.py
```diff
@@ -567,34 +567,48 @@ def _update_ref(self, sha, ref='master'):
...
-            tree['tree'].append(copy.deepcopy(target))
-        elif src_path.is_dir:
-            for item in keep:
-                tree['tree'].remove(item)
+            for blob in blobs:
+                tree['tree'].append(copy.deepcopy(blob))
 
-        target['path'] = target['path'].replace(src_path.path.strip('/'), dest_path.path.strip('/'), 1)
+        for blob in blobs:
+            blob['path'] = blob['path'].replace(src_path.path, dest_path.path, 1)
```
chrisseto: What's the purpose of doing this? it looks like `blobs` is only referenced once after this?

waterbutler/providers/github/provider.py
felliott: Yep.

waterbutler/providers/github/provider.py
felliott: I.... don't know.  It might of been a copy-and-paste artifact that accidentally worked?  I'll delete it.

waterbutler/providers/github/provider.py
felliott: `blobs` are references to the blob entries from the original tree.  By modifying them we modify the tree structure, yes?

waterbutler/providers/github/provider.py
felliott: Will do.

waterbutler/providers/github/provider.py
```diff
@@ -646,11 +661,13 @@ def _do_intra_move_or_copy(self, src_path, dest_path, is_copy):
 
         folder.children = []
 
-        for item in keep:
-            item['path'] = item['path'].replace(src_path.path, dest_path.path, 1)
+        tree = yield from self._fetch_tree(orig_tree['sha'])
```
chrisseto: Any chance we can avoid making this request?

felliott: Done, with the rest of the code simplified in the process.

waterbutler/providers/github/provider.py
```diff
@@ -577,35 +577,43 @@ def _update_ref(self, sha, ref='master'):
...
 
         old_commit_sha = branch_data['commit']['sha']
         old_commit_tree_sha = branch_data['commit']['commit']['tree']['sha']
 
         tree = yield from self._fetch_tree(old_commit_tree_sha, recursive=True)
-        exists = any(x['path'] == dest_path.path for x in tree['tree'])
+        exists = any(x['path'] == dest_path.path.rstrip('/') for x in tree['tree'])
 
-        target, keep = None, []
+        blobs = [item for item in tree['tree'] if item['path'].startswith(src_path.path)]
```
chrisseto: Should this be `rstrip`ed?

waterbutler/providers/github/provider.py
```diff
@@ -646,11 +655,13 @@ def _do_intra_move_or_copy(self, src_path, dest_path, is_copy):
 
         folder.children = []
 
-        for item in keep:
+        for item in blob_copies:
```
chrisseto: Couldn't you just iterate of the `blobs` list as that one already has the paths adjusted?
I remember commenting about this previously and I'm very sorry if the two comments are conflicting.

waterbutler/providers/github/provider.py
felliott: No.  We want to identify all children of the path.  If we rstrip, we could end up pulling out paths with similar names..      mv foo/   could end up moving foobar.txt.

...dang, but now I think there's a bug when handling files.  I might need to add another check.

waterbutler/providers/github/provider.py
felliott: I remember I had a specific reason for doing this, but I'm blanking on what it was.  I think I was worried about having the trees removed, but know that I look at it, that shouldn't happen.  I'll reinvestigate (and add a comment).

waterbutler/providers/github/provider.py
chrisseto: Something like this, maybe?
```python
[
    item
    for item in tree['tree'] 
    if item['path'].startswith(src_path.path) 
    or item[path'] == src_path.path.rstrip('/')
]
```

chrisseto:
Looking good, 2 final comments

felliott:
Updated.  I couldn't remember my rationale for the second comment about `blob_copies`, so I've done away with that.  For the blob extraction, I added an extra conditional to make sure that `mv /foo /A/` doesn't try to move `/foobar` as well.

waterbutler/providers/github/provider.py
```diff
@@ -577,35 +577,46 @@ def _update_ref(self, sha, ref='master'):
...
-                tree['tree'].remove(item)
+            tree['tree'].extend([copy.deepcopy(blob) for blob in blobs])
 
-        target['path'] = target['path'].replace(src_path.path.strip('/'), dest_path.path.strip('/'), 1)
+        for blob in blobs:
+            blob['path'] = blob['path'].replace(src_path.path, dest_path.path, 1)
 
+        # github infers tree contents from blob paths
+        # see: http://www.levibotelho.com/development/commit-a-file-with-the-github-api/
+        tree['tree'] = [item for item in tree['tree'] if item['type'] != 'tree']
```
chrisseto: Have you tried without this line?
Just wondering if github had fixed this issue.

chrisseto:
LGTM
@felliott, are you ready for this to be merged?

cwisecarver:
I just went through the code as a CR2. Nothing jumped out at me but I also couldn't add github to my local osf. I was having some sort of problem around the oauth exchange. I'm happy to take another look if a helpful @felliott or @chrisseto gives me some protips.

waterbutler/providers/github/provider.py
felliott: > In waterbutler/providers/github/provider.py <https://github.com/CenterForOpenScience/waterbutler/pull/83#discussion_r43536590>:
> >  
> > +        # github infers tree contents from blob paths
> > +        # see: http://www.levibotelho.com/development/commit-a-file-with-the-github-api/
> > +        tree['tree'] = [item for item in tree['tree'] if item['type'] != 'tree']
> Have you tried without this line?
> Just wondering if github had fixed this issue.
> 

Just did, and it still fails.  I don’t think this is a bug on their end, just a side effect of the git model

felliott:
> On Oct 30, 2015, at 2:32 PM, Chris Seto <notifications@github.com> wrote:
> 
> LGTM
> @felliott <https://github.com/felliott>, are you ready for this to be merged?
> 
> 

Sure, but let me see if we can get @cwisecarver’s oauth working so he can put it through its paces.  I’ll send him a message on Flowdock.

cwisecarver:
@felliott @chrisseto Everything checks out in practice. Correct status codes coming back from move operations and overwrite operations. Things get moved inside the root, outside the root, from the root to outside the root and vice versa. :+1: for CR2.

chrisseto:
Thanks for the update @cwisecarver, I'm going to hold off until merging until some of QA's time is freed up

PR #85, opened 2015-10-13
[CR] return 409 when trying to create a folder on a fileish path

felliott:
We had been returning 400.  All other actions,kind combos are legit.

Minor change, no JIRA ticket created.

### DIFF
--- a/tests/server/api/v1/test_create_mixin.py
+++ b/tests/server/api/v1/test_create_mixin.py
@@ -58,7 +58,8 @@ class TestValidatePut(BaseCreateMixinTest):
         with pytest.raises(exceptions.InvalidParameters) as e:
             self.mixin.validate_put()
 
-        assert e.value.message == 'Path must end with a / if kind is folder'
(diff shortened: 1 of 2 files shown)


tests/server/api/v1/test_create_mixin.py
```diff
@@ -58,7 +58,7 @@ def test_kind_must_be_folder(self):
         with pytest.raises(exceptions.InvalidParameters) as e:
             self.mixin.validate_put()
 
-        assert e.value.message == 'Path must end with a / if kind is folder'
+        assert e.value.message == 'Path must be a folder (and end with a "/") if trying to create a subfolder'
```
chrisseto: Mind throwing in a quick assert on `code == 409`?

felliott:
Updated.  To test, try performing a `new_upload` action on either a WaterButler file path or a folder path with the trailing slash stripped.

    PUT http://127.0.0.1:7777/v1/resources/{node_id}/providers/{provider_name}/{path_to_file}?kind=folder&name=foo

This had been returning a 400 but should be returning a 409.

PR #86, opened 2015-10-16
[CR] implement recursive delete for S3 folders

felliott:
# Summary

If you try to delete a non-empty folder stored in s3, you will get a successful response (204 No Content), but neither the folder nor its contents will be deleted.

This PR implements recursive delete for the Amazon S3 provider.   Amazon doesn't treat folders as first-class objects, so recursively deleting a folder means deleting all of its child files.  Both the file list action and bulk delete action are limited to 1000 entries, so we must supported making successive requests.

Fixes [#OSF-4881]

# Testing

## Situation:

Create a folder on s3 with at least one file in it.  Try to run a DELETE against that folder's delete action.

## Expected:

Both folder and child file are deleted.

## Actual:

Both folder and child file survive the ordeal.

## Desired:

No survivors.

### DIFF
--- a/waterbutler/providers/s3/provider.py
+++ b/waterbutler/providers/s3/provider.py
@@ -5,6 +5,8 @@ from urllib import parse
 
 import xmltodict
 
+from boto.compat import BytesIO
+from boto.utils import compute_md5
 from boto.s3.connection import S3Connection
 from boto.s3.connection import OrdinaryCallingFormat
 from boto.s3.connection import SubdomainCallingFormat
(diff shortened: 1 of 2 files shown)


waterbutler/providers/s3/provider.py
```diff
@@ -167,12 +169,92 @@ def delete(self, path, **kwargs):
...
+        that folder is completely empty.  To fully delete an occupied folder, we must delete all
+        of the comprising objects.  Amazon provides a bulk delete operation to simplify this.
+        """
+
+        more_to_come = True
+        content_keys = []
+        marker = None
+
+        while more_to_come:
+            query_params = {'prefix': path.path}
```
chrisseto: Any reason to not declare this outside of the while loop?

waterbutler/providers/s3/provider.py
```diff
@@ -167,12 +169,92 @@ def delete(self, path, **kwargs):
...
+
+            if isinstance(contents, dict):
+                contents = [contents]
+
+            content_keys.extend([content['Key'] for content in contents])
+            if len(content_keys) > 0:
+                marker = content_keys[-1]
+
+        while len(content_keys) > 0:
+            key_batch = content_keys[0:1000]
```
chrisseto: this can just be `key_batch = content_keys[:1000]`

chrisseto:
LGTM, any chance you could write a test or 2?

PR #92, opened 2015-11-10
[CR] repair logging for Waterbutler delete/create actions

felliott:
Problem
--

Create and delete actions via WaterButler v1 API are not being logged by the OSF.  Example:

I issue a PUT against the WB v1 API to create a new folder, "pr-example", under osfstorage:

screen shot 2015-11-10 at 6 07 24 pm

... and it succeeds.  But when I visit the node's page, I can see the folder in the file browser, but no entry in the node logs:

screen shot 2015-11-10 at 6 08 13 pm

The same is true for delete actions.

Testing
--

Run create folder, create file, and delete actions against waterbutler.  All actions should show up in the node logs on the OSF side.

Fixes: [#OSF-5148]

### DIFF
--- a/tests/providers/box/test_provider.py
+++ b/tests/providers/box/test_provider.py
@@ -571,3 +571,4 @@ class TestCreateFolder:
         assert resp.name == '50 shades of nope'
         assert resp.path == '/{}/'.format(folder_object_metadata['id'])
         assert isinstance(resp, BoxFolderMetadata)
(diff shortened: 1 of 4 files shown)


felliott:
Also, I spent most of today trying to get the proper mocks set up to unit test these fixes, with no luck.  This would be better tested with Runscope tests anyway.  Once the NodeLogs PR to osf.io is merged, I'll do that.

PR #100, opened 2015-12-01
[CR] consistently handle file/folder naming conflicts

felliott:
When creating a new folder or uploading a new file in a directory via WaterButler, our conflict-handling heuristics are all over the map. Some providers (osfstorage, S3) allow a file and folder with the same name to coexist in a directory. Others do not, but behave inconsistently when encountering a conflict. For those providers, which ever entity is created first should succeed, and the following request should throw a 409 Conflict. 

Fixes: [#OSF-5226]

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -75,6 +75,7 @@ class MockProvider(provider.BaseProvider):
     validate_v1_path = None
     validate_path = None
     revalidate_path = None
+    can_duplicate_names = True
 
     def __init__(self, auth=None, settings=None, creds=None):
         super().__init__(auth or {}, settings or {}, creds or {})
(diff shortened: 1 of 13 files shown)


tests/utils.py
```diff
@@ -118,6 +119,9 @@ def metadata(self, path, throw=None, **kwargs):
     def download(self, path, **kwargs):
         return b''
 
+    def can_duplicate_names(self):
```
chrisseto: Duplicate from line 78?

felliott: I don't think so.  `MockProvider1` doesn't inherit from `MockProvider`.

chrisseto: Ah, sorry the diff was throwing me off.
Should the one on line 78 be callable? or is it just there to satisfy the `abstractmethod`?

felliott: Just there to satisfy `abstractmethod`.

PR #101, opened 2015-12-02
[CR] Added in Dryad Provider

kwierman:
So this is pretty much [felliot's dryad_develop](https://github.com/felliott/waterbutler/tree/feature/dryad_provider) feature branch pulled into my repo.


felliott:
Hey @kweirman, sorry for the delay, I'm just starting to really look at this now.  I can confirm that it works for me as-is, but I'll need to review the code and we'll probably need to make some adjustments. 

Cheers,
Fitz

felliott:
BTW, APIv1 support is a must-have before merging.  We're trying to get away from v0.  As I review the code, I'll look for place where v1 support is missing.

felliott:
Hey @kwierman, it's looking good so far.  I need to stop for now, but I'll get back to it Monday.  In the meantime, can you address the issues I raised above?  It could also use some unit tests in `tests/providers/dryad/`.  Thank you!

Cheers,
Fitz

kwierman:
@felliott, These last three commits should address everything except the unit tests. That'll be finished up soon.

kwierman:
Unit tests Added! In addition, I had a small change that was needed for reliability. Dryad kept on 503'ing on me, which ended up being a great test of waterbutler's connectivity

kwierman:
Also, I added in the can_duplicate_names method. I assume this is a v1 change. Once again, after you check this out, I can squash down these 6 commits to one.

kwierman:
Is there documentation on how to use the aiohttpretty marker in pytests?

felliott:
> Is there documentation on how to use the aiohttpretty marker in pytests?

Not really.  Basically, it captures calls to aiohttp.request and redirects them to its own fake_request method[1].  You’ll need to either update  pyDryad to use aiohttp, or update the provider to make the requests via `waterbutler.core.provider.make_request`, and use pyDryad for response/request formatting.  Most of our providers don’t depend on external libraries.  The exception is s3, which uses boto to do some of its url-building.

Cheers,
Fitz

[1] https://github.com/CenterForOpenScience/aiohttpretty/blob/master/aiohttpretty.py#L87

felliott:
Hey @kwierman,

I'm going to hold off further reviewing for now.  Can you ping me when you get the folder semantics and testing issues fixed?  It's coming along nicely!

Cheers,
Fitz

kwierman:
Hey @felliott ,

I think I have all the changes you mentioned fixed. The major change is that all http requests are performed via `aiohttp` now. Everything I've done should be just in the one commit. Interestingly, there was an odd problem with tests passing on my personal computer, but failing on Travis.

felliott:
Hey @kwierman,

Thanks!  I have to do CR for the OneDrive provider first, but I will hopefully get to this soon.

Cheers,
Fitz

felliott:
One minor detail: you have defined a bunch of additional properties in `BaseDryadMetadata`.   Those that can be pulled out from the raw data should be stored in `extra` (which you've done), but don't need to be properties.  Please delete those.

felliott:
All of the paths have a leading `doi:10.5061` path part.  What is the likelihood that Dryad is going to adopt another doi namespace?  If it's not likely, I would like to get rid of it, as it's repetitive.  If there is a need for it, would it make sense to have that as the root folder?   Also, if we do keep it, could you strip off the `doi:` portion of the prefix?  That serves a double purpose: it makes it possible to directly append the id to the resolver url (http://dx.doi.org/), and it avoids the need to percent-encode the colon in our action links.

felliott:
Since the dryad provider is currently read-only, can you remove the `move`, `delete`, and `upload` links from the JSON-API response.  See `metadata.BaseMetadata._json_api_links` for this.

felliott:
Slight issue in `validate_v1_path`:   with v1 we're trying to be stricter about path semantics.  If a user makes a request to a folder path, but leaves off the trailing slash, we return a 404.  If they request a file and append a trailing slash, we return a 404.  The reason for this is that some providers don't use unique ids for their paths and also allow a file and a folder to have the same name.  That means that a request for '/foo/' could return something different than a request for '/foo'.   I know this isn't an issue for Dryad, but we want to maintain a consistent interface across all our providers.

Summary:

This should return serialized folder metadata:
`GET  http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/dryad.1850/?meta=`

This  (same url but without a trailing slash) should return a 404 Not Found:
`GET http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/dryad.1850?meta=`

This should return serialized file metadata:
`GET  http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/dryad.1850/1?meta=`

This  (same url but *with* a trailing slash) should return a 404 Not Found:
`GET http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/dryad.1850/1/?meta=`

felliott:
Hey @kwierman,

It seems like there's some more path confusion.  With WB, the intention is that all the path parts of the url correspond 1-1 to an entity on the provider.  So given this url from my current dev setup:

`http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/dryad.1850/1?meta=`

I expect that all of the following are distinct resources:

`http://localhost:7777/v1/resources/ek46t/providers/dryad/?meta=`
`http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/?meta=`
`http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/dryad.1850/?meta=`
`http://localhost:7777/v1/resources/ek46t/providers/dryad/doi:10.5061/dryad.1850/1?meta=`

But right now, the first three return the same thing.  If we take out the doi portion of the path, we're left with:

`http://localhost:7777/v1/resources/ek46t/providers/dryad/?meta=`
`http://localhost:7777/v1/resources/ek46t/providers/dryad/dryad.1850/?meta=`
`http://localhost:7777/v1/resources/ek46t/providers/dryad/dryad.1850/1?meta=`

This makes a little more sense, if a node can have multiple dryad packages linked.  Then I would expect the first url to return a list of `DryadPackageMetadata` entities, the second to return a list of `DryadFileMetadata` entities, and the last to return a single `DryadFileMetadata` entity.  However, on the OSF side, the addon seems set up to only allow one Dryad package per node.  Is that correct?  Is that the goal, or are there plans to allow more than one package?  If only one package is allowed, then the package id should also be removed from the WaterButler path, giving us:

`http://localhost:7777/v1/resources/ek46t/providers/dryad/?meta=`
`http://localhost:7777/v1/resources/ek46t/providers/dryad/1?meta=`

where the first returns list of `DryadFileMetadata` entities in the dryad.1850 package, and the second return the individual `DryadFileMetadata` entity.  The package id would come from the auth call to the OSF, and is saved in the `folder` attribute of the provider.  For an example, see the `__init__` of the google drive provider.  gdrive lets you specify a particular folder as the root of your storage.  This config parameter is stored in the OSF and passed to WB when it requests the auth token.  The provider then takes care to set it as the root folder for all paths and requests.

I'm sorry, I know this a lot of new information that's not really documented in WB.  If it's any consolation, I'm using this and the other code reviews as a guide for adding proper dev docs to WB.  Thanks for your patience and let me know if you have any questions.

Cheers,
Fitz

felliott:
Pausing CR on this for now.  Please let me know when you've addressed the above issues.

kwierman:
In regards to the doi prefix issue; yes, and the reason why is that there is currently no internal tracking of packages between Dryad's staging, review, and archive databases. As there will be a future update that will include pushing to these databases, I don't want to exclude the possibility of using `uuid` or `iuid` in place of `doi`, a database designation in place `10.5061` or addressing server in place of `/dryad`.

The tl;dr version of this is that I'm being very careful about how I choose my designations since the second version of this provider will require updates to Dryad itself that won't be made until January/February.

kwierman:
In regards to copy and move issues, I'm still not sure why we had to remove the copy functionality. This obviates OSF's ability to "drag and drop" files between providers. I used the drag and drop feature in my addon demo for Dryad, so dropping this might turn into an issue.

felliott:
> In regards to copy and move issues, I'm still not sure why we had to remove the copy functionality. This obviates OSF's ability to "drag and drop" files between providers. I used the drag and drop feature in my addon demo for Dryad, so dropping this might turn into an issue.

I just went back and looked at this and it finally clicked for me.  I was thinking of move/copy in terms of moving/copying TO dryad and completely forgot the case of copying FROM dryad to another provider.  I’m sorry for being so dense.  You are absolutely right, copying from Dryad to another provider should work, while copying to Dryad should throw a 501.  Moving should throw a 501 in either case.  Please add back the copy code when you have a chance and I’ll review it.

Cheers,
Fitz

felliott:
> On Dec 22, 2015, at 12:56 PM, Kevin Wierman <notifications@github.com> wrote:
> 
> In regards to the doi prefix issue; yes, and the reason why is that there is currently no internal tracking of packages between Dryad's staging, review, and archive databases. As there will be a future update that will include pushing to these databases, I don't want to exclude the possibility of using uuid or iuid in place of doi, a database designation in place 10.5061 or addressing server in place of /dryad.
> 
> The tl;dr version of this is that I'm being very careful about how I choose my designations since the second version of this provider will require updates to Dryad itself that won't be made until January/February.
> 
Do you mean that Dryad the service is changing how they’re addressing the data?  If so, that is a conundrum.  Is there a document up somewhere that outlines their plans?  Thank you!

kwierman:
Hey @felliott ,

All right, now with the refactor to osf.io complete, I can get back to wb. So, there isn't a document regarding their plans, but you can expect them to follow the [SWORD API](http://swordapp.org/) pretty closely.

Originally, I used the full designation as a way of providing some amount of forwards compatibility insurance since the full designation is used in Dryad. I see that this can be handled more arbitrarily now. 

So, for this release, I'll use the simple addresses of /<doi_suffix>/ for packages and /<doi_suffix>/<file_index> for files. Sound good?

P.S. The other path issues should be fixed now.

kwierman:
Hey @felliott , 

I'm having some build problems. I cloned develop into my fork and had travis build. The output is [here](https://travis-ci.org/kwierman/waterbutler/builds/101930470). Am I the only one with this problem?

felliott:
> On Jan 12, 2016, at 3:10 PM, Kevin Wierman <notifications@github.com> wrote:
> 
> Hey @felliott <https://github.com/felliott> ,
> 
> I'm having some build problems. I cloned develop into my fork and had travis build. The output is here <https://travis-ci.org/kwierman/waterbutler/builds/101930470>. Am I the only one with this problem?
> 
> 


Hey @kwierman,

Nope, this is bug in freezegun that was triggered by the most recent release of the cryptography package.  There’s a PR in to fix this, but it hasn’t been accepted yet.  Once it has, I’ll bump the freezegun dependency and travis should start working again.  In the meantime, you can fix your local environment by uninstalling cryptography==1.2 and installing cryptography==1.1.2.

I’ll ping you on this PR once the fix goes in.

Cheers,
Fitz

kwierman:
@felliott Much thanks!

felliott:
@kwierman I've updated develop to depend on a forked version of freezegun, until a new version is released.  The Travis builds should be passing again.

Cheers,
Fitz

kwierman:
@felliott Cool. Looks like everything is building fine atm.

kwierman:
@felliott Hey, I just wanted to check in to see if you are still looking at this PR. I haven't heard from you guys in a while, and I'd like to get whatever resolved before I continue.

Thanks,
Kevin

felliott:
Hey @kwierman,

Sorry for not communicating in so long.  We are still looking at this PR, but we’ve been sidetracked with some other tasks.   I’ll see if I can set aside some time for this soon, possibly next week.  In the meantime, you might check out the `feature/python-3.5` branch.  I’ll be trying to get this merged into develop this week, because it gives up a big speed up.  To convert the dryad branch to 3.5 all that should be necessary is some simple syntax substitutions.  This:

````
@asyncio.coroutine
def foo():
````

becomes:

`async def foo`

and `yield from` becomes `await`.  That should be all that's required.  If you don't have time for it, I can update the PR when I review it.

I'll give you a heads up when 3.5 is merged and when I start reviewing the PR.  I've also made a note to contact you if I can't get to it by next Wednesday.  Take care, and thank you again for your effort!

Cheers,
@felliott

kwierman:
Hey @felliott,

I put the python3.5 changes into it's own branch and associated PR. The last commit here was cleaning up the paths.

kwierman:
I'll cleanup the commits sometime this weekend.

kwierman:
Hey @felliott ,

In this latest commit, I fixed up the paths so that 
~~~ URL
/
~~~
Should get back Dryad information
~~~ URL
/XXXX/
~~~
Should get back information on Dryad package: doi:10.5061/dryad.XXXX and is intolerant of a lack of trailing backslash.
~~~ URL
/XXXX/YY
~~~
Should get back on file YY of Dryad package doi:10.5061/dryad.XXXX/YY and is intolerant of including a trailing backslash.

Hopefully that should clear up the path issues!

kwierman:
Oh also, the python3.5 compatible version should be ready to go [here](https://github.com/kwierman/waterbutler/tree/feature/dryad/python3.5). I wasn't sure whether to issue a PR with `develop` as the target or `develop/feature/python-3.5`.

kwierman:
I'm closing this out in favor of the python3.5 version [here](https://github.com/CenterForOpenScience/waterbutler/pull/130).

PR #103, opened 2015-12-09
 [CR] Update Google Drive provider to use identifiers instead of file paths to reference files.

TomBaxter:
## Purpose
Update Google Drive provider to use identifiers instead of file paths to reference files.

## Changes
* Update metadata.py to use id instead of path
* Update provider.py to use id instead of path
  * def validate_path
  * def revalidate_path
  * def upload
  * def create_folder
  * def _resolve_path_to_ids
* Update api/v1/provider/__init__.py to add googledrive to IDENTIFIER_PATHS tuple
* Update test_metadata.py to reflect changes in metadata.py
* Update test_provider.py to reflect changes in provider.py

## Side Effects - Caution Required
**This change will likely require a migration strategy!**
Data stored in mongo storedfilenode.find({ "provider" : "googledrive" }) may need to be migrated.

### DIFF
--- a/tests/providers/googledrive/test_metadata.py
+++ b/tests/providers/googledrive/test_metadata.py
@@ -23,14 +23,13 @@ def test_file_metadata_drive(basepath):
     parsed = GoogleDriveFileMetadata(item, path)
 
     assert parsed.provider == 'googledrive'
-    assert parsed.id == item['id']
     assert path.name == item['title']
     assert parsed.name == item['title']
     assert parsed.size == item['fileSize']
     assert parsed.modified == item['modifiedDate']
     assert parsed.content_type == item['mimeType']
     assert parsed.extra == {'revisionId': item['version'], 'webView': item['alternateLink']}
-    assert parsed.path == '/' + os.path.join(*[x.raw for x in path.parts])
+    assert parsed.path == '/' + item['id']
     assert parsed.materialized_path == str(path)
 
 
@@ -40,14 +39,13 @@ def test_file_metadata_drive_slashes(basepath):
     parsed = GoogleDriveFileMetadata(item, path)
 
     assert parsed.provider == 'googledrive'
-    assert parsed.id == item['id']
     assert parsed.name == item['title']
     assert parsed.name == path.name
     assert parsed.size == item['fileSize']
     assert parsed.modified == item['modifiedDate']
     assert parsed.content_type == item['mimeType']
     assert parsed.extra == {'revisionId': item['version'], 'webView': item['alternateLink']}
-    assert parsed.path == '/' + os.path.join(*[x.raw for x in path.parts])
+    assert parsed.path == '/' + item['id']
     assert parsed.materialized_path == str(path)
 
 
@@ -66,10 +64,9 @@ def test_folder_metadata():
     parsed = GoogleDriveFolderMetadata(item, path)
 
     assert parsed.provider == 'googledrive'
-    assert parsed.id == item['id']
     assert parsed.name == item['title']
     assert parsed.extra == {'revisionId': item['version']}
-    assert parsed.path == '/' + os.path.join(*[x.raw for x in path.parts]) + '/'
+    assert parsed.path == '/' + item['id'] + '/'
     assert parsed.materialized_path == str(path)
 
 
@@ -79,10 +76,9 @@ def test_folder_metadata_slash():
     parsed = GoogleDriveFolderMetadata(item, path)
 
     assert parsed.provider == 'googledrive'
-    assert parsed.id == item['id']
     assert parsed.name == item['title']
     assert parsed.extra == {'revisionId': item['version']}
-    assert parsed.path == '/' + os.path.join(*[x.raw for x in path.parts]) + '/'
+    assert parsed.path == '/' + item['id'] + '/'
     assert parsed.materialized_path == str(path)
 
 
--- a/tests/providers/googledrive/test_provider.py
+++ b/tests/providers/googledrive/test_provider.py
@@ -88,6 +88,40 @@ def actual_file_response():
     }
 
 @pytest.fixture
+def bogus_file_parent_response():
+    return {
+        'error': {
+            'errors': [
+                {'domain': 'global',
+                 'reason': 'notFound',
+                 'message': 'File not found: notthere',
+                 'locationType': 'other',
+                 'location': 'file',
+                },
+            ],
+            'code': 404,
+            'message': 'File not found: notthere',
+        }
+    }
+
+@pytest.fixture
+def bogus_file_child_response():
+    return {
+        "error": {
+            "errors": [
+                {"domain": "global",
+                 "reason": "notFound",
+                 "message": "File not found: ",
+                 "locationType": "other",
+                 "location": "file"
+                }, 
+            ],
+            "code": 404,
+            "message": "File not found: "
+        }
+    }
+
+@pytest.fixture
 def folder_query_response():
     return {
         'id': 'whyis6afraidof7',
@@ -110,7 +144,9 @@ class TestValidatePath:
     def test_validate_v1_path_file(self, provider,
                                    actual_file_response,
                                    search_for_parent_response,
-                                   file_name_query_response):
+                                   file_name_query_response,
+                                   bogus_file_parent_response,
+                                   bogus_file_child_response):
         file_name = 'file.txt'
         file_id = '1234ideclarethumbwar'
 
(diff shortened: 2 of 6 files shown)


felliott:
Hey @TomBaxter,

This is fortuitous!  I just today started working on the same issue, but from the migrations side.  It's going to take me a little while to get those correct, but I'll have a look through your PR and give you some feedback in the meantime.  If you'd like to start adding the python3.5 syntax, @chrisseto's feature/python35 branch is the place to start.  He's done most of it, but neither of us has time to finish it up right now.  Thanks again!

Cheers,
Fitz

TomBaxter:
Awesome! I would very much appreciate any and all critique. I'll start working on the 3.5 syntax today.

TomBaxter:
I have merged @chrisseto's waterbutler/feature/python35 into current COS waterbutler/develop. Resolved conflicts and updated so that all tests are passing. Next steps are update the remaining references to "yield" syntax and then update feature/gdpathing for python 3.5.
  Question, should I try to keep python35 and gdpathing separate? Or will they be pulled in at the same time? If separate, should I make a separate python35 pull request to COS waterbutler/develop? Or make a pull request to @chrisseto's waterbutler/feature/python35 .

Thanks, Tom

felliott:
> I have merged @chrisseto <https://github.com/chrisseto>'s waterbutler/feature/python35 into current COS waterbutler/develop. Resolved conflicts and updated so that all tests are passing. Next steps are update the remaining references to "yield" syntax and then update feature/gdpathing for python 3.5.
> Question, should I try to keep python35 and gdpathing separate? Or will they be pulled in at the same time? If separate, should I make a separate python35 pull request to COS waterbutler/develop? Or make a pull request to @chrisseto <https://github.com/chrisseto>'s waterbutler/feature/python35 .
> 
> 

Hey @TomBaxter!

First off, sorry for the radio silence.  I was pulled away to work on some other tasks, so I haven’t had a chance to review your GoogleDrive PR yet.  I’ve (mostly) cleared off my plate, so I should get to it late today or tomorrow.

Re: python35.  Please keep that branch separate from the googledrive branch.  I’m not sure which is going to go in first, but we’ll make the adjustments when we get it figured out.  You can submit your PR against the COS WB.

Thanks again for your work on this, and I’ll get back to you soon!

Cheers,
Fitz

felliott:
First, a little thing: `tests/providers/googledrive/test_provider.py` has the word "query" misspelled as "querry" throughout.

felliott:
Please delete the commented out code in `googledrive/metadata.py`.

felliott:
Hey @TomBaxter,

I've read through the diff and it's looking pretty good to me.  I'm going to finish up my OSF-side migrations, so I can run them side-by-side properly.  I'll get back to you if I find anything.  Thanks!

Cheers,
@felliott

TomBaxter:
Nice, thanks! I'll fix the spelling and delete commented code.

TomBaxter:
Spelling updated and comments removed.

felliott:
One bug I found:  when validating a non-existent path through the v1 API, I get a 500 because `_resolve_paths_to_ids` is returning 'reval' instead of throwing a 404 Not Found.

TomBaxter:
Thanks, working on it.

felliott:
Hey @TomBaxter,

I've just finished my first pass at a migration script for the OSF, so I'm taking a closer look at this.  One thing I noticed:  I think your `_resolve_path_to_ids` and @chrisseto's `_resolve_id_to_parts` are intended to do the same thing.  I'm guessing `_resolve_id_to_parts` was unfinished; the try block should be outdented by one stop and the recursive call passes `self` when it shouldn't.  But otherwise, it seems to work the same to me.  I'm still playing with it, but was there a technical reason you wrote your version instead of fixing up Chris's?  Why does your version return the `reval` code?  Are the side effects or corner cases you had to handle?

If there aren't any technical reasons to stick with yours, I think I prefer his version slightly.  Normally, 
recursion gives me anxiety, but in this case I find it more understandable here than the stack-based implementation.  I also like the name a **lot** better, so even if we go with stack-based, I'd like to rename `_resolve_path_to_ids` to `_resolve_id_to_parts` and delete Chris's version.

Sorry for not noticing this earlier. I just assumed that GDrive provided lineage and didn't understand why these were necessary.  That seems like a weird oversight on their part.

Thanks and have a happy new years!

Cheers,
Fitz

felliott:
*Bleargh*.  While trying to figure out if there was a better way of finding lineage, I stumbled on this little gem:  GDrive folders aren't really folders, they're labels.  Which means a file with id `1234` can be both in `/foo/bar/` and `/baz/` at the same time.  So what does `GET $node_id/googledrive/1234?meta=` return?  I'm not sure.  I talked it over with @chrisseto and nothing jumped out at us.  Of course sticking with paths has its own issues: if I delete `/foo/bar/1234`, then `/baz/1234` disappears, too.  Yuck.

Right now, we're leaning towards warning people OSF-side that multi-labels are unsupported or problematic.  I'm going to talk this over with the interface folks, but this PR will be on hold until we decide what to do.

TomBaxter:
Thanks @felliott 

  Yes, they are intended to do the same thing and I agree that @chrisseto 's approach is the more elegant one. I spent some time with it and got stuck on one issue. Any given file or folder can have multiple parents and there needs to be a mechanism for identifying the correct path. As a new developer, I don't doubt there is a solution but I couldn't come up with it in a timely fashion so I reverted to what I know. Other then that issue, I can easily finish up _resolve_id_to_parts. I'll put some more thought into it, please let me know if there is something obvious I'm missing.
  I agree with the naming change as well. I left it as is for two reasons. One, to change as little as possible and two, I anticipated that we might want to take a second look at @chrisseto 's code.
  As to the "reval" return. That isn't for an corner case, it's for uploads or folder creation. Again, it had been my goal to change the code as little as possible. Now that I review the code again though, I see that it is an ugly kludge and I will rework it. The crux of it is that when creating a new file/folder we don't know the new "path"(ID) yet, so the old path based logic didn't work in those cases.
  I'll also dive back into the GoogleDrive API to fully understand the folders as labels issue.

Happy New Year to you both as well, Tom

felliott:
<going to reply out of order here>
> As to the "reval" return. That isn't for an corner case, it's for uploads or folder creation. Again, it had been my goal to change the code as little as possible. Now that I review the code again though, I see that it is an ugly kludge and I will rework it. The crux of it is that when creating a new file/folder we don't know the new "path"(ID) yet, so the old path based logic didn't work in those cases.
> 
That makes sense.  I always forget what `revalidate_path` is for.  I’m going to make sure to document it this time.
> I'll also dive back into the GoogleDrive API to fully understand the folders as labels issue.
> 

As I started replying to the next section, I ended up typing up a summary of the issue below.
> Yes, they are intended to do the same thing and I agree that @chrisseto <https://github.com/chrisseto> 's approach is the more elegant one. I spent some time with it and got stuck on one issue. Any given file or folder can have multiple parents and there needs to be a mechanism for identifying the correct path. As a new developer, I don't doubt there is a solution but I couldn't come up with it in a timely fashion so I reverted to what I know. Other then that issue, I can easily finish up _resolve_id_to_parts. I'll put some more thought into it, please let me know if there is something obvious I'm missing.
> 
If I’m understanding Chris’s code correctly, he just does a depth-first search of the parent hierarchy and returns when it hits the root folder.  So I guess which path it resolves is random, depending on the order that drive returns the parent list.  That said, I don’t think that’s a solvable problem if we are using the id as the path.  We might need to move to using materialized id paths.  Take the following structure:

    /Pictures/  (id: 1A2B)
      /Ducks/   (id: 3C4D)
        platypus.jpg (id: 5E6F)
      /Beavers/ (id: 7G8H)
        platypus.jpg (id: 5E6F)

In this case `/Pictures/Ducks/platypus.jpg` and `/Pictures/Beavers/platypus.jpg` are the same file, both with the id 5E6F.  If we try to `GET /googledrive/5E6F?meta=`  we can get most of the meta, but there’s no way of knowing which `materialized_path` to return.  It’s ambiguous.  We could instead use a materialized id path, so that the path for `/Pictures/Ducks/platypus.jpg` becomes `/1A2B/3C4D/5E6F` and for `/Pictures/Beavers/platypus.jpg` becomes ‘/1A2B/7G8H/5E6F’.  Then, `GET /googledrive/1A2B/7G8H/5E6F` is unambiguous.  But then we have a different problem, where `DELETE /googledrive/1A2B/7G8H/5E6F` deletes both copies of platypus.jpg. Likewise, a rename against one affects the other.  We might be able to pretend that gdrive is a regular filesystem by getting clever with how we do update and delete operations, but this is the design issue that I have to talk over with the interface folks first.  The question is how much danger are we willing to let the user take on, how to we alert them to it, and how important is consistency of interface.

I’ll be talking to them today, and it may result in a significant rewrite of the GDrive provider.  When we’ve made a decision, I’ll let you know, but I wouldn’t spend too much time on this right now.  Also, I totally get where you’re coming from with the approach you took.  An ounce of Get It Done is worth pound of Elegance.  You’ve been doing great work on this.  Thank you!

Cheers,
Fitz

TomBaxter:
Thanks for the comments @felliott ! I appreciate your trying to save me some time, but I was already pretty invested in hunting down a solution and so have spent a little more time on it. I feel like I've done what I can as far as research. As near as I can tell, to use identifiers will require a change to waterbutler core. "create_file" and "upload" will need to be able to return multiple "GoogleDrivePath/WaterButlerPath"s and therefore waterbutler core would need to be able to handle that.
  This may be more then we want to take on at the moment. On the other hand, this seems likely to come up with other providers.
  I'll stop work on this now until I get further direction. Please let me know if there is anything further I can do.

Thanks, Tom

PR #105, opened 2015-12-16
[WIP]Upgrade to Python 3.5

TomBaxter:
## Purpose
Update WaterButler for Python 3.5

## Changes
* This is a merge and update from @chrisseto's feature/python35 branch
* Requirements updated for new versions including Tornado 4.3
* code changes for new Python 3.5 async syntax

## Side Effects 
Requires Python 3.5
Requires Tornado 4.3


TomBaxter:
@felliott I will work further to change all yields to await for new python 3.5 syntax.

  While fixing tests for new async syntax I ran into a bug that I'm not sure where to document. As near as I can figure pytest and aiohttp are playing nice together when using the new async "with as" syntax. My best guess is that some combination of pytest and the anyc "with as" are closing the aiohttp response and then aiohttp tries to close it again producing an exception. See below for details.
  My question is, where should I document something like this? It seems we may want to avoid this construct in the tests until this is resolved.

Thanks, Tom

Raises exception but test passes

with pytest.raises(exceptions.NotFoundError) as exc:
    await provider.validate_v1_path('/' + file_id + '/')
    assert exc.value.code == client.NOT_FOUND

passes without raising exception

try:
    await provider.validate_v1_path('/' + file_id + '/')
except Exception as exc:
    assert exc.code == client.NOT_FOUND


### The error raised

============================= test session starts ==============================
platform linux -- Python 3.5.0, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /home/tkb/work/venv/cos/wb35/waterbutler, inifile: 
plugins: cov-2.2.0, asyncio-0.2.0
collected 3 items 

tests/providers/box/test_provider.py .

=========================== 1 passed in 0.16 seconds ===========================
Exception ignored in: <bound method ClientResponse.__del__ of <ClientResponse(https://api.box.com/2.0/folders/5000948880?fields=id,name,path_collection) [404 None]>
<CIMultiDict()>
>
  File "/home/tkb/.virtualenvs/cos/wb35/lib/python3.5/site-packages/aiohttp/client_reqrep.py", line 554, in __del__
    self._loop.call_exception_handler(context)
AttributeError: 'NoneType' object has no attribute 'call_exception_handler'p

TomBaxter:
@felliott Tornado 4.3 was released November 6th. This new version supports the new python 3.5 async await syntax. Would we consider upgrading? Otherwise I think we will need to stick with the yield from syntax for much of the api/core.

Thanks, Tom

felliott:
> @felliott <https://github.com/felliott> Tornado 4.3 was released November 6th. This new version supports the new python 3.5 async await syntax. Would we consider upgrading? Otherwise I think we will need to stick with the yield from syntax for much of the api/core.
> 

Pinging @chrisseto.  He knows way more about this than me.

Cheers,
Fitz

TomBaxter:
I think I've done all I can to convert to the new async await syntax without upgrading tornado. Here is what's left....

(wb35)tkb:waterbutler$ grep -rc "yield from" . | grep -v ":0$"
./waterbutler/server/auth.py:2
./waterbutler/server/utils.py:1
./waterbutler/server/api/v1/provider/metadata.py:12
./waterbutler/server/api/v1/provider/create.py:2
./waterbutler/server/api/v1/provider/movecopy.py:5
./waterbutler/server/api/v1/provider/__init__.py:17
./waterbutler/server/api/v0/copy.py:3
./waterbutler/server/api/v0/zip.py:2
./waterbutler/server/api/v0/metadata.py:1
./waterbutler/server/api/v0/move.py:3
./waterbutler/server/api/v0/revisions.py:1
./waterbutler/server/api/v0/crud.py:9
./waterbutler/server/api/v0/core.py:11
./waterbutler/tasks/copy.py:2
./waterbutler/tasks/move.py:2
./waterbutler/tasks/core.py:3
./waterbutler/auth/osf/handler.py:6
./tests/server/api/v1/test_create_mixin.py:1

TomBaxter:
On a positive note, the tests run noticeably faster now. So hopefully this results in a performance win for users of WaterButler.

felliott:
> On Dec 18, 2015, at 10:56 AM, Tom Baxter <notifications@github.com> wrote:
> 
> On a positive note, the tests run noticeably faster now. So hopefully this results in a performance win for users of WaterButler.
> 
> 

Awesome!  Thanks, Tom!  I’ll try to take a look at this over the holidays.

Cheers,
Fitz

TomBaxter:
Thanks, happy holidays.
Got my osf.io issues worked out (re-installed.) Initial manual testing looks good. Also noticeably faster.

chrisseto:
@TomBaxter +1 to upgrading tornado

TomBaxter:
I've created a separate branch for the GoogleDrive update to merge into the Python 3.5 PR once that's in. I am now working on upgrading Tornado to 4.2

TomBaxter:
@felliott I've upgraded to tornado 4.3 and updated remaining code "yield from -> await" All tests are passing and I'm working on doing as much manual testing as I can.

@chrisseto I ran into the bytearray issue and saw that you have a PR in at Tornado for it. I also noticed that you have a temporary fix in MFR for the same issue. Would you suggest that I use the same fix in WB? Would it be worth while pursuing the Tornado PR?

Thank you both for your time.

chrisseto:
@TomBaxter I'd say use the fix that's in MFR for the moment. There are other considerations for a PR to tornado.

TomBaxter:
@felliott I think this PR is ready for code review. It is passing all tests and passing all the manual tests I've thrown at it. The manual tests have not been exhaustive so there is room for more manual testing. I'll work on that tomorrow, though I'm not sure if all providers allow access to the public. Some of them I've never heard of, but I'll give it a go.

Thanks, Tom

TomBaxter:
Looks like this isn't ready for review yet. During manual testing I ran into a couple instances of a bug. The stack trace returned was all tornado. It errored while trying to return the underlying error. I'm going to get Raven running to see if it gets me more info. Since all tests are currently passing, I'll try to make a test for this issue once I've tracked it down.

Thanks, Tom

TomBaxter:
TIL that Raven is a Sentry client, ha. So didn't do that. I've gotten a little further into the issue and it looks like it may have to do with...

https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with

So far I've only run into this with intra_move s. Not sure if it is effecting anything else. I'll keep working on it.

felliott:
Heads up, @TomBaxter, we're currently having Travis failures due to a conflict between some libraries.  cryptography 1.2 (just released Friday) revealed a bug in freezegun.  I've reported the issue, and someone else has a PR in to fix it.

TomBaxter:
@felliott  Nice, thanks for catching this. I'll leave it be then. Tests are passing in my Dev Environment. And I've done all the manual testing I can think of. I believe this is ready for review.

Thanks, Tom

felliott:
The rest of the changes in that commit need to be reverted, too.  Those tests will never be reached, because in all of those cases, the `await`-ed call is expected to throw an exception.  When it does, that block exits, but the exception is caught by the `pytest.raises` and saved into the given variable.  We can then test the exception at the same level as the `with`.

TomBaxter:
Ok, thanks. Done.

felliott:
Closing as superseded by #114.

PR #111, opened 2016-01-13
[CR] s3: support periods in bucketnames for all regions

felliott:
A bogus error message in the OSF S3 addon revealed some shenanigans
in how we handle bucket names outside the US (N. Virginia) region.
Amazon permits bucket names that follow subdomain-naming conventions
(labels separated by periods).  WB had been accessing buckets using
the subdomain calling format, which would access a bucket named 'foo'
at `https://foo.s3.amazonaws.com/`.  However, wildcard nested
subdomain certificates are deprecated, and Pythons newer than 2.7.9
will throw a certificate checking error if you try to access them.
IOW a bucket named 'foo' will still work, but a bucket named 'foo.bar'
will try to connect to `https://foo.bar.s3.amazonaws.com/` and
will fail validation.

`boto`, the library WB uses to communicate with S3, supports another
calling format that avoids this issue, the `OrdinaryCallingFormat`.
Under OCF, a call to a bucket named `foo.bar` is directed to
`https://s3.amazonaws.com/foo.bar/`, and so the certificate will pass
validation.  The catch is that buckets in regions outside US
(N. Virginia) require the request to be sent to a region-specific URL.
The OSF doesn't currently track the bucket's region, so WB has to
look it up via the S3 API.  Once we have the region, we can update the
host, and all S3 actions should succeed.

Fixes: [#OSF-5509]

References:
----

* [S3 naming conventions](http://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html#bucketnamingrules)
* [S3 bucket region lookup](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGETlocation.html)

### DIFF
--- a/tests/providers/s3/test_provider.py
+++ b/tests/providers/s3/test_provider.py
@@ -1,6 +1,6 @@
 import pytest
 
-from tests.utils import async
+from tests.utils import async, MockCoroutine
 
 import io
 import base64
@@ -46,7 +46,9 @@ def settings():
 
 @pytest.fixture
 def provider(auth, credentials, settings):
-    return S3Provider(auth, credentials, settings)
+    provider = S3Provider(auth, credentials, settings)
+    provider._check_region = MockCoroutine()
+    return provider
 
 
 @pytest.fixture
@@ -244,6 +246,13 @@ def version_metadata():
         </Version>
     </ListVersionsResult>'''
 
+def location_response(location):
+    return (
+        '<?xml version="1.0" encoding="UTF-8"?>\n'
+        '<LocationConstraint xmlns="http://s3.amazonaws.com/doc/2006-03-01/">'
+        '{}</LocationConstraint>'
+    ).format(location)
+
 def list_objects_response(keys, truncated=False):
     response = '''<?xml version="1.0" encoding="UTF-8"?>
     <ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/">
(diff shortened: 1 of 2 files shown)


tests/providers/s3/test_provider.py
```diff
@@ -285,11 +292,46 @@ def build_folder_params(path):
     return {'prefix': path.path, 'delimiter': '/'}
 
 
+class TestRegionDetection:
+
+    @async
+    @pytest.mark.aiohttpretty
+    def test_region_host(self, provider):
+        tests = (
```
chrisseto: Small thing but you may want to look at pytest's parametrize option

tests/providers/s3/test_provider.py
```diff
@@ -315,6 +357,7 @@ def test_validate_v1_path_file(self, provider, file_metadata):
     @async
     @pytest.mark.aiohttpretty
     def test_validate_v1_path_folder(self, provider, folder_metadata):
+        provider._check_region = MockCoroutine()
```
chrisseto: You want to add this to the provider fixture to avoid typing it out everytime

waterbutler/providers/s3/provider.py
```diff
@@ -414,3 +426,35 @@ def _metadata_folder(self, path):
...
+
+        Manually constructing the connection hostname allows us to use OrdinaryCallingFormat
+        instead of SubdomainCallingFormat, which can break on buckets with periods in their name.
+        """
+        if self.region is None:
+            self.region = yield from self._get_bucket_region()
+            if self.region == 'EU':
+                self.region = 'eu-west-1'
+
+            if self.region != '':
```
chrisseto: When is region an empty string?

waterbutler/providers/s3/provider.py
```diff
@@ -414,3 +426,35 @@ def _metadata_folder(self, path):
...
+    @asyncio.coroutine
+    def _check_region(self):
+        """Lookup the region via bucket name, then update the host to match.
+
+        Manually constructing the connection hostname allows us to use OrdinaryCallingFormat
+        instead of SubdomainCallingFormat, which can break on buckets with periods in their name.
+        """
+        if self.region is None:
+            self.region = yield from self._get_bucket_region()
+            if self.region == 'EU':
```
chrisseto: Is `EU` the only special case?

chrisseto:
@felliott LGTM just a couple thoughts.

PR #116, opened 2016-02-10
[RFR][Feature]Fix logic for setting "created" in def _folder_file_op

TomBaxter:
## Purpose
[OSF-4895](https://openscience.atlassian.net/browse/OSF-4895)
Update so that copy operations that create new folders return HTTP 201
#OSF-4895

## Changes
Update waterbutler/core/provider.py

## Side effects
As this is a change to the base provider, may effect other providers

### DIFF
--- a/waterbutler/providers/s3/provider.py
+++ b/waterbutler/providers/s3/provider.py
@@ -29,6 +29,10 @@ from waterbutler.providers.s3.metadata import S3FileMetadataHeaders
 
 class S3Provider(provider.BaseProvider):
     """Provider for the Amazon's S3
+
+    Quirks:
+        On S3, folders are not first-class objects, but are instead inferred from the names of their children.  A regular DELETE request issued against a folder will not work unless that folder is completely empty.  To fully delete an occupied folder, we must delete all of the comprising objects.  Amazon provides a bulk delete operation to simplify this.
+        A GET prefix query against a non-existant path returns 200
     """
     NAME = 's3'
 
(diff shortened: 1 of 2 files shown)


felliott:
Hey @TomBaxter,

The patch is absolutely correct, but left me wondering why s3 worked when gdrive didn't, since they both use this code path.  Looking into it, it turns out that s3's delete folder method doesn't 404 if `dest_path` is missing.  So applying this patch, while correct, will cause s3 to start behaving incorrectly.  Could you fix s3 and add docstrings to `_file_folder_op` and `delete` in `waterbutler.core.provider` mentioning this issue?  Thank you!

Cheers,
Fitz

waterbutler/providers/s3/provider.py
```diff
@@ -29,6 +29,10 @@
 
 class S3Provider(provider.BaseProvider):
     """Provider for the Amazon's S3
+
+    Quirks:
+        On S3, folders are not first-class objects, but are instead inferred from the names of their children.  A regular DELETE request issued against a folder will not work unless that folder is completely empty.  To fully delete an occupied folder, we must delete all of the comprising objects.  Amazon provides a bulk delete operation to simplify this.
```
felliott: Minor: please wrap lines at 100 chars.

waterbutler/providers/s3/provider.py
```diff
@@ -227,10 +231,17 @@ def delete(self, path, **kwargs):
...
+
+        Called from: func: delete if not path.is_file
+
+        Calls: func: self._check_region
+               func: self.make_request
+               func: self.bucket.generate_url
+
+        :param *ProviderPath path: Path to be deleted
+
+        On S3, folders are not first-class objects, but are instead inferred from the names of their children.  A regular DELETE request issued against a folder will not work unless that folder is completely empty.  To fully delete an occupied folder, we must delete all of the comprising objects.  Amazon provides a bulk delete operation to simplify this.
```
felliott: Ditto on the line wrap.

waterbutler/providers/s3/provider.py
```diff
@@ -29,6 +29,10 @@
 
 class S3Provider(provider.BaseProvider):
     """Provider for the Amazon's S3
+
+    Quirks:
```
felliott: :+1:  I **really** like the idea of a Quirks section.   I might be adding these to all providers soon.

waterbutler/providers/s3/provider.py
```diff
@@ -29,6 +29,10 @@
...
+        On S3, folders are not first-class objects, but are instead inferred from the names of their children.  A regular DELETE request issued against a folder will not work unless that folder is completely empty.  To fully delete an occupied folder, we must delete all of the comprising objects.  Amazon provides a bulk delete operation to simplify this.
+        A GET prefix query against a non-existant path returns 200
```
felliott: Minor: please put another newline between this paragraph and the one before, to set them apart.

felliott:
PR looks good, just needs some minor formatting updates to the docs!  Sending back to In Progress.

waterbutler/providers/s3/provider.py
TomBaxter: Probably won't need it very often, not that many quirks.  ;)

felliott:
Merged into `develop-2`.  This will be merged into `develop` once registrations are sorted out.

PR #117, opened 2016-02-11
[RFR][Feature]Fix handle_name_conflict to handle empty list out of func: exists OSF5701

TomBaxter:
## Purpose: 
Fix issue [OSF-5701](https://openscience.atlassian.net/browse/OSF-5701) causing unintended deletion of folders.

## Change:
Properly handle empty list.

waterbutler/core/provider.py def handle_name_conflict was treating empty folders the same as non-existant folders.

## Aslo fixes
[OSF-5703](https://openscience.atlassian.net/browse/OSF-5703)
[OSF-5704](https://openscience.atlassian.net/browse/OSF-5704)
[OSF-5705](https://openscience.atlassian.net/browse/OSF-5705)

#OSF-5701 #OSF-5703 #OSF-5704 #OSF-5705

### DIFF
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -254,6 +254,8 @@ class BaseProvider(metaclass=abc.ABCMeta):
         :param WaterButlerPath dest_path: The path that is being copied to or into
         :param str rename: The desired name of the resulting path, may be incremented
         :param str conflict: The conflict resolution strategy, replace or keep
+
+        Returns: WaterButlerPath dest_path: The path of the desired result.
         """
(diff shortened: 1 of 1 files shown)


TomBaxter:
While reviewing the next issue I noticed that the call to self.exists should return either "False" or a GoogleDrive[ File | Folder ]Metadata object. Not clear to me how self.metadata(path, **kwargs) can result in an empty list, but that appears to be what's happening.

felliott:
Merged into `develop-2`. This will be merged into `develop` once registrations are sorted out.

PR #119, opened 2016-02-15
[RFR][Feature]Fix logic for handle_name_conflict

TomBaxter:
## Purpose
[OSF-5700](https://openscience.atlassian.net/browse/OSF-5700)
Update so that naming conflicts are handled correctly during move/copy

## Changes
Update waterbutler/core/provider.py

## Side effects
As this is a change to the base provider, will effect all providers
Each conflict checked for will trigger an additional revalidate call

## Also fixes
[OSF-5702](https://openscience.atlassian.net/browse/OSF-5702)
[OSF-5706](https://openscience.atlassian.net/browse/OSF-5706)

#OSF-5700 #OSF-5702 #OSF-5706

### DIFF
--- a/waterbutler/core/path.py
+++ b/waterbutler/core/path.py
@@ -194,7 +194,11 @@ class WaterButlerPath:
 
     @property
     def identifier_path(self):
-        """ Returns the ID formatted as a path for providers that use unique ids """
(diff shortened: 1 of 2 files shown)


felliott:
Merged.

PR #120, opened 2016-02-17
Add FakeAuthHandler for running waterbutler without OSF

aaxelb:
This allows running waterbutler in a dev context without requiring OSF for authentication.

(by Abram Booth, not himicakumar, sorry for the username confusion)


samchrisinger:
As a note to readers-- Abram was planning to write docs, but I distracted him for his last hour and a half of time.

PR #122, opened 2016-02-18
OSF-5505 GDrive "def empty_folder" for use in "def delete" against root

TomBaxter:
## Purpose
[OSF-5505](https://openscience.atlassian.net/browse/OSF-5505)
Catch delete calls against provider root in GoogleDrive to prevent from actually deleting the provider root. Just delete contents of provider root

## Changes
Add def empty_folder for use by def delete in the case of root provider

## Side effects
None

#OSF-5505

### DIFF
--- a/waterbutler/providers/googledrive/provider.py
+++ b/waterbutler/providers/googledrive/provider.py
@@ -209,12 +209,41 @@ class GoogleDriveProvider(provider.BaseProvider):
         if not path.identifier:
             raise exceptions.NotFoundError(str(path))
 
+        if path.is_root:
+            yield from self.empty_folder(path)
+            return
+
         yield from self.make_request(
             'DELETE',
             self.build_url('files', path.identifier),
             expects=(204, ),
             throws=exceptions.DeleteError,
         )
+
+    @asyncio.coroutine
+    def empty_folder(self, path):
+        fid = path.identifier
+        if not fid:
+            raise exceptions.NotFoundError(str(path))
+        resp = yield from self.make_request(
+            'GET',
+            self.build_url('files',
+                           q="'{}' in parents".format(fid),
+                           fields='items(id)'),
+            expects=(200, ),
+            throws=exceptions.MetadataError)
+
+        try:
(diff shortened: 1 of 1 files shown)


waterbutler/providers/googledrive/provider.py
```diff
@@ -206,9 +206,25 @@ def upload(self, stream, path, **kwargs):
...
+            If the WaterButlerPath given is for the provider root path, then
+            the contents of provider root path will be deleted. But not the
+            provider root itself.
+        """
         if not path.identifier:
             raise exceptions.NotFoundError(str(path))
 
+        if path.is_root:
+            yield from self.empty_folder(path)
+            return
```
felliott: This should return a 204 response

waterbutler/providers/googledrive/provider.py
```diff
@@ -216,6 +232,39 @@ def delete(self, path, **kwargs):
...
+    def empty_folder(self, path):
+        """Given a WaterButlerPath, delete all contents of folder
+
+        :param WaterButlerPath: Folder to be emptied
+        :rtype: None
+        :raises: :class:`waterbutler.core.exceptions.NotFoundError`
+        :raises: :class:`waterbutler.core.exceptions.MetadataError`
+        :raises: :class:`waterbutler.core.exceptions.DeleteError`
+        """
+        fid = path.identifier
```
felliott: Please rename `fid` to something more informative like `file_id`.

waterbutler/providers/googledrive/provider.py
TomBaxter: def empty_folder returns 204. I tested on my laptop and waterbutler logs show 204 response. If I'm missing something here, I'm not sure what.

waterbutler/providers/googledrive/provider.py
TomBaxter: Done

felliott:
The scope of the original issue has changed.  This PR is being closed and a new PR will be opened later including this and the other delete-storage-root PRs.

PR #123, opened 2016-02-19
OSF-5503 modify GD provider to use trash instead of deleting.

TomBaxter:
## Purpose
[OSF-5503](https://openscience.atlassian.net/browse/OSF-5503)
Update GoogleDrive provider to use trash instead of deleting files/folders

## Changes
Update def delete to set labels:trashed to true , instead of deleting files/folders
Update all searches to ignore files/folders with  labels:trashed = true

## Side effects
Files/Folders will not be deleted from GoogleDrive, which may or may not be what the user is expecting.
User Doc might need to be updated to remind users that their files/folders are in the trash

#OSF-5503

### DIFF
--- a/tests/providers/googledrive/test_provider.py
+++ b/tests/providers/googledrive/test_provider.py
@@ -6,6 +6,7 @@ import io
 from http import client
 
 import aiohttpretty
+from json import dumps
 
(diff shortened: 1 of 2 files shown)


felliott:
Merged.

PR #124, opened 2016-02-24
OSF-5761 update GD provider to check for name duplication when creating new folder

TomBaxter:
## Purpose
[OSF-5761](https://openscience.atlassian.net/browse/OSF-5761)
Update GoogleDrive provider to check for name duplication when creating folder or uploading a file

## Changes
Update def create_folder to check for existing folder with validate_path
Update def upload to check for existing folder with validate_path

## Side effects
Most calls to create_folder and upload will now require an additional call to Google API.

While fixing a test for upload I had to change the call to validate_path in "def upload" from path.path to path.raw_path . After discussing with Fitz, I opened OSF-5769 to change all calls to validate_path in GoogleDrive provider to use path.raw .

#OSF-5761

### DIFF
--- a/waterbutler/providers/googledrive/provider.py
+++ b/waterbutler/providers/googledrive/provider.py
@@ -265,6 +265,14 @@ class GoogleDriveProvider(provider.BaseProvider):
     @asyncio.coroutine
     def create_folder(self, path, **kwargs):
         GoogleDrivePath.validate_folder(path)
+        orig_path = path
+
+        # Check for duplicate folder names
+        if not path.identifier:
+            try:
+                path = yield from self.validate_v1_path(path.path)
+            except (exceptions.NotFoundError):
+                path = orig_path
 
         if path.identifier:
             raise exceptions.FolderNamingConflict(str(path))
--- a/waterbutler/providers/googledrive/provider.py
+++ b/waterbutler/providers/googledrive/provider.py
@@ -197,7 +197,7 @@ class GoogleDriveProvider(provider.BaseProvider):
         # Check for duplicate file names
         if not path.identifier:
             try:
-                path = yield from self.validate_v1_path(path.raw_path)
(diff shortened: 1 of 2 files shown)


waterbutler/providers/googledrive/provider.py
```diff
@@ -192,6 +192,14 @@ def download(self, path, revision=None, range=None, **kwargs):
     @asyncio.coroutine
     def upload(self, stream, path, **kwargs):
         assert path.is_file
+        orig_path = path
+
+        # Check for duplicate file names
+        if not path.identifier:
+            try:
+                path = yield from self.validate_v1_path(path.raw_path)
```
felliott: `validate_v1_path` shouldn't be used to test for existence inside the action methods.  It could throw an error if given a v0-style path.  Instead, use `exists` or `metadata`.

waterbutler/providers/googledrive/provider.py
```diff
@@ -265,6 +273,14 @@ def revisions(self, path, **kwargs):
     @asyncio.coroutine
     def create_folder(self, path, **kwargs):
         GoogleDrivePath.validate_folder(path)
+        orig_path = path
+
+        # Check for duplicate folder names
+        if not path.identifier:
+            try:
+                path = yield from self.validate_v1_path(path.path)
```
felliott: Ditto.

waterbutler/providers/googledrive/provider.py
felliott: Correction, you can use `validate_path` here as it should work for both v0 and v1 calls.  But `validate_v1_path` isn't guaranteed to work for v0 calls.

felliott:
This was fixed by d45391ddc22ec55baa12e4e63c7ee4660cc1e136 with a different approach.  Closing without merging

PR #128, opened 2016-02-25
[RFR][Feature] OSF-5786 OSFStorage provider root deletes

TomBaxter:
## Purpose
[OSF-5786](https://openscience.atlassian.net/browse/OSF-5786)
Update OSFStorage provider to allow for ...
DELETE http://host:7777/v1/resources/XXXXX/providers/osfstorage/

## Changes
Update def delete to check for path.is_root and call def _delete_folder_contents
Create def _delete_folder_contents to find and delete folder children

## Side effects
None

#OSF-5786


felliott:
The scope of the original issue has changed.  This PR is being closed and a new PR will be opened later including this and the other delete-storage-root PRs.

PR #129, opened 2016-02-26
[feature] throttle max requests to avoid rate limits

icereval:


waterbutler/providers/googledrive/provider.py
```diff
@@ -288,6 +288,10 @@ def create_folder(self, path, **kwargs):
 
         return GoogleDriveFolderMetadata((yield from resp.json()), path)
 
+    @asyncio.coroutine
+    def make_request(self, *args, **kwargs):
+        return (yield from super().make_request(*args, **kwargs))
+
```
felliott: I might be missing something here, but isn't this redundant?

waterbutler/tasks/settings.py
```diff
@@ -42,5 +42,6 @@
     'waterbutler.tasks.move'
 ])
 
+CELERY_ACKS_LATE = True
```
felliott: This doesn't seem to be used anymore.

icereval: Looks like we need to get this turned on in more places http://docs.celeryproject.org/en/latest/faq.html#should-i-use-retry-or-acks-late

waterbutler/providers/googledrive/provider.py
chrisseto: 100% redundant, nice catch!
I think I had originally throttled per provider but later moved the throttle to be global again.

waterbutler/tasks/settings.py
felliott: Derp.  That was me being dumb, thinking it was an unused local config setting.

PR #130, opened 2016-02-29
Feature/dryad/python 3.5

kwierman:
@felliott ,

As promised, the python 3.5 version of the Dryad addon. This passes Travis on my fork, so I'm _hoping_ it works on the main repo. Based off the latest conversations, this should be the final iteration or close to it...


kwierman:
@felliott 

That last commit just replaced the path verifcation with calls to WaterButlerPath. Hopefully it should be a bit more readable.

felliott:
Minor:  can you add a `/docs/provider.dryad.rst` and add it to `/docs/providers.rst`?  That'll make it easier to see the Dryad docs on readthedocs.  If you want to preview your docs:

````
$ pip install -r doc-requirements.txt
$ cd docs
$ make html
$ open _build/html/index.html
````

File: waterbutler/providers/dryad/metadata.py [hunk 1]
```diff
@@ -0,0 +1,131 @@
...
+        return super(BaseDryadMetadata, self).extra.update({
+            'part_of': self._get_element("dcterms:isPartOf")
+        })
+
+
+class DryadPackageMetadata(BaseDryadMetadata, metadata.BaseFolderMetadata):
+
+    @property
+    def path(self):
+        return BaseDryadMetadata.path(self) + '/'
```
felliott: This code throws an error because `.path` is a property, not a method.  I think you want `super(DryadPackageMetadata, self).path + '/'`.

felliott: Actually, that can just be `super().path + '/'`.

kwierman:
@felliott . Done and done. How does it look to you? More detail in the docs?

felliott:
Thanks, @kwierman! The detail is good.  Can you add some markup to the class docstring to set off the code and non-code bits?  [The docstring for WaterButlerPath](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/core/path.py#L66) ([rendered](http://waterbutler.readthedocs.org/en/latest/waterbutler.core.html#waterbutler.core.path.WaterButlerPath)) uses inline and block code markup, as an example.

kwierman:
@felliott , Sure thing. I did notice something earlier when trying to cross reference functions in the doc strings. In the WaterButlerPath module, if you try and reference `:func:WaterButlerPath.path` (I left out the backticks because we're inside github flavoured markdown) from elsewhere inside the same module, it would create a link to the method documentation. However, inside the `dryad.provider` module, I need to use more verbose references. For instance, inside the class docstring, I need to use `:func:waterbutler.providers.dryad.provider.DryadProvider.validate_path` whereas I thought I should be able to use `:func:DryadProvider.validate_path` or even `:func:.validate_path`. I'm guessing this is because the build directory is not the same as the module directory.

Obviously, it's not a huge deal to verbosely reference other methods, but you'll see in the source code that I'm running over the 80 character line limit every time. If you know a better way to clean it up, I'd like to know. Otherwise, the documentation itself is looking pretty good...

felliott:
Sorry, this was closed automatically when the `feature/python-3.5` branch was deleted.  `feature/python-3.5` has been merged into `master` and `develop`.  Can you reopen this against `develop`?

kwierman:
Woops! Sure thing. I need to take off in a few minutes, but I'll get to this when I get back tonight or tomorrow morning.

PR #131, opened 2016-03-01
[RFR][Feature] OSF-5504 Require confirm_delete=1 for root folder deletes

TomBaxter:
## Purpose
[OSF-5504](https://openscience.atlassian.net/browse/OSF-5504)
Update WaterButler to accept...
DELETE http://host:7777/v1/resources/XXXXX/providers/XXXX/?confirm_delete=1
and require it for deletion of root folder

Update providers to allow for
DELETE http://host:7777/v1/resources/XXXXX/providers/osfstorage/

## Changes
Update waterbutler/server/api/v1/provider/__init__.py to pass on confirm_delete=1 for deletes
Update def delete to check for confirm_delete=1 when deleting root folder in the following providers...
google drive
github
dropbox
box
osfstorage
Update def delete to check for path.is_root and call def _delete_folder_contents
Create def _delete_folder_contents to find and delete folder children

## Side effects
None

#OSF-5504

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -141,7 +141,7 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
 ::
 
     Method:        DELETE
-    Query Params:  <none>
+    Query Params:  ?confirm_delete=1 //required for root folder delete only
     Success:       204 No Content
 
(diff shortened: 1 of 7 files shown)


felliott:
Merged.

PR #134, opened 2016-03-03
[WIP][Feature] OSF-5557 Determine mimetype during WB Dropbox upload

TomBaxter:
## Purpose
[OSF-5557](https://openscience.atlassian.net/browse/OSF-5557)
Use python-magic(libmagic) to determine mimetypes at time of upload

## Changes
update requirements.txt to include python-magic
update waterbutler/core/streams/metadata.py for new class MimeStreamWriter
update waterbutler/core/streams/__init__.py to import MimeStreamWriter
update waterbutler/providers/dropbox/provider.py "def upload" to utilize MimeStreamWriter
update tests/providers/dropbox/test_provider.py to use actual pdf file

ToDo:
Update other providers
Use "def upload" metadata return to update stored mimetype

## Side effects
May be both performance and system resource issues
Are uploads now slower?
Does the upload process now consume additional system memory

#OSF-5557


tests/providers/dropbox/test_provider.py
```diff
@@ -38,21 +38,25 @@ def settings():
 def provider(auth, credentials, settings):
     return DropboxProvider(auth, credentials, settings)
 
-
```
whit537: The standard in this file (and per PEP 8) is two lines between these top-level functions. Here I'm seeing that we trim to one, but inconsistently. Let's restore the original two-line spacing.

waterbutler/providers/dropbox/provider.py
```diff
@@ -195,7 +196,9 @@ def upload(self, stream, path, conflict='replace', **kwargs):
             throws=exceptions.UploadError,
         )
 
+        content_type = stream.writers['mime'].mimetype.decode("utf-8")
         data = yield from resp.json()
+        data['mime_type'] = content_type
```
whit537: Why are we only sniffing the mime type for Dropbox? In [OSF-5557](https://openscience.atlassian.net/browse/OSF-5557) it's reported that `contentType` is inconsistent across providers (though only observed to vary w/ update under Dropbox):

> A text file uploaded through the GUI or the API gets "content type" ...
>
> - box - "contentType": null,
- dataverse - "contentType": "text/plain; charset=US-ASCII",
- dropbox - "contentType": "application/octet-stream",
>
after update of .png file
>
- box - stays "contentType": null,
- dataverse - updates to "contentType": "application/octet-stream",
- dropbox - stays "contentType": "application/octet-stream",

Why not standardize the mime type for _all_ providers?

whit537:
##### Punchlist

- [ ] rebase on `develop`
- [ ] undo newline removal—https://github.com/CenterForOpenScience/waterbutler/pull/134#discussion_r67568847
- [ ] explain why we only care about Dropbox—https://github.com/CenterForOpenScience/waterbutler/pull/134#discussion_r67569345

waterbutler/core/streams/metadata.py
```diff
@@ -15,3 +19,22 @@ def write(self, data):
...
+
+
+class MimeStreamWriter():
+    """Stream-like object that determines mime-type and discards its input."""
+    def __init__(self):
+        self.mybytes = io.BytesIO()
+
+    @property
+    def mimetype(self):
+        return magic.from_buffer(self.mybytes.getvalue(), mime=True)
```
whit537: From [OSF-5557](https://openscience.atlassian.net/browse/OSF-5557):

> does it have a significant performance impact[?]

Rather than consuming the entire stream before sniffing the content type, is it possible to check when we've only consumed part of the stream? The man page for file(1) makes it sound like most filetypes are determinable from bytes towards the beginning of the file.

requirements.txt
```diff
@@ -11,6 +11,7 @@ stevedore==1.2.0
 tornado==4.2.1
 xmltodict==0.9.0
 pyjwe==0.1.6
+python-magic==0.4.10
 
 # Issue: certifi-2015.9.6.1 and 2015.9.6.2 fail verification (https://github.com/certifi/python-certifi/issues/26)
-certifi==2015.4.28
\ No newline at end of file
+certifi==2015.4.28
```
whit537: What changed on this line? I don't even see a whitespace or newline change.

tests/providers/dropbox/test_provider.py
```diff
@@ -38,21 +38,25 @@ def settings():
...
 @pytest.fixture
 def file_stream(file_like):
     return streams.FileStreamReader(file_like)
 
+@pytest.fixture
+def pdf_file_content():
+    return open('tests/resources/mypdf.pdf', "rb")
+
```
whit537: Two spaces here per the rest of the file.

PR #141, opened 2016-04-05
[SVCS-359] Feature/dryad/python 3.5

kwierman:
Replacing the last PR for the Dryad provider


waterbutler/providers/dryad/provider.py
```diff
@@ -0,0 +1,264 @@
...
+        """
+        raise exceptions.ReadOnlyProviderError(self)
+
+    def can_intra_copy(self, other, path=None):
+        """
+            All files in Dryad are able to be copied out (if accessible).
+
+            :returns: `True` Always
+        """
+        return True
```
felliott: I think this should be `False`.  `intra_copy` and `intra_move` are for copying/moving data efficiently within a single provider.  This would be appropriate if you were copying from one package in dryad to another.  Since the only use case is to copy from Dryad to another provider, you can rely on `waterbutler.core.provider.copy`, which does an intercopy.  It's implementation is essentially the same as your `_do_intra_move_or_copy`.  It downloads from provider A, then uploads to provider B.

felliott: Oh, and I tested this locally and the default `copy` seems to work without modification.

waterbutler/providers/dryad/metadata.py
```diff
@@ -0,0 +1,146 @@
...
+    @property
+    def etag(self):
+        return '{}::{}'.format(self.name, self.dryad_doi)
+
+    @property
+    def size(self):
+        return 0
+
+    def _json_api_links(self, resource):
+        return {}
```
felliott: We still want the `download` and `move` links (we don't support move, but we do support copy out).  Only `upload` and `delete` should be removed.  See my [Bitbucket PR](https://github.com/CenterForOpenScience/waterbutler/pull/198/files#diff-800939d0e06754bb0fbf1627f8749a7cR68) for an example.

` link is the same as the copy`

waterbutler/providers/dryad/metadata.py
```diff
@@ -0,0 +1,146 @@
...
+    @property
+    def file_parts(self):
+        return self._get_element_list_("dcterms:hasPart")
+
+    @property
+    def extra(self):
+        return super(BaseDryadMetadata, self).extra.update({
+            'references': self._get_element_("dcterms:references"),
+            'file_parts': self.file_parts
+        })
```
felliott: Add a class for `DryadFileRevisionMetadata`, even if it's just a dummy class whose version is always `'latest'`.

waterbutler/providers/dryad/metadata.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+class DryadFileMetadata(BaseDryadMetadata, metadata.BaseFileMetadata):
+    """
+        DryadFileMetadata needs to be instantiated with the same metadata from
+        the raw API call that normal packages do, AND with metadata from the file
+        metadata API and information that is only found in the bit stream.
+    """
+    def __init__(self, raw, doi, raw_file, file_name):
+        BaseDryadMetadata.__init__(self, raw, "doi:10.5061/dryad." + doi)
```
felliott: Please make this hardcoded string either a global variable or config setting.

tests/providers/dryad/test_provider.py
```diff
@@ -0,0 +1,365 @@
...
+    @pytest.mark.asyncio
+    @pytest.mark.aiohttpretty
+    async def test_metadata_root(self, provider, object_package_metadata):
+
+        aiohttpretty.register_uri('GET', DRYAD_META_URL +provider.doi.split('.')[-1],
+                                  body=object_package_metadata,
+                                  headers={'Content-Type': 'application/xml'})
+
+        path = WaterButlerPath('/')
+        root_metadata = await provider.metadata('/')
```
felliott: First arg to `metadata()` should be a wbpath, not a string.

kwierman:
All right, looks like we're good to go on this end. Heading back over to osf.io!

felliott:
Observers: this work has been accepted and resides on the [`feature/dryad` branch](https://github.com/CenterForOpenScience/waterbutler/tree/feature/dryad) with some changes.  It will be merged later.  I'm closing this PR.

kwierman:
Cool, thanks!

PR #148, opened 2016-05-13
[CR] logging refactor

felliott:
We construct and send logging payloads in at least five different places.  Making these all behave the same is a huge pain, and we've frequently gotten it wrong in the past.  This PR collects all logging
behavior into one method and updates the server endpoints to use it.

I've added a simple LogPayload object to standardize the representation of the path and metadata objects we serialize.

[#OSF-6316]

### DIFF
--- a/waterbutler/auth/osf/handler.py
+++ b/waterbutler/auth/osf/handler.py
@@ -82,11 +82,14 @@ class OsfAuthHandler(auth.BaseAuthHandler):
         if view_only:
             view_only = view_only[0].decode()
 
-        return (await self.make_request(
+        payload = (await self.make_request(
             self.build_payload(bundle, cookie=cookie, view_only=view_only),
             headers,
             dict(request.cookies)
         ))
+
+        payload['auth']['callback_url'] = payload['callback_url']
+        return payload
 
     async def get(self, resource, provider, request):
         """Used for v1"""
@@ -104,7 +107,7 @@ class OsfAuthHandler(auth.BaseAuthHandler):
             # View only must go outside of the jwt
             view_only = view_only[0].decode()
 
-        return (await self.make_request(
+        payload = (await self.make_request(
             self.build_payload({
                 'nid': resource,
                 'provider': provider,
@@ -113,3 +116,6 @@ class OsfAuthHandler(auth.BaseAuthHandler):
             headers,
             dict(request.cookies)
         ))
+
+        payload['auth']['callback_url'] = payload['callback_url']
+        return payload
--- a/waterbutler/core/log_payload.py
+++ b/waterbutler/core/log_payload.py
@@ -1,0 +1,62 @@
+from waterbutler.constants import IDENTIFIER_PATHS
+from waterbutler.core.path import WaterButlerPath
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class LogPayload:
+    """Simple wrapper class to abstract the object being logged.  Expects the parent resource,
+    provider, and either a metadata or WaterButlerPath object.
+    """
+
+    def __init__(self, resource, provider, metadata=None, path=None):
+        if path is None and metadata is None:
+            raise Exception("Log payload needs either a path or metadata.")
+
+        self.resource = resource
+        self.provider = provider
+        self.metadata = metadata
+        self.path = path or WaterButlerPath.from_metadata(metadata)
+
+    def serialize(self):
+        """Turn this LogPayload into something decent for mixed company.
+
+        A serialized LogPayload will always have the following keys::
+
+            {
+              'kind': 'file', // 'file' or 'folder'
+              'materialized': '/foo/bar/baz', // 'full human-readable path'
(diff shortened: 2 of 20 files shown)


waterbutler/auth/osf/handler.py
```diff
@@ -82,12 +82,15 @@ def build_payload(self, bundle, view_only=None, cookie=None):
...
             view_only = view_only[0].decode()
 
-        return (await self.make_request(
+        payload = (await self.make_request(
             self.build_payload(bundle, cookie=cookie, view_only=view_only),
             headers,
             dict(request.cookies)
         ))
 
+        payload['auth']['callback_url'] = payload['callback_url']
```
chrisseto: Is this for testing only or did the osf's return value change?

waterbutler/core/log_payload.py
```diff
@@ -0,0 +1,61 @@
...
+            }
+
+        If a metadata object is available it will also contain the `etag` and `extra` properties.
+
+        """
+        payload = {
+            'nid': self.resource,
+            'resource': self.resource,
+        }
+        if self.metadata is not None:
```
chrisseto: This should only happen in the case of a delete correct?

waterbutler/core/utils.py
```diff
@@ -102,6 +106,37 @@ def _async_retry(func):
...
+
+    resp = await send_signed_request('PUT', source.provider.auth['callback_url'], log_payload)
+    resp_data = await resp.read()
+
+    if resp.status // 100 != 2:
+        raise Exception(
+            'Callback for {} request failed with {!r}, got {}'.format(
+                action, resp, resp_data.decode('utf-8')
+            )
+        ) from sys.exc_info()[1]
```
chrisseto: ?

waterbutler/server/api/v0/copy.py
```diff
@@ -41,14 +39,12 @@ class CopyHandler(core.BaseCrossProviderHandler):
...
 
-        metadata = metadata.serialized()
-
         if created:
             self.set_status(201)
         else:
             self.set_status(200)
 
-        self.write(metadata)
+        self.write(metadata.serialized())
```
chrisseto: 👍

waterbutler/server/api/v1/provider/__init__.py
```diff
@@ -118,14 +117,19 @@ class ProviderHandler(core.BaseHandler, CreateMixin, MetadataMixin, MoveCopyMixi
...
+        if any((method in ('HEAD', 'OPTIONS'), status == 202, status // 100 != 2)):
+            return
+
+        if method == 'GET' and ((self.path.is_file and 'meta' in self.request.query_arguments) or
+                                (self.path.is_dir and 'zip' not in self.request.query_arguments)):
             return
 
         # Done here just because method is defined
         action = {
+            'GET': lambda: 'download_file' if self.path.is_file else 'download_zip',
```
chrisseto: Just for analytics?

waterbutler/server/api/v1/provider/__init__.py
```diff
@@ -135,55 +139,25 @@ def on_finish(self):
...
+
+            source = LogPayload(self.resource, self.provider, path=self.path)
+            destination = LogPayload(
+                self.dest_resource,
+                self.dest_provider,
+                metadata=self.dest_meta,
             )
+        elif action in ('create', 'create_folder', 'update'):
+            source = LogPayload(self.resource, self.provider, metadata=self.metadata)
+        elif action in ('delete'):
```
chrisseto: Could you add a trailing comma here just for clarity?
`('delete', )`

waterbutler/auth/osf/handler.py
felliott: The OSF returns `callback_url` at the same level as `auth`: https://github.com/CenterForOpenScience/osf.io/blob/develop/website/addons/base/views.py#L284

I wanted to make it part of the provider, so I just saved it into the auth dict.

waterbutler/core/log_payload.py
felliott: The opposite. I just realized that's a double negative, which is confusing.  But it also applies to the source of move and copy operations.  The source just has a validated path object, it's the destination that has a metadata object.

waterbutler/core/utils.py
felliott: Copied that from somewhere, `tasks/copy.py` perhaps?  I'll double-check to see if that's there for a reason, otherwise I'll make it simpler.

waterbutler/server/api/v1/provider/__init__.py
felliott: Will do!

waterbutler/server/api/v1/provider/__init__.py
felliott: Yep.  I originally had them entangled, then I split out analytics into a separate branch.  I should have removed this.  Good catch!

felliott:
merged.

PR #149, opened 2016-05-16
Allow non-cookie cross-origin requests

samchrisinger:
# Purpose

Waterbutler was not allowing cross-origin requests as a rule. We want to lessen this restriction some to allow cross-origin requests that are **not** sending cookies but are sending a bearer token.

# Changes

Add a check in `CORsMixin.set_default_headers` to check if a cross-origin request is being made under the circumstances described above. Add unit tests for the CORsMixin.

### DIFF
--- a/tests/server/test_utils.py
+++ b/tests/server/test_utils.py
@@ -28,16 +28,17 @@ class MockRequest(object):
             self,
             origin=None,
             method='GET',
-            cookies=None,
+            cookies=False,
             headers=None
     ):
         self.origin = origin or ''
         self.method = method
-        self.cookies = cookies or True
+        self.cookies = cookies
         self.headers = headers or {
             'Origin': origin
         }
 
+@mock.patch('waterbutler.server.settings.CORS_ALLOW_ORIGIN', '')
 class TestCORsMixin(ServerTestCase):
 
     def setUp(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


tests/server/test_utils.py
```diff
@@ -0,0 +1,99 @@
...
+            assert origin not in self.handler.headers['Access-Control-Allow-Origin']
+
+    @testing.gen_test
+    def test_set_default_headers_cross_origin_no_cookie_but_auth_header(self):
+        origin = 'http://foo.com'
+
+        for method in ('GET', 'POST', 'PUT', 'PATCH', 'DELETE'):
+            self.handler.request = MockRequest(
+                origin=origin,
+                method='GET',
```
felliott: Should be `method=method`?

tests/server/test_utils.py
```diff
@@ -86,14 +87,15 @@ def test_set_default_headers_cross_origin_no_cookie_no_auth_header(self):
     def test_set_default_headers_cross_origin_no_cookie_but_auth_header(self):
         origin = 'http://foo.com'
 
+        import pdb; pdb.set_trace()
```
felliott: Nope!

samchrisinger: 🐑

felliott:
LGTM.

PR #150, opened 2016-05-16
if no Origin header passed, don't set CORS headers

felliott:
Non-browser clients may or may not set an Origin header when making a
request, but browsers always should.  Don't set ANY CORS headers if
Origin isn't set.  Non-browser clients don't pay attention to CORS, and
a browser that doesn't set Origin should reject the response.


samchrisinger:
Tests and Mixin both greatly improved. Also does what it says it will do. LGTM.

PR #151, opened 2016-05-18
[CR] Waterbutler OneDrive Plugin

caseyrygt:
@felliott: This is ready for you to review.

Could you please review the copy functionality?  OneDrive's "Copy" function is asynchronous (the other methods are not) and requires the client to check the status of the copy job in order to know when it has completed.  If there's a better way than a for loop to check the OneDrive job status, please let me know.

Thanks!


waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,408 @@
...
+from waterbutler.providers.onedrive import settings
+from waterbutler.providers.onedrive.metadata import OneDriveRevision
+from waterbutler.providers.onedrive.metadata import OneDriveFileMetadata
+from waterbutler.providers.onedrive.metadata import OneDriveFolderMetadata
+
+logger = logging.getLogger(__name__)
+
+
+class OneDriveProvider(provider.BaseProvider):
+    NAME = 'onedrive'
```
felliott: Could you add a docstring to this giving a link to the API documentation and listing any quirks you've encountered?  A "quirk" would be any behavior that differs from what you would expect from a normal filesystem.  I've started adding these to the other providers, you can check them out for formatting examples: [S3](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/s3/provider.py#L31), [GitHub](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/github/provider.py#L41), [GDrive](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/googledrive/provider.py#L38)

I think I read in the OneDrive docs that text files are gzipped before sending.  That's a good example of a quirk to document, assuming I didn't completely fabricate that memory.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,408 @@
...
+            url = self._build_root_url("drive/root:", self._get_names(data), str(path))
+
+        resp = await self.make_request(
+            'GET',
+            url,
+            expects=(200, 404, ),
+            throws=exceptions.ProviderError
+        )
+
+        if (resp.status == 404):
```
felliott: The version of aiohttp we use requires responses to be released.  Calling `.json()` or another content-fetching method will automatically release the response after the read.  But since `.resp.json()` isn't called in this branch of the conditional, you'll need to manually `await resp.release()`.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,408 @@
...
+        logger.info('intra_move data:{}'.format(data))
+
+        if 'folder' not in data.keys():
+            return OneDriveFileMetadata(data, self.folder), True
+
+        folder = OneDriveFolderMetadata(data, self.folder)
+
+        return folder, True
+
+    async def download(self, path, revision=None, range=None, **kwargs):
```
felliott: I have an entry in my notes that downloading plain text files is broken because the content-lengths don't match, since they're gzipping the content.  Yay, I didn't make that up! :)

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,408 @@
...
+            expects=(200, 202),
+            throws=exceptions.IntraCopyError,
+        )
+        data = await resp.json()
+        status = data.get('status')
+        logger.info('_copy_status  status::{} resp:{} data::{}'.format(repr(status), repr(resp), repr(data)))
+        return True if resp.status == 200 else False
+
+    async def intra_move(self, dest_provider, src_path, dest_path):
+        #  https://dev.onedrive.com/items/move.htm
```
felliott: :+1: :+1: :+1: for the links to the doc pages!  Can you make this and the other comments a docstring instead, so it'll be visible on readthedocs?

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,408 @@
...
+
+        #  PATCH /drive/items/{item-id}
+        #  use cases: file rename or file move or folder rename or folder move
+
+        url = self.build_url(src_path.identifier)
+        payload = json.dumps({'name': dest_path.name,
```
felliott: According to the API docs, `name` isn't part of the request payload.

felliott: Oh, I didn't read carefully enough!  I missed the bit about being able to combine the move with other updates.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,408 @@
...
+
+        return base.child(path, _id=ids, folder=folder)
+
+    @property
+    def default_headers(self):
+        return {
+            'Authorization': 'Bearer {}'.format(self.token),
+        }
+
+    async def intra_copy(self, dest_provider, src_path, dest_path):
```
felliott: @caseyrygt: I don't see a better way of doing intra-copy than this, and it seems less than ideal to me.  I will defer to those with more understanding:

Pinging @chrisseto:  OneDrive's internal copy function requires polling an endpoint to find out when the copy job is completed.  Presumably this could be a celery task, but I don't think our architecture makes it easy to call a provider-specific task.  Am I correct on this?  Would it be better to just make OneDrive `can_intra_copy` return `False` and let the current async copy handle it?

felliott:
Hey @caseyrygt,

It's looking good!  I'm going to pause on this for now, but I should be able to resume next week.

Cheers,
@felliott

waterbutler/providers/onedrive/provider.py
chrisseto: @felliott Do you have any idea what the copy times are like?
if they are sub-45 seconds I would just say throw it in a loop w/ `asyncio.sleep` it won't be a performance hit at all.
The biggest consideration is how long we want to hold the user's attention.

felliott: @chrisseto, no idea myself.  @caseyrygt, could you try copying a really large file and let us know how long it takes?

caseyrygt: @chrisseto: at least when I was testing it, it was pretty quick, < 10-15
seconds.

I think there were one or two outliers but I believe they were all < 45
seconds.

So you're thinking of adding a asyncio.sleep(5) to the existing for loop?

Thanks.

On Thu, May 19, 2016 at 3:46 PM, Chris Seto <notifications@github.com>
wrote:

> In waterbutler/providers/onedrive/provider.py
> <https://github.com/CenterForOpenScience/waterbutler/pull/151#discussion_r63941686>
> :
>
> > +            ids = None
> > +            folder = False
> > +        else:
> > +            data = await resp.json()
> > +            ids = self._get_ids(data)[-1]
> > +            folder = ('folder' in data.keys())
> > +
> > +        return base.child(path, _id=ids, folder=folder)
> > +
> > +    @property
> > +    def default_headers(self):
> > +        return {
> > +            'Authorization': 'Bearer {}'.format(self.token),
> > +        }
> > +
> > +    async def intra_copy(self, dest_provider, src_path, dest_path):
>
> @felliott <https://github.com/felliott> Do you have any idea what the
> copy times are like?
> if they are sub-45 seconds I would just say throw it in a loop w/
> asyncio.sleep it won't be a performance hit at all.
> The biggest consideration is how long we want to hold the user's attention.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/CenterForOpenScience/waterbutler/pull/151/files/8a8a1d31e52eeac115cae84ef27bfd023122937a#r63941686>
>



-- 
Ryan Casey
CEO



<http://www.growingtechnologies.com/>
growingtechnologies.com
rcasey@growingtechnologies.com
Direct: (434) 326-0938
For Technical Support: 434-296-1500 option 2 /
techsupport@growingtechnologies.com

waterbutler/providers/onedrive/provider.py
caseyrygt: Argh!  Thanks for catching that...I tested JPGs, PDFs but not text!

waterbutler/providers/onedrive/provider.py
caseyrygt: @felliott - thanks...that explains all of the unclosed connection warnings I'm getting this morning. :)

For the download method, where I call: return streams.ResponseStreamReader(resp)

Will ResponseStreamReader automatically close the connection?  I would assume that falls under "content-fetching" but want to make sure.

Thanks.

waterbutler/providers/onedrive/provider.py
caseyrygt: Yes, that's done now.

waterbutler/providers/onedrive/provider.py
caseyrygt: I went ahead and added a asyncio.sleep(5) so that Waterbuttler will poll OneDrive every 5 seconds for up to 90 seconds.  

Copying a 50MB file took ~20 seconds today.  (I am getting an unrelated error which I posted to slack about...)

Will this work?

waterbutler/providers/onedrive/provider.py
felliott: Yep!

waterbutler/providers/onedrive/metadata.py
```diff
@@ -0,0 +1,96 @@
...
+#          logger.info('BaseOneDriveMetadata raw:{} path_obj:{}'.format(repr(raw), repr(path_obj)))
+
+    @property
+    def provider(self):
+        return 'onedrive'
+
+    @property
+    def materialized_path(self):
+        #  return '/{}/{}'.format(self.raw['parentReference']['path'].replace('/drive/root:/', ''), self.raw['name'])
+        return str(self._path_obj)
```
felliott: I'm getting errors when posting back the logs to the OSF because of this line.  When you construct the OneDriveMetadata objects, you pass `self.folder` as the `path_obj` arg in the constructor.  But `self.folder` is a string, not a WaterButlerPath object.  The string lacks the leading slash that is required for a valid `materialized_path`, so it fails validation.

You can either build a WaterButlerPath object from `self.folder` and pass that in, manually prepend the slash to the folder here, or forego passing `path_obj` entirely and build the materialized path from the `raw`.  Since this is the only place you use `self._path_obj`, I lean towards the third option, if the necessary information is available.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+        logger.info('validate_v1_path self::{} path::{}  url:{}'.format(repr(self), repr(path), self.build_url(path)))
+
+        resp = await self.make_request(
+            'GET', self.build_url(path),
+            expects=(200, 400),
+            throws=exceptions.MetadataError
+        )
+
+        if resp.status == 400:
+            return WaterButlerPath(path, prepend=self.folder)
```
felliott: Need to release resp before returning.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+        resp = await self.make_request(
+            'POST',
+            url,
+            data=payload,
+            headers={'content-type': 'application/json', 'Prefer': 'respond-async'},
+            expects=(202, ),
+            throws=exceptions.IntraCopyError,
+        )
+        if resp is None:
+            raise exceptions.IntraCopyError
```
felliott: Need to release resp before returning.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+    async def delete(self, path, **kwargs):
+        """ OneDrive API Reference: https://dev.onedrive.com/items/delete.htm """
+        resp = await self.make_request(
+            'DELETE',
+            self.build_url(path.identifier),
+            data={},
+            expects=(204, ),
+            throws=exceptions.DeleteError,
+        )
+        resp.release()
```
felliott: This needs to be `await`-ed to work (I do this all this time).

waterbutler/providers/onedrive/provider.py
felliott: Err, ignore me, you don't need to.  Sorry for the noise.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+        )
+        if resp is None:
+            raise exceptions.IntraCopyError
+        logger.info('resp::{}'.format(repr(resp)))
+        status_url = resp.headers['LOCATION']
+        await resp.release()
+        logger.info('status_url::{}'.format(repr(status_url)))
+        i = 0
+        status = None
+        while (i < 30):
```
felliott: Minor: please specify the iteration count and the sleep interval in the onedrive settings file.

felliott: Also, what happens if we don't see a copy-finished response after 30 iterations?  It looks like you pass `None` to `self_.contstruct_metadata` which results in an unhandled exception.

felliott: That said, this has been working well for me, and I've yet to run into an actual error with it.  Good job!

waterbutler/providers/onedrive/provider.py
felliott: <wince> actually, yes you do.  If you don't, you won't see the error immediately, but it will pop up eventually.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+        if path == '/':
+            return WaterButlerPath(path, prepend=self.folder)
+
+        logger.info('validate_v1_path self::{} path::{}  url:{}'.format(repr(self), repr(path), self.build_url(path)))
+
+        resp = await self.make_request(
+            'GET', self.build_url(path),
+            expects=(200, 400),
+            throws=exceptions.MetadataError
+        )
```
felliott: What does the 400 response code indicate?  Also, `validate_v1_path` should throw a 404 error if `path` doesn't exist.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+        return data if resp.status == 200 else None
+
+    async def intra_move(self, dest_provider, src_path, dest_path):
+        """ OneDrive API Reference: https://dev.onedrive.com/items/move.htm
+            Use Cases: file rename or file move or folder rename or folder move
+        """
+
+        url = self.build_url(src_path.identifier)
+        payload = json.dumps({'name': dest_path.name,
+                              'parentReference': {'id': dest_path.parent.full_path.strip('/') if dest_path.parent.identifier is None else dest_path.parent.identifier}})  # TODO: this feels like a hack.  parent.identifier is None
```
felliott: Minor: can you reformat this to respect line length limits?  You can pull it out into a temporary variable if needed.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+            resp = await self.make_request(
+                'PATCH',
+                url,
+                data=payload,
+                headers={'content-type': 'application/json'},
+                expects=(200, ),
+                throws=exceptions.IntraMoveError,
+            )
+        except exceptions.IntraMoveError as e:
+            if e.code != 403:
```
felliott: What should happen if OD sends back a 403 here?

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+        data = await resp.json()
+
+        logger.info('intra_move data:{}'.format(data))
+
+        if 'folder' not in data.keys():
+            return OneDriveFileMetadata(data, self.folder), True
+
+        folder = OneDriveFolderMetadata(data, self.folder)
+
+        return folder, True
```
felliott: The second return argument from both `intra_move` and `intra_copy` should reflect whether or not a new file was created, i.e. did we create a new file or overwrite an existing file?

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,419 @@
...
+        ]
+
+    async def create_folder(self, path, **kwargs):
+        """
+        OneDrive API Reference: https://dev.onedrive.com/items/create.htm
+        :param str path: The path to create a folder at
+        """
+        WaterButlerPath.validate_folder(path)
+
+        folderName = path.full_path.split('/')[-2]
```
felliott: There are a lot of place in the code where you do these sorts of operations on a WaterButlerPath object.  It would be nicer to create a subclass of WBP called OneDrivePath that's hides these behind descriptive method names.  `_get_one_drive_id` and `_get_sub_folder_path` would also be good candidates for that.  You can do it inline like the Github provider (https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/github/provider.py#L31) or split it out into its own `path.py` file.

waterbutler/providers/onedrive/metadata.py
```diff
@@ -0,0 +1,96 @@
...
+    def content_type(self):
+        if 'file' in self.raw.keys():
+            return self.raw['file'].get('mimeType')
+        return 'application/octet-stream'
+
+    @property
+    def extra(self):
+        return {
+            'id': self.raw.get('id'),
+            'etag': self.raw.get('etag'),
```
felliott: This is returning `null` for me.  Should it be `self.raw.get('eTag')` or just `self.raw['eTag']` if it's always present?

felliott:
Hey @caseyrygt,

Sorry for the long wait.  This is looking good, and your solution for copying is working perfectly for me!  I've left a few more issues to be addressed, and I think the downloading text files is still busted, but we're getting closer.  Thank you!

Cheers,
@felliott

caseyrygt:
hey @felliott, thanks for the feedback.  I'll get this and the OSF side of
things taken care of!

Best,
-Ryan

On Mon, Jul 11, 2016 at 3:13 PM, Fitz Elliott <notifications@github.com>
wrote:

> Hey @caseyrygt <https://github.com/caseyrygt>,
>
> Sorry for the long wait. This is looking good, and your solution for
> copying is working perfectly for me! I've left a few more issues to be
> addressed, and I think the downloading text files is still busted, but
> we're getting closer. Thank you!
>
> Cheers,
> @felliott <https://github.com/felliott>
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/CenterForOpenScience/waterbutler/pull/151#issuecomment-231834944>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABVki-JVpHuPB4KVdmdU92KEAT3Gl-Gfks5qUpXpgaJpZM4IhV5S>
> .
>



-- 
Ryan Casey
CEO



<http://www.growingtechnologies.com/>
growingtechnologies.com
rcasey@growingtechnologies.com
Direct: (434) 326-0938
For Technical Support: 434-296-1500 option 2 /
techsupport@growingtechnologies.com

waterbutler/providers/onedrive/metadata.py
caseyrygt: @felliott: how can I test this to see if I've fixed this?  I tried uploading a new PNG file in OSF, but under "Recent Activity" for the project I just get: Unable to retrieve log details.  I might have a bug on the OSF side, but is this the place where you'd expect to see this?  Thanks.

felliott: Yep, that is where should it be.  Are you up-to-date with the OSF?  There were some significant changes to how the logs are built in the 0.80 release, so the OSF-side logging code may need some updates.

The easiest way to make sure that the logging payloads are formatted correctly is to dump them in the OSF method that handles waterbutler logs.  See `website.addons.base.views.create_waterbutler_log`.

felliott: *EDIT*: I'm being silly.  This is the metadata object.  WB serializes this as the response to the action.  If you make a request directly to WB, it will return the `materialized_path` in the response object.

waterbutler/providers/onedrive/provider.py
caseyrygt: @felliott: I added count and sleep interval to the settings file, good point/idea.  As far as what happens if it fails, I would think it should throw an exception since that should be an exceptional / unusual event.  

What would be a "good" exception and text/language to use?  Something like: OneDrive file copy failed.  Please wait for 1-2 minutes and reload the page to see if OneDrive was able to copy the file.

waterbutler/providers/onedrive/provider.py
caseyrygt: @felliott: IIRC, OneDrive throws a 400 error if the file does not exist yet.  I think this is an issue on Move as I believe it was throwing an error when trying to move a file to a folder and the folder did **not** have a file with the same name.

As far as a 404 error, are you talking about the throws=exceptions.MetadataError ?

waterbutler/providers/onedrive/provider.py
caseyrygt: @felliott: you know, real programmers use ; and write everything in one line.  :)

waterbutler/providers/onedrive/provider.py
caseyrygt: @felliott: I suspect that was a work around for a bug somewhere else.  I've removed it.

waterbutler/providers/onedrive/provider.py
caseyrygt: I used dest_path.identifier is None like Box did to determine if it is new or an overwrite.

felliott: 👍

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,450 @@
...
+            await resp.release()
+            raise exceptions.IntraCopyError
+        logger.info('resp::{}'.format(repr(resp)))
+        status_url = resp.headers['LOCATION']
+        await resp.release()
+        logger.info('status_url::{}'.format(repr(status_url)))
+        i = 0
+        status = None
+
+        while (i < settings.ONEDRIVE_COPY_INTERATION_COUNT):
```
felliott: Typo: `INTERATION` => `ITERATION`.  Ditto in `settings.py`.

waterbutler/providers/onedrive/provider.py
felliott: It should throw `exceptions.NotFoundError(str(path))` if it can't find the file.

Re: "I think this is an issue on Move as I believe it was throwing an error when trying to move a file to a folder and the folder did not have a file with the same name."  Not sure I understand this.  When you move a file into a folder, it should validate that the folder exists.  It'll only check if a file with the same name exists if the `replace` argument is `warn` or `keep`.

waterbutler/providers/onedrive/provider.py
caseyrygt: I think I've taken care of this.  I created a OneDrivePath class modeled after the github one.  I also refactored the Upload method and I'll turn to create_folder as well.  Thanks.

waterbutler/providers/onedrive/provider.py
caseyrygt: @felliott: This is fixed.  However, I fixed it by setting accept-encoding to blank so the OneDrive web server will not encode/gzip the files.  

If I'm reading the documentation for aoihttp's client correctly, a gziped stream is supposed to be unzipped automatically but that doesn't seem to be working: http://aiohttp.readthedocs.io/en/stable/client.html#streaming-response-content

Thanks.

waterbutler/providers/onedrive/provider.py
caseyrygt: @felliott: Here's the use case, right or wrong :), when I upload a file, validate_path is called which calls validate_v1_path and when the file does not exist in the folder, OneDrive's API returns a 400 error.  Is that the correct work flow?

waterbutler/providers/onedrive/provider.py
caseyrygt: OK, I changed the message slightly and I think it's working as expected.  

**This is the error in WB logs:**
waterbutler.core.exceptions.CopyError: 202, OneDrive API file copy has not responded in a timely manner.  Please wait for 1-2 minutes, then query for the file to see if the copy has completed

**Here's what the OSF user sees:**
screen shot 2016-08-02 at 10 01 29 am


**Here's the result from the API, in Chrome's dev tools:**
screen shot 2016-08-02 at 10 02 44 am

waterbutler/providers/onedrive/provider.py
felliott: This turned out to be a good workaround.  WB might be updated in the future to not set the Content-Length when we receive gzipped content, but I need to investigate the ramifications of that.  In fact, this workaround is so good that [I stole it for the Box provider](https://github.com/CenterForOpenScience/waterbutler/blob/0.21.4/waterbutler/providers/box/provider.py#L228).  Thanks, @caseyrygt!

waterbutler/providers/onedrive/provider.py
felliott: Oops, never answered this!  Yes, this LGTM.

waterbutler/providers/onedrive/provider.py
felliott: :+1:

waterbutler/providers/onedrive/provider.py
felliott: @caseyrygt: Not quite.  `validate_v1_path` should be checking two things:

1. Does a file or folder exist at the given path?  If not, it should throw a `NotFoundError`.

2. If it does exist, do the object available at that path have the same type as inferred from the passed-in `path` argument?  If `path` ends with a slash, then we expect the object at that path to be a folder as well.  If it isn't, then we throw a NotFoundException.  Check out the Dropbox provider for a simple example.  The Box provider obscures this a bit because they have different endpoints for files and folders, meaning the type is automatically verified.

If both of these checks succeed, then `validate_v1_path` should return a `WaterButlerPath` (or `OneDrivePath` in this case).

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+            await resp.release()
+            return WaterButlerPath(path, prepend=self.folder)
+
+        data = await resp.json()
+
+        od_path = OneDrivePath(path)
+        names = od_path.file_path(data)
+        ids = od_path.ids(data)
+
+        wb_path = WaterButlerPath(names, _ids=ids, folder=path.endswith('/'))
```
felliott: Return a `OneDrivePath` here.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+        )
+
+        return streams.ResponseStreamReader(resp)
+
+    async def create_folder(self, path, **kwargs):
+        """
+        OneDrive API Reference: https://dev.onedrive.com/items/create.htm
+        :param str path: The path to create a folder at
+        """
+        WaterButlerPath.validate_folder(path)
```
felliott: Call `OneDrivePath.validate_folder()` instead here.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+        return OneDriveFolderMetadata(data, self.folder)
+
+    async def upload(self, stream, path, conflict='replace', **kwargs):
+        """ OneDrive API Reference: https://dev.onedrive.com/items/upload_put.htm
+            Limited to 100MB file upload. """
+        path, exists = await self.handle_name_conflict(path, conflict=conflict)
+
+        logger.info("upload path:{} path.parent:{} path.path:{} self:{}".format(repr(path), path.parent, path.full_path, repr(self)))
+
+        od_path = OneDrivePath(str(path))
```
felliott: This should already be a `OneDrivePath`.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+            for item in data['value']
+            if not item.get('deleted')
+        ]
+
+    async def _revisions_json(self, path, **kwargs):
+        """ OneDrive API Reference: https://dev.onedrive.com/items/view_delta.htm
+            As of May 20, 2016: for files, the latest state is returned.  There is not a list of changes for the file.
+        """
+        if path.identifier is None:
+                raise exceptions.NotFoundError(str(path))
```
felliott: Nit: this is over-indented

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+        return {
+            'Authorization': 'Bearer {}'.format(self.token),
+        }
+
+    async def intra_copy(self, dest_provider, src_path, dest_path):
+        """ OneDrive API Reference: https://dev.onedrive.com/items/copy.htm """
+
+        url = self.build_url(src_path.identifier, 'action.copy')
+        payload = json.dumps({'name': dest_path.name,
+                              'parentReference': {'id': dest_path.parent.full_path.strip('/') if dest_path.parent.identifier is None else dest_path.parent.identifier}})  # TODO: this feels like a hack.  parent.identifier is None in some cases.
```
felliott: Nit: line length

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+        while (i < settings.ONEDRIVE_COPY_ITERATION_COUNT):
+            logger.info('status::{}  i:{}'.format(repr(status), i))
+            status = await self._copy_status(status_url)
+            if (status is not None):
+                break
+            await asyncio.sleep(settings.ONEDRIVE_COPY_SLEEP_INTERVAL)
+            i += 1
+
+        if i >= settings.ONEDRIVE_COPY_ITERATION_COUNT:
+            raise exceptions.CopyError('OneDrive API file copy has not responded in a timely manner.  Please wait for 1-2 minutes, then query for the file to see if the copy has completed',
```
felliott: Nit: line length

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+from waterbutler.providers.onedrive.metadata import OneDriveFolderMetadata
+
+logger = logging.getLogger(__name__)
+
+
+class OneDrivePath(path.WaterButlerPath):
+    """OneDrive specific WaterButlerPath class to handle some of the idiosyncrasies of
+    file paths in OneDrive."""
+    def file_path(self, data):
+        parent_path = data['parentReference']['path'].replace('/drive/root:', '')
```
felliott: `/drive/root:` appears a bunch in this package.  Can you make it a constant?

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+from waterbutler.providers.onedrive.metadata import OneDriveFileMetadata
+from waterbutler.providers.onedrive.metadata import OneDriveFolderMetadata
+
+logger = logging.getLogger(__name__)
+
+
+class OneDrivePath(path.WaterButlerPath):
+    """OneDrive specific WaterButlerPath class to handle some of the idiosyncrasies of
+    file paths in OneDrive."""
+    def file_path(self, data):
```
felliott: Please add docstrings to these methods.

waterbutler/providers/onedrive/provider.py
```diff
@@ -0,0 +1,441 @@
...
+
+    def ids(self, data):
+        ids = [data['parentReference']['id'], data['id']]
+        url_segment_count = len(urlparse(self.file_path(data)).path.split('/'))
+        if (len(ids) < url_segment_count):
+            for x in repeat(None, url_segment_count - len(ids)):
+                ids.insert(0, x)
+        return ids
+
+    def one_drive_parent_folder(self, path):
```
felliott: Why does this method take a `path` arg?  `OneDrivePath` already represents a path.  Also, consider just overriding `WaterButlerPath.parent`, unless there's a specific reason you need both.

felliott:
I think the `OneDrivePath` code might need a rework.  Except for `validate_v1_path` and `validate_path`, the `path` argument to the methods you've implemented should always be an instance of `OneDrivePath`.  It's okay for you to provide OneDrive-specific implementations of `WaterButlerPath` methods.  `one_drive_parent_folder` should probably be `parent` and operate on `self`.  I'm not entirely sure but it looks `file_path` and `ids` are doing double-duty as constructors and accessors.  You might replace them with a `_build_from_onedrive_response(self, data)` class method that builds a new `OneDrivePath` object.  Then, the accessor functionality of those methods come from the `WaterButlerPath.parts`.  Also, `WaterButlerPath` has a `from_parts` class method that lets you build a new path object from an existing objects `parts`.

felliott:
Closing in favor of #205, which extends the work done in this PR.  Thanks, @caseyrygt!

PR #152, opened 2016-06-28
[WIP] Fedora provider

markpatton:
This is an initial attempt at a Fedora 4 provider. There are unit tests as well as an integration test. (Known issue: After the update to python 3.5 the integration test fails to have its tests collected.) 

The implementation is at least plausible, but the required behavior of various methods is not entirely clear from the documentation. In addition this pull request does not meet all the guidelines for contributing.

@mfraezz @felliott


waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,101 @@
...
+    def name(self):
+        return self.wb_path.name
+
+    @property
+    def provider(self):
+        return 'fedora'
+
+    @property
+    def modified(self):
+        return self._get_property('http://fedora.info/definitions/v4/repository#lastModified')
```
felliott: Minor: can you put this url in the `settings.py`?

waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,101 @@
...
+
+        raise exceptions.MetadataError('Could not find property {} in {}'.format(property_uri, self.fedora_id), code=404)
+
+
+class FedoraFolderMetadata(BaseFedoraMetadata, metadata.BaseFolderMetadata):
+    def list_children_metadata(self, repo_id):
+        return [self._create_child_metadata(child_id, repo_id) for child_id in self._list_children()]
+
+    def _list_children(self):
+        return [o['@id'] for o in self._get_resource(self.fedora_id).get('http://www.w3.org/ns/ldp#contains', [])]
```
felliott: Minor: please move this url to `settings.py`.

waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+            return FedoraFileMetadata(self.raw, child_id, WaterButlerPath(child_path_str), self.resource_index)
+
+
+# TODO Add hashes to extras.
+
+class FedoraFileMetadata(BaseFedoraMetadata, metadata.BaseFileMetadata):
+    @property
+    def name(self):
+        val = self._get_property('http://www.ebu.ch/metadata/ontologies/ebucore/ebucore#filename')
```
felliott: `settings.py`

waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,101 @@
...
+
+        if val:
```
felliott: Nit: `return val if val else self.wb_path.name`

waterbutler/providers/fedora/provider.py
```diff
@@ -0,0 +1,256 @@
...
+            'MOVE', src_url,
+            headers={'Destination': dest_url},
+            expects=(201,),
+            throws=exceptions.IntraMoveError
+        ) as move_resp:
+            # Delete tombstone of original file
+            async with self.request(
+                 'DELETE', src_url + '/fcr:tombstone',
+                 expects=(204, ),
+                 throws=exceptions.DeleteError,
```
felliott: These three lines are over-indented by one space, which is causing travis to fail.

waterbutler/providers/fedora/provider.py
```diff
@@ -0,0 +1,256 @@
...
+        url = self.build_repo_url(path)
+
+        # Must provide a Content-Type. Otherwise a container is created.
+        mime_type, encoding = mimetypes.guess_type(url)
+
+        if mime_type is None:
+            mime_type = 'application/octet-stream'
+
+        # Must not say content is RDF because a container will be created.
+        if mime_type in ['text/turtle', 'text/rdf+n3', 'application/n3', 'text/n3'
```
felliott: Minor: please move these to `settings.py`

markpatton:
I've updated the pull request as requested and I think its looking good. 

@felliott

TomBaxter:
Hello @markpatton , @felliott asked me to do some manual testing of this PR. I've spent some time on it but have not completed the task. I am however now turning my attention to another task for a bit and want to give you the results of my testing so far.

1. While testing I deleted a file through the fedora interface resulting in a tombstone. It appears that WB provider does not currently handle tombstones. This seems like something that users will run into, so I think we should try to handle the possibility of tombstones.

(The rest of these notes come from testing done against the WB API via the postman app)

2. Deletes succeed, but produce errors in osf and WB for logging call back

`Task <function log_to_callback at 0x7f0826e1c840> failed with Exception('Callback for delete request failed with <ClientResponse(http://localhost:5000/api/v1/project/qy85d/waterbutler/logs/) [500 INTERNAL SERVER ERROR]`

`File "/home/tkb/work/venv/cos/osf/osf.io/website/addons/base/views.py", line 450, in addon_delete_file_node
folder_children = FileNode.resolve_class(provider, FileNode.ANY).find(
File "/home/tkb/work/venv/cos/osf/osf.io/website/files/models/base.py", line 369, in resolve_class
raise exceptions.SubclassNotFound(provider)
SubclassNotFound: fedora`

3. File updates succeed but return return 204, should return 200. Also not returning file metadata.

4. File copy with replace argument True fails

HTTP code 412
`{
"response": "Destination resource already exists"
}`

5, File copy from osfstorage provider to fedora when file already exists fails

Different returns depending on what conflict arg is set to
no conflict specified default "warn" 
`{
"code": 500,
"message": "Internal Server Error"
}`
either "replace" or "keep" specified
`{
"code": 204,
"message": null
}`

6. Folder copies are returning incorrect path. Missing trailing "/"

7. When copying( with conflict=keep arg) an osfstorage folder with a file in it to fedora with a folder of the same name, the folder is copied but the original is not kept, and the file contained in it is left at the parent directory...

Example:

osfstorage:/folder1/subfolder1/file1   copied(keep) to fedora:/folder1/   when fedora:/folder1/subfolder1 already exists results in...

/folder1/file1
/folder1/subfolder1/ is overwritten
/folder1/subfolder1 (1)/  is not created

8. Implement confirm_delete parameter and _delete_folder_contents method in Provider class to handle root folder deletes.

See dropbox provider for example.


Thanks for your effort on this, I'll return to testing as soon as I can.
@TomBaxter

markpatton:
I missed the November 1 comments. Those looks easy to resolve.

As for more recent comments from Tom:
1. Tombstones are an interesting case. Remember that Fedora is not a singular user oriented service like Dropbox, its a component that would be deployed as part of a broader system. That broader system might want to treat tombstones all sorts of different ways. I believe that right now that through the OSF the user cannot see tombstones and that tombstones are deleted when a file is deleted. Does Water Butler/OSF has some notion of a tombstone that could be used? Do you want the user to be able to see tombstones and not delete tombstones when a file is deleted though the OSF? What happens when a user tries to overwrite a tombstone?
2. Is this all deletes or just some? I'll have to poke around and see what's in the Fedora logs.
3. Easy enough to change, but I based the behavior on the dropbox client which also returns file metadata on upload.
4. Have to look into the move behavior. I wasn't aware of this flag and dropbox doesn't appear to examine it, perhaps because move on dropbox behaves differently.
5. Same comments on move apply to copy.
6. Should be easy enough to fix. (Unit tests should catch this.)
7. I wasn't aware of keep and various (1) behavior. Again dropbox doesn't seem to have any special handling so I suspect its not hard to fix.
8. That requires some thought. If the "root" folder is the Fedora root, then it can't be deleted. But this should not be hard to support.

There is a need for a common integration testing framework to ensure the correct behavior of storage providers and help with implementation. (Some of the above should even have been caught by unit tests.) I did my own minimal integration testing, but it was not nearly sufficient as these problems show. In addition, the documentation for storage providers should specify all the required behavior including status codes, side effects, etc.

I need to see if I can get some time to go through these issues. Unless there are some hidden mismatches with Fedora behavior I suspect it won't be too bad. Most of the time will be spent updating the integration tests.

@TomBaxter @felliott

felliott:
Hey @markpatton!

This is a quick followup to your response from [Dec. 2](https://github.com/CenterForOpenScience/waterbutler/pull/152#issuecomment-264530633).  I'm starting to CR this addon now, so there may be more to follow.

1. The issue of tombstones is a tricky one.  WB doesn't have a concept of them.  I'm going to have a think about this.
2. This issues is caused by not having a Fedora-specific file subclass on the OSF-side.  See this directory: https://github.com/CenterForOpenScience/osf.io/tree/develop-backup/website/files/models
3. Yes, it should be returning a 200 OK + the updated file metadata.
4. The parameter is named `conflict` the value is `replace`.  We recently got the docs building on rtd again, there's an overview of the api here: https://waterbutler.readthedocs.io/en/latest/api.html  If you drop down to the section titled *Move & Copy*, it'll explain the allowed body parameters and what they do.  If anything in there is unclear, please let me know.  I wrote that awhile ago, and I'm not sure how lucid it is.
5. See 4.
6. :+1:
7. [core.provider.BaseProvider.handle_naming](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/core/provider.py#L343) is the method that handles appending ` (n)` when `conflict=keep` is specified.
8. Calling `DELETE` on a folder root should never delete the root folder.  Instead, it should return a 400 and an error message.  However, if the user passes in the `confirm_delete` query parameter, then WB should delete all children of the root folder, but not the folder itself.  The intended purpose of this is to allow easy clearing of a storage folder without rendering the addon in an unusable state (with nothing to connect to).

Re: integration tests:  yes, absolutely!  We just haven't had the resources to devote to it so far.  @TomBaxter did start on a series of Postman collections that can be used to verify copy behavior after creating and authorizing a couple of test nodes on your local dev OSF.  See https://waterbutler.readthedocs.io/en/latest/testing.html for more.  The collections live in https://github.com/CenterForOpenScience/waterbutler/tree/develop/tests/postman.  I'm excited to start reviewing your integration tests, and hope we can adapt that for some of the other providers.

Cheers,
@felliott

markpatton:
@felliott 

I believe I have addressed all the issues that have been raised. More problems may turn up when I move on to the next step of getting the Fedora addon talking to this provider.

waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,127 @@
...
+        if self.resource_index is None:
+            self.resource_index = {o['@id'].rstrip('/'): o for o in raw}
+
+    @property
+    def path(self):
+        return '/' + self.wb_path.full_path
+
+    @property
+    def name(self):
+        return self.wb_path.name
```
felliott: `name` and `materialized` should be decoded.  If I upload a file called `mü-ทนน.zip`, the `path` is (correctly) `/mu%CC%88-%E0%B8%97%E0%B8%99%E0%B8%99.zip`, but `name` should be `mü-ทนน.zip`, while `materialized` should be `/mü-ทนน.zip`.

waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,127 @@
...
+    @property
+    def modified(self):
+        return self._get_property(settings.LAST_MODIFIED_PROPERTY_URI, None)
+
+    @property
+    def modified_utc(self):
+        return self._get_property(settings.LAST_MODIFIED_PROPERTY_URI, None)
+
+    @property
+    def created_utc(self):
```
felliott: `modified_utc` and `created_utc` should both be in the `YYYT-MM-DDThh:mm:ss+00:00` format (e.g. `2017-03-13T16:22:39+00:00`).  It currently returns the `Z` format and includes milliseconds.

waterbutler/providers/fedora/provider.py
```diff
@@ -0,0 +1,330 @@
...
+
+class FedoraProvider(provider.BaseProvider):
+    """Provider for a Fedora 4 repository.
+
+    This provider uses the Fedora 4 REST API which is based on LDP.
+
+    API docs: https://wiki.duraspace.org/display/FEDORA4x/Fedora+4.x+Documentation
+
+    Quirks:
+    * Fedora 4 is a component, not a singular service like Dropbox or Box.
```
felliott: What special characters does Fedora allow or disallow.  We currently have a project going on to try to document this for the other providers, so it would be nice to have it here too.  e.g. some providers don't allow you to have a literal `%` in a filename, others do.  How does it handle unicode filenames?

waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,127 @@
...
+
+class FedoraFileMetadata(BaseFedoraMetadata, metadata.BaseFileMetadata):
+    @property
+    def name(self):
+        # Do not return filename property because the OSF user cannot change it
+        return self.wb_path.name
+
+    @property
+    def size(self):
+        return self._get_property(settings.SIZE_PROPERTY_URI, None)
```
felliott: Cast this to a `int`.

waterbutler/providers/fedora/metadata.py
```diff
@@ -0,0 +1,127 @@
...
+
+from waterbutler.providers.fedora import settings
+
+# The raw data is a list of JSON_LD resources parsed as JSON.
+# The resource_index is an index of the raw data keyed by '@id' normalized by having any trailing slashes stripped.
+# The fedora_id is the id of the JSON_LD resource which the metadata resource represents.
+# The wb_path is WaterButlerPath of resource.
+
+
+class BaseFedoraMetadata(metadata.BaseMetadata):
```
felliott: Your `extra` field is currently empty.  If you want to return any Fedora-specific metadata you can put it in there.

waterbutler/providers/fedora/provider.py
```diff
@@ -0,0 +1,330 @@
...
+    async def validate_v1_path(self, path, **kwargs):
+        """Return WaterButlerPath for a path string.
+
+        Ensure that if the path is to a folder, it corresponds to a Fedora container
+        and otherwise corresponds to a Fedora binary. Fedora resource must also exist.
+
+        Throw NotFoundError if resource does not exist or types do not match
+        """
+
+        wb_path = WaterButlerPath(path)
```
felliott: You seem to have some double-encoding issues.  I have a file called `meow%mix.txt`.  Its path should be `/meow%25mix.txt`.  But if I try to download-as-zip my fedora repo, it builds a url with the path `/meow%2525mix.txt`.  You might need to subclass `WaterButlerPath` to get it to keep track of whether the path has already been encoded.  The GoogleDrive provider does this.

waterbutler/providers/fedora/metadata.py
markpatton: Good to know. I'm not sure what would be useful at the moment. Triples associated with the Fedora object could be stuffed in extras, but I don't want to do that without a plan.

waterbutler/providers/fedora/metadata.py
markpatton: Fix added.

waterbutler/providers/fedora/metadata.py
markpatton: Done.

waterbutler/providers/fedora/metadata.py
markpatton: I believe I have this fixed. I added an integration test and manual testing seems to work.

waterbutler/providers/fedora/provider.py
markpatton: Fedora 4 is based on LDP which means that any valid URL can be a resource.

waterbutler/providers/fedora/provider.py
markpatton: This works for me now.

felliott:
This work has been archived on the [`feature/fedora` branch](https://github.com/CenterForOpenScience/waterbutler/tree/feature/fedora).  Closing this PR for now.

PR #156, opened 2016-07-11
Fix logic for handling google docs extentions

TomBaxter:
## Purpose:
Fix logic for handling google docs extentions.
This fixes the duplicate folder name of the parent of a google doc file when using the google docs extension in the API call.

## Changes:
Update waterbutler/providers/googledrive/provider.py add "continue" line to logic handling google doc extention.

## Side effects
None

[#OSF-6659]


felliott:
Merged into master and released.  Thanks, @TomBaxter!

PR #158, opened 2016-07-11
[OSF-6405]Registration fails if 2 files in Google Drive have the same name.

TomBaxter:
## Purpose:

[OSF-6405](https://openscience.atlassian.net/browse/OSF-6405)
Enable the correct handling of google docs files of the same name but different type
## Changes:

Update  waterbutler/providers/googledrive/metadata.py add extension to path and materialized for google docs files.

Update waterbutler/providers/googledrive/provider.py to query by specific mime type for google docs files.

Update waterbutler/providers/googledrive/utils.py add "def get_mimetype_from_ext(ext)"
## Side effects:

None

[#OSF-6405]


felliott:
Merged into `develop`.

PR #159, opened 2016-07-15
Attach level=info to 404 errors on their way to sentry

TomBaxter:
## Purpose:
[OSF-6704](https://openscience.atlassian.net/browse/OSF-6704)
Set 404 errors to level info in sentry

## Changes:
Update  waterbutler/server/api/v1/core.py add check for exc.code == 404 and add data={'level': 'info'}

## Side effects:
None

[#OSF-6704]


felliott:
Merged with some changes.  Thanks, @TomBaxter!

PR #162, opened 2016-07-27
[OSF-6584]Convert Dropbox provider to Dropbox v2 API

TomBaxter:
## Purpose:

[OSF-6584](https://openscience.atlassian.net/browse/OSF-6584)
Convert Dropbox provider to Dropbox v2 API
## Changes:

Update tests/providers/dropbox/test_provider.py update uris and responses for v2 API
Added waterbutler/providers/dropbox/exceptions.py added exceptions for v2 409s
Updated  waterbutler/providers/dropbox/metadata.py updated for v2 API returns
Updated waterbutler/providers/dropbox/provider.py extensive changes throughout for v2 API
## Side effects:

Dropbox no longer returns MimeType
OSF still needs to convert to v2 API before deadline
list_revisions limit has been reduced from 1,000 to 100. We previously set limit to 250.
Dropbox now imposes a hard limit of 10,000 files/folders during any one operation. returns "too_many_files Void The operation would involve more than 10,000 files and folders."
## QA note:

During QA be sure to test large quantities of files (>2000) in a single folder. folder metadata call now involves pagination in 2000 result increments.

[#OSF-6584]

### DIFF
--- a/waterbutler/providers/dropbox/metadata.py
+++ b/waterbutler/providers/dropbox/metadata.py
@@ -19,49 +19,50 @@ class BaseDropboxMetadata(metadata.BaseMetadata):
             path = path[len(self._folder):]
         return super().build_path(path)
 
-    @property
-    def extra(self):
-        return {
-            'revisionId': self.raw['rev']
-        }
-
 
 class DropboxFolderMetadata(BaseDropboxMetadata, metadata.BaseFolderMetadata):
 
     @property
     def name(self):
-        return os.path.split(self.raw['path'])[1]
+        return os.path.split(self.raw['path_display'])[1]
 
     @property
     def path(self):
-        return self.build_path(self.raw['path'])
+        return self.build_path(self.raw['path_display'])
 
 
 class DropboxFileMetadata(BaseDropboxMetadata, metadata.BaseFileMetadata):
 
     @property
     def name(self):
-        return os.path.split(self.raw['path'])[1]
+        return os.path.split(self.raw['path_display'])[1]
 
     @property
     def path(self):
-        return self.build_path(self.raw['path'])
+        return self.build_path(self.raw['path_display'])
 
     @property
     def size(self):
-        return self.raw['bytes']
+        return self.raw['size']
 
     @property
     def modified(self):
-        return self.raw['modified']
+        return self.raw['server_modified']
 
     @property
     def content_type(self):
-        return self.raw['mime_type']
+        # return self.raw['mime_type']
+        return 'dropbox/whatarewegonnadowithyou'
 
     @property
     def etag(self):
         return self.raw['rev']
+
+    @property
+    def extra(self):
+        return {
+            'revisionId': self.raw['rev']
+        }
 
 
 # TODO dates!
@@ -77,4 +78,10 @@ class DropboxRevision(metadata.BaseFileRevisionMetadata):
 
     @property
     def modified(self):
-        return self.raw['modified']
+        return self.raw['server_modified']
+
+    @property
+    def extra(self):
+        return {
+            'revisionId': self.raw['rev']
+        }
--- a/waterbutler/providers/dropbox/metadata.py
+++ b/waterbutler/providers/dropbox/metadata.py
@@ -51,8 +51,7 @@ class DropboxFileMetadata(BaseDropboxMetadata, metadata.BaseFileMetadata):
 
     @property
     def content_type(self):
-        # return self.raw['mime_type']
-        return 'dropbox/whatarewegonnadowithyou'
+        return None
 
     @property
     def etag(self):
--- a/waterbutler/providers/dropbox/metadata.py
+++ b/waterbutler/providers/dropbox/metadata.py
@@ -60,7 +60,8 @@ class DropboxFileMetadata(BaseDropboxMetadata, metadata.BaseFileMetadata):
     @property
     def extra(self):
         return {
-            'revisionId': self.raw['rev']
+            'revisionId': self.raw['rev'],
+            'id': self.raw['id']
         }
 
 
@@ -82,5 +83,6 @@ class DropboxRevision(metadata.BaseFileRevisionMetadata):
     @property
     def extra(self):
         return {
-            'revisionId': self.raw['rev']
+            'revisionId': self.raw['rev'],
+            'id': self.raw['id']
         }
--- a/waterbutler/providers/dropbox/provider.py
+++ b/waterbutler/providers/dropbox/provider.py
@@ -34,43 +34,48 @@ class DropboxProvider(provider.BaseProvider):
         self.token = self.credentials['token']
         self.folder = self.settings['folder']
 
-    async def dropbox_request(self,
-                              url,
-                              body,
-                              throws,
-                              expects=(200, 409,)):
+    async def dropbox_request(self, url, body, throws,
+                              expects=(200, 409,), *args, **kwargs):
+        if 'path' in body:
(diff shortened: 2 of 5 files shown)


waterbutler/providers/dropbox/provider.py
```diff
@@ -32,23 +34,46 @@ def __init__(self, auth, credentials, settings):
...
+                if resp.status == 409:
+                        self.dropbox_conflict_error_handler(data)
+        except:
+            raise
+        return data
+
+    def dropbox_conflict_error_handler(self, data):
+        if 'error' in data.keys():
+            if data['error'][data['error']['.tag']]['.tag'] == 'not_found':
+                raise exceptions.NotFoundError(data['error_summary'])
```
felliott: I just fell victim to this, too!  The argument to `NotFoundError` is the path that wasn't found, not a replacement error message.  When I hit this I get: `"Could not retrieve file or directory path/not_found/."`

waterbutler/providers/dropbox/provider.py
```diff
@@ -59,147 +84,116 @@ def can_duplicate_names(self):
...
-                },
-                expects=(200, ),
-                throws=exceptions.IntraMoveError,
-            )
-        except exceptions.IntraMoveError as e:
-            if e.code != 403:
-                raise
-
+            data = await self.dropbox_request(url, body, throws)
+        except exceptions.DropboxNamingConflictError:
```
felliott: `AttributeError: module 'waterbutler.core.exceptions' has no attribute 'DropboxNamingConflictError'`

waterbutler/providers/dropbox/provider.py
```diff
@@ -59,147 +84,116 @@ def can_duplicate_names(self):
...
+                url = self.build_url('files', 'copy_reference', 'save')
+                body = json.dumps({"copy_reference": from_ref,
+                                   'path': dest_path.full_path.rstrip('/')})
+                expects = (200, 201, 409)
+                data = await dest_provider.dropbox_request(url,
+                                                           body,
+                                                           throws,
+                                                           expects=expects)
+                data = data['metadata']
+        except exceptions.DropboxNamingConflictError:
```
felliott: `AttributeError: module 'waterbutler.core.exceptions' has no attribute 'DropboxNamingConflictError'`

waterbutler/providers/dropbox/metadata.py
```diff
@@ -19,50 +19,50 @@ def build_path(self, path):
...
+        return None
 
     @property
     def etag(self):
         return self.raw['rev']
 
+    @property
+    def extra(self):
+        return {
+            'revisionId': self.raw['rev']
```
felliott: Minor: please include the `id` in `extra`.  We might not be using it yet, but it might come in handy later.

waterbutler/providers/dropbox/provider.py
TomBaxter: Complete.

waterbutler/providers/dropbox/provider.py
```diff
@@ -32,23 +34,46 @@ def __init__(self, auth, credentials, settings):
         self.token = self.credentials['token']
         self.folder = self.settings['folder']
 
+    async def dropbox_request(self,
```
TomBaxter: I've added *args and **kwargs to dropbox_request.

I'd rather not subclass make_request and instead am using request, an async context manager for make_request. I don't believe any functionality is lost and that way I can use "async with" which appears to be the new preferred syntax for 3.5.2+

waterbutler/providers/dropbox/provider.py
```diff
@@ -32,23 +34,46 @@ def __init__(self, auth, credentials, settings):
...
+                              url,
+                              body,
+                              throws,
+                              expects=(200, 409,)):
+        try:
```
TomBaxter: Without it, I don't get the chance to catch the 409s. A down side for async with.

waterbutler/providers/dropbox/provider.py
```diff
@@ -32,23 +34,46 @@ def __init__(self, auth, credentials, settings):
...
+                data = await resp.json()
+                if resp.status == 409:
+                        self.dropbox_conflict_error_handler(data)
+        except:
+            raise
+        return data
+
+    def dropbox_conflict_error_handler(self, data):
+        if 'error' in data.keys():
+            if data['error'][data['error']['.tag']]['.tag'] == 'not_found':
```
TomBaxter: Complete.

waterbutler/providers/dropbox/provider.py
TomBaxter: Complete

waterbutler/providers/dropbox/provider.py
```diff
@@ -32,23 +34,46 @@ def __init__(self, auth, credentials, settings):
...
-            expects=(200,),
-            throws=exceptions.MetadataError
-        )
-
-        data = await resp.json()
-        explicit_folder = data['is_dir']
+        url = self.build_url('files', 'get_metadata')
+        body = json.dumps({'path': self.folder.rstrip('/') + path.rstrip('/')})
+        throws = exceptions.MetadataError
+        data = await self.dropbox_request(url, body, throws)
```
TomBaxter: Complete

waterbutler/providers/dropbox/provider.py
```diff
@@ -59,147 +84,116 @@ def can_duplicate_names(self):
...
-        for item in data['contents']:
-            if item['is_dir']:
-                folder.children.append(DropboxFolderMetadata(item, dest_folder))
-            else:
-                folder.children.append(DropboxFileMetadata(item, dest_folder))
-
+        folder.children = [item for item in await dest_provider.metadata(dest_path)]
         return folder, True
 
     async def intra_move(self, dest_provider, src_path, dest_path):
```
TomBaxter: I'm not sure I'm following, I don't think intra_move is currently using `move()`

felliott: Derp, it's not.  I left out a bunch of context on this.  What I meant is that `intra_copy` works both within and across accounts.  `intra_move` only works within an account.  That's why the implementation of `can_intra_copy` and `can_intra_move` differ.  But couldn't intra_move be update to call intra_copy, then follow that up with a delete of the source path?  If so, that would mean that intra_move can be used both within and across accounts.

waterbutler/providers/dropbox/provider.py
TomBaxter: Complete

waterbutler/providers/dropbox/provider.py
TomBaxter: Took me a while to dig up why I did it this way. I think it is correct.

As near as I can tell, currently WaterButler/OSF only has access to one set of credentials per Provider for any given WaterButler/OS authentication. Therefore the only intra move possible is from one project to another that both are using the same dropbox credentials and therefore the same dropbox account.

  Because of this, the current intra_copy and intra_move both work for the only valid intra project/inter Provider scenario possible. It is actually the intra copy that has more going on then necessary. I figured I would use copy_reference, some what for future proofness, but mostly, just cause.
  I can remove the copy_ref section if you like. intra_copy will work fine without it for the only scenario that can work with out current credentials setup.
  Also, in answer to you original question, there is no reason I can think of that we can't have intra_move just do an intra copy and then a delete.

TomBaxter:
All comments responded to. Ready for further review.

felliott:
Merged into `develop`.

PR #164, opened 2016-08-01
[CR] ownCloud provider

kwierman:
## Purpose

This feature consists of the ownCloud provider. ownCloud utilizes the WebDAV interface for file access and modification.

## Changes

These changes should be fairly independent of other WB changes.

## Side effects

In intra-move/copy, if the move and copy commands are handled in the main event loop, a bland RunTime error is thrown. However, if the operations are handled in celery, then the operations continue as normal. There is nothing in either command to indicate improper async coro wrapper creation. Thus, an `is` operator is used in `can_intra_move/copy` to deny celery access to these methods.

This seems to be a python3.5+ bug.

## Ticket

n/A


waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,253 @@
...
+
+
+class OwnCloudProvider(provider.BaseProvider):
+    """Provider for the ownCloud cloud storage service.
+
+    This provider uses the OCS1.7 standard for communication
+
+    API docs: https://www.freedesktop.org/wiki/Specifications/open-collaboration-services-1.7/
+
+    Quirks:
```
felliott: :heart_eyes_cat:

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,253 @@
...
+            return self.url + '/remote.php/webdav/'
+        return self.url + 'remote.php/webdav/'
+
+    async def validate_v1_path(self, path, **kwargs):
+        """
+            Prepends folder setting on paths.
+        """
+        if path == '/':
+            return WaterButlerPath(path, prepend=self.folder)
+        return WaterButlerPath(path, prepend=self.folder)
```
felliott: Is it possible to check whether the path exists or not?  If so, `validate_v1_path` should do that.  If the file/folder can't be found, it should throw a NotFoundError.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,253 @@
...
+        if len(items) == 1:
+            return items[0], True
+        meta = self.metadata(dest_path)
+        meta.children = items
+        return resp
+
+    async def intra_copy(self, dest_provider, src_path, dest_path):
+        response = await self.do_dav_move_copy(src_path, dest_path, 'COPY')
+
+        return response, True
```
felliott: Second return argument from `intra_copy` and `intra_move` indicates whether a new file was created (`True`), or an old file was overwritten (`False`).

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,253 @@
...
+        if metadata.size is not None:
+            return streams.ResponseStreamReader(download_resp, size=metadata.size)
+
+        stream = streams.StringStream(await download_resp.read())
+        if download_resp.headers.get('Content-Type'):
+            stream.content_type = download_resp.headers['Content-Type']
+        stream.name = metadata.name
+        return stream
+
+    async def upload(self, stream, path, conflict='replace', **kwargs):
```
felliott: Seeing something odd here.  When I upload files via the provider, they're being saved in owncloud as the form-encoded data.  e.g. when I upload a file that only contains the string `fesnabbaurnm`, it's being saved in local storage as:
````
--169f57f74e814c33a26670e06ffa77a8
Content-Disposition: form-data; name="attributes"

{"name": "aaa-01.txt", "parent": {"id": null}}
--169f57f74e814c33a26670e06ffa77a8
Content-Disposition: form-data; name="file"; filename="aaa-01.txt"
Content-Type: application/octet-stream
Content-Transfer-Encoding: binary

fesnabbaurnm

--169f57f74e814c33a26670e06ffa77a8--
````

If I upload the same file directly via the Owncloud interface, it gets saved as just the text file.  Is it possible that the Content-Type header isn't being set correctly, so owncloud doesn't realize it needs to decode it?

File: waterbutler/providers/owncloud/provider.py [hunk 5]
```diff
@@ -0,0 +1,253 @@
...
+            data=data_stream,
+            headers=data_stream.headers,
+            expects=(201, 200),
+            throws=exceptions.UploadError,
+            auth=self._auth,
+            connector=self.connector
+        )
+        await response.release()
+        meta = await self.metadata(path)
+        return meta, True
```
felliott: Second arg should indicate whether the file is new or if another file was overwritten.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,253 @@
...
+        response = await self.make_request('PROPFIND',
+            self._webdav_url_ + parse.quote(path.full_path),
+            expects=(200, 204, 207),
+            throws=exceptions.MetadataError,
+            auth=self._auth,
+            connector=self.connector
+        )
+
+        items = []
+        if response.status == 207:
```
felliott: If I'm reading this right, only a 207 means that files were found in the folder.  A 200 or 204 means that the directory is empty.  Is that correct?

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def can_intra_move(self, dest_provider, path=None):
+        return self is dest_provider
+
+    async def do_dav_move_copy(self, src_path, dest_path, operation):
+        """Performs a quick copy or move operation on the remote host.
+
+        :param str src_path: user-supplied path to the source object
+        :param str dest_path: user-supplied path to the destination object
+        :param str dest_path: user-supplied path to the destination object
```
felliott: Nit: you've listed `dest_path` twice.

felliott:
General comment: most of your calls to `make_request` include the `200` status code in the `expects` parameter.  If that is really a legitimate return value (it might be, I don't know WebDAV/owncloud), then no problem.  But if the endpoint is not supposed to return a 200, then leave it out of the `expects` so we'll be notified if we encounter it.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,253 @@
...
+    def __init__(self, auth, credentials, settings):
+        super().__init__(auth, credentials, settings)
+
+        self.folder = settings['folder']
+        self.url = credentials['host']
+        self._auth = aiohttp.BasicAuth(
+            credentials['username'],
+            credentials['password']
+        )
+        self.connector = aiohttp.TCPConnector(verify_ssl=False)
```
felliott: I saw your note in *Quirks*.  Do you mean that `verify_ssl` isn't compatible with some actual SSL implementation that are guarding live owncloud servers, or do you mean that we shouldn't verify when trying to test it locally without SSL?  If it's the latter, then make this a configuration parameter that defaults to True in the settings, and must be overridden to be False when developing.

waterbutler/providers/owncloud/metadata.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @property
+    def etag(self):
+        return str(self.attributes['{DAV:}getetag'])
+
+    @property
+    def modified(self):
+        return self.attributes['{DAV:}getlastmodified']
+
+    def __init__(self, href, folder, attributes=None):
```
felliott: Nit: can you make `__init__` the first method declared in the class?

felliott:
Looking good, so far, @kwierman!  I'm pausing for now.  I'll test move/copy, DAZ, and funky file names after the upload encoding problem is fixed.

waterbutler/providers/owncloud/provider.py
kwierman: Yup. Fixed. Will be in next iteration of PR. I was wondering about that since it looks like some of the providers such as cloudfiles don't implement remote validation. On the other hand, dropbox for example does have remote validation.

In an case, it's done.

waterbutler/providers/owncloud/provider.py
kwierman: Since this value is not returned in the server response from the quick move/copy, I'm going to implement this as validating the destination path first.

waterbutler/providers/owncloud/provider.py
kwierman: Alright, I've been dumb. I assumed the FormDataStream set `Content-Type` explicitly like the FileStreamReader. One line fix.

waterbutler/providers/owncloud/provider.py
kwierman: Roger. I'm utilizing the same trick as before.

waterbutler/providers/owncloud/provider.py
kwierman: OCS1.7 is wierd. 207 means successful return on multipart metadata, which occurs for a number of reasons, mostly folder listing and single file parsing. 204 for non-multipart data, which usually occurs for empty folders. 200 occurs for all PUT/PATCH/POST requests. I was having trouble predicting which was which in the beginning as the ownCloud implementation of OCS1.7 doesn't use the full standard. The python2.7 implementation kinda washed over all of this.

tl;dr: I went through and cleaned this up. My apologies for the white wash.

waterbutler/providers/owncloud/provider.py
kwierman: Woops. Fixed.

waterbutler/providers/owncloud/provider.py
kwierman: So, locally, I have SSL turned on in my docker instance. However, if I try to connect to the server with either my python2.5 implementation or the python2.7 pyocclient, I get an SSL verification error. Turns out [I'm not the only one](https://github.com/owncloud/pyocclient/issues/102).

The current solution is to implement an SSL verify setting in OSF. I'm going to pass this as a setting to WB, and remove this "quirk". Basically, I'm passing the buck forward to OSF.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,277 @@
...
+            self._webdav_url_ + parse.quote(full_path.full_path),
+            expects=(204, 207, 404),
+            throws=exceptions.MetadataError,
+            auth=self._auth,
+            connector=self.connector
+        )
+        await response.release()
+        if response.status == 404:
+            raise exceptions.NotFoundError(str(full_path.full_path))
+        return full_path
```
felliott: `validate_v1_path` should also make sure that the implied path type matches the found path type. If I give it the path `/foo/`, then I'm asking for a directory named `foo`.  If we find a file but not a directory with that name, we should still throw a 404.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,277 @@
...
+
+    async def upload(self, stream, path, conflict='replace', **kwargs):
+        """
+            Utilizes default name conflict handling behavior then adds the
+            appropriate headers and creates the upload request.
+
+            :param str path: user-supplied path to upload.
+            :raises: :class:`waterbutler.core.exceptions.UploadError`
+        """
+        print("HEY KEVIN UPLOAD IS BEING CALLED")
```
felliott: `print` is inappropriate here; it should text you. :smile:

kwierman: lol. Sorry, this got pushed mid fix as well. On it.

felliott:
Just noticed all the `parse.quote`s sprinkled throughout.  If the path needs to always be url-encoded when building the url, you can specify a default encoder/decoder in the `WaterButlerPath` object.  The googledrive provider does this by [creating a GoogleDrivePath class](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/googledrive/provider.py#L28) that subclasses `WaterButlerPath`.  Now `path_obj.path` is automatically encoded.  If you need the unencoded path, you can use the `raw_path` property.

waterbutler/providers/owncloud/provider.py
felliott: No problem!  Could you throw in a comment or docstring explaining this?

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,277 @@
...
+        """Deletes path on remote host
+
+        :param str path: user-supplied path to delete.
+        :raises: :class:`waterbutler.core.exceptions.DeleteError`
+        """
+
+        delete_resp = await self.make_request(
+            'DELETE',
+            self._webdav_url_ + parse.quote(path.full_path),
+            expects=(200, 201),
```
felliott: When does `DELETE` return 200 and when does it return 201?  I'm just surprised, because usually this is a 204.

kwierman: Should be 201 only. If that's not bad enough, the [documentation](https://www.freedesktop.org/wiki/Specifications/open-collaboration-services-1.7/) states that it should be 100.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,277 @@
...
+            data=data_stream,
+            headers=data_stream.headers,
+            expects=(200, 201),
+            throws=exceptions.UploadError,
+            auth=self._auth,
+            connector=self.connector
+        )
+        await response.release()
+        meta = await self.metadata(path)
+        return meta, not exists
```
felliott: I know OCS has weird ideas about return codes, but instead of determining `exists` at the beginning of the function, could you check the status code?  In most APIs the 200 indicates an update and 201 indicates creation.

waterbutler/providers/owncloud/provider.py
kwierman: Alright, turns out this was a two-part problem. Fixed now.

waterbutler/providers/owncloud/provider.py
kwierman: This got washed away in the recent series of commits. We're making SSL optional. Mainly, I'd like to disable during dev operations. This is being set in OSF and forwarded to WB through serialized settings.

waterbutler/providers/owncloud/provider.py
kwierman: In the latest set of updates, you'll see this implemented as `status == 200` for most of the CRUD like operations. Best as I can tell from playing with the API, this works.

waterbutler/providers/owncloud/provider.py
kwierman: I think I got this figured out. The python2.7 implementation from OC used a generalized response function which had me REALLY confused.

kwierman:
All right, I think I got everything covered both here and in OSF. I'm going to clean up the commits and make this something more... professional.


This latest round raised a question regarding path verification. If I try and call the v1 method from the v0 method, the 404 exception goes uncaught for some operations. For instance, during uploads, verification gets called before the upload on the destination path. Naturally, this throws a 404 because the path doesn't exist yet. I guess what I'd like to know is what are the differences between the context that v0 and v1 path verification get called in.

mfraezz:
There needs to be a `v2` route for folder listing, otherwise you'll only be able to see the top level directories. If everything is wired up correctly, this might be as simple as adding `'owncloud'` to `api.settings.ADDONS_FOLDER_CONFIGURABLE`

screen shot 2016-09-13 at 11 11 48

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,301 @@
...
+            self._webdav_url_ + full_path.full_path,
+            expects=(200, 207, 404),
+            throws=exceptions.MetadataError,
+            auth=self._auth,
+            connector=self.connector
+        )
+        content = await response.content.read()
+        await response.release()
+
+        item = await utils.parse_dav_response(content, self.folder)
```
felliott: I just tried to upload a file through the OSF file browser and got an error.  The problem is the v0 API calls `validate_path` with the path for the new file, which does not yet exist.  OC returns a 404, which you expect, but the response content is an error. `parse_dav_response` crashes when it tries to parse it.  Here's the error content I got:

````
<?xml version="1.0" encoding="utf-8"?>\n<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">\n  <s:exception>Sabre\\DAV\\Exception\\NotFound</s:exception>\n  <s:message>File with name 1VN7B37-1.jpeg could not be located</s:message>\n</d:error>
````

Uploading via v1 works perfectly!

Comment (kwierman) [inline waterbutler/providers/owncloud/provider.py]: This _should_ have been fixed yesterday via [this edit](https://github.com/CenterForOpenScience/waterbutler/pull/164/files#diff-12716d06776b109681372e60fb56db2dR46). Throwing the 404 still causes actions like uploads and folder creations to fail though

Comment (felliott) [inline waterbutler/providers/owncloud/provider.py]: That edit is sort of the problem.  `validate_path` shouldn't throw a 404 Not Found.  It should instead create a WaterButlerPath object without the identifier set.  You could wrap the call to `parse_dav_response` in a `try: except` block like in `validate_v1_path`, but `pass` instead of re-raising the error.

Comment (kwierman) [inline waterbutler/providers/owncloud/provider.py]: Ah! Ok. So 404 for the v1 path, but not for the v0. Gotcha.

File: waterbutler/providers/owncloud/provider.py [hunk 15]
```diff
@@ -0,0 +1,303 @@
...
+            connector=self.connector
+        )
+        content = await response.content.read()
+        await response.release()
+
+        try:
+            item = await utils.parse_dav_response(content, '/')
+        except exceptions.NotFoundError:
+            pass
+        if full_path.is_dir and type(item[0]) != OwnCloudFolderMetadata:
```
felliott: This still dies because `item` isn't set.  This and the next few lines should probably be in an `else:` clause.

kwierman: Woops, meant to use a r`eturn full_path` instead of `pass`. The following checks won't pass if the 404 is thrown anyway. Either way, I think this explains my earlier problems.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,303 @@
...
+            self._webdav_url_ + self.folder + path.path,
+            expects=(201,),
+            throws=exceptions.CreateFolderError,
+            auth=self._auth,
+            connector=self.connector
+        )
+        await resp.release()
+        # get the folder metadata
+        meta = await self.metadata(path.parent)
+        return [m for m in meta if m.path == OwnCloudFolderMetadata('/' + path.path, {}).path][0]
```
felliott: This doesn't work for me, because `OwnCloudFolderMetadata('/' + path.path, {}).path` is `/`.  It might be simpler to compare `m.path` against either `path.materialized_path` or `path.full_path` (both of which were added to WaterButlerPath fairly recently).  Both will add the leading slash.  `full_path` includes the storage root, `materialized_path` does not.

Also, if you need to add some OwnCloud specific extensions to WaterButlerPath, you can create a custom subclass like [the GitHub provider does](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/github/path.py).

kwierman:
Wow, really cool! I have one last osf related issue that might effect wb. I'll fix it and include these changes in a minute.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,304 @@
...
+            expects=(201,),
+            throws=exceptions.CreateFolderError,
+            auth=self._auth,
+            connector=self.connector
+        )
+        await resp.release()
+        # get the folder metadata
+        meta = await self.metadata(path.parent)
+        expected = OwnCloudFolderMetadata(path.path, {}).path
+        return [m for m in meta if m.full_path == expected][0]
```
felliott: D'oh! Should have caught this before.  This crashes because `.full_path` is a property of WaterButlerPaths, not BaseMetadatas.

kwierman: Would the preferred solution then be to construct a WB and then use `.full_path`?

felliott: You probably don't need to construct a new WBPath, you already have one (`path`).  `OwnCloudFolderMetadata`'s `.path` property seems to have the same construction as `path.materialized_path`, so I would just compare those:

````
return [m for m in meta if m.path == path.materialized_path][0]
````

File: waterbutler/providers/owncloud/provider.py [hunk 18]
```diff
@@ -0,0 +1,304 @@
...
+
+            :param str path: user-supplied path to create. must be a directory.
+            :param boolean precheck_folder: flag to check for folder before attempting create
+            :rtype: :class:`waterbutler.core.metadata.BaseFolderMetadata`
+            :raises: :class:`waterbutler.core.exceptions.FolderCreationError`
+        """
+        resp = await self.make_request(
+            'MKCOL',
+            self._webdav_url_ + self.folder + path.path,
+            expects=(201,),
```
felliott: This is annoying.  OC throws a 405 instead of a 409 when you try to create a folder with the same name as an existing folder.  Can you expect the 405 and re-throw it as a 409?

kwierman: Sure thing. I feel your pain with OCS1.7. It's quite the bear.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,304 @@
...
+
+        try:
+            item = await utils.parse_dav_response(content, '/')
+        except exceptions.NotFoundError:
+            # Re-raise with the proper path
+            raise exceptions.NotFoundError(str(full_path.full_path))
+        if full_path.is_dir and type(item[0]) != OwnCloudFolderMetadata:
+            raise exceptions.NotFoundError(str(full_path.full_path))
+        elif not full_path.is_dir and type(item[0]) == OwnCloudFolderMetadata:
+            raise exceptions.NotFoundError(str(full_path.full_path))
```
kwierman: Done

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,304 @@
...
+        await response.release()
+
+        try:
+            item = await utils.parse_dav_response(content, '/')
+        except exceptions.NotFoundError:
+            return full_path
+        if full_path.is_dir and type(item[0]) != OwnCloudFolderMetadata:
+            raise exceptions.NotFoundError(str(full_path.full_path))
+        elif not full_path.is_dir and type(item[0]) == OwnCloudFolderMetadata:
+            raise exceptions.NotFoundError(str(full_path.full_path))
```
kwierman: Done.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,304 @@
...
+    def can_duplicate_names(self):
+        return True
+
+    def can_intra_copy(self, dest_provider, path=None):
+        return self is dest_provider
+
+    def can_intra_move(self, dest_provider, path=None):
+        return self is dest_provider
+
+    async def do_dav_move_copy(self, src_path, dest_path, operation):
```
kwierman: Done.

waterbutler/providers/owncloud/provider.py
```diff
@@ -0,0 +1,304 @@
...
+        )
+        content = await resp.content.read()
+        items = utils.parse_dav_response(content, self.folder)
+        if len(items) == 1:
+            return items[0], True
+        meta = self.metadata(dest_path)
+        meta.children = items
+        return resp, resp.status_code == 200
+
+    async def intra_copy(self, dest_provider, src_path, dest_path):
```
kwierman: Done.

waterbutler/providers/owncloud/provider.py
kwierman: Cool, this seems to work.

waterbutler/providers/owncloud/provider.py
kwierman: Done and done. This helped catch another error elsewhere in the code. I'm used to checking `response.status_code` in python2.7 requests, but here it's `response.status`. I'm going to give one last shake down .

felliott:
Looks good to me, @kwierman!  I'm going to rebase and clean up the PR a bit, then I'll get this merged in.

kwierman:
Cool! If you's prefer, I can do the rebase and some of the cleanup.

felliott:
Nah, don't worry about it.  It's quick.

felliott:
This PR has been rebased, pushed to a feature branch, and resubmitted as #168.

PR #168, opened 2016-09-15
Add ownCloud as a supported provider to WB

felliott:
## Purpose / Changes

This PR will add @kwierman's ownCloud provider (#164) to WaterButler.  It does *not* depend on [the OSF-side PR](https://github.com/CenterForOpenScience/osf.io/pull/6316), but will be useless until that is merged.  Both PRs are being maintained as feature branches for testing purposes, and will be merged to develop later.

## Testing

This is a full-feature read+write provider and will have to pass all the same tests that our other providers do.  The runscope tests will need to be updated, and manual testing of create/read/update/delete/move/copy operations done to make sure the v0 API is exercised

### DIFF
--- a/docs/provider.owncloud.rst
+++ b/docs/provider.owncloud.rst
@@ -1,0 +1,13 @@
+Metadata
+--------
+
+.. automodule:: waterbutler.providers.owncloud.metadata
+    :members:
+    :undoc-members:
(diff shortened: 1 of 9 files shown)


felliott:
Done and merged!

PR #169, opened 2016-09-15
Pass user id on file download for analytics [OSF-3628]

darioncassel:
## Purpose

Send auth information (user id) on download so that analytics can know whether the downloader is a contributor to the file's project or not. As part of [https://openscience.atlassian.net/browse/OSF-3628](https://openscience.atlassian.net/browse/OSF-3628).

## Changes

- Pass the ```{'user': auth['id']}``` if it exists or ```{}``` as param in ```async with self.signed_request``` for ```async def download(self, path, version=None, revision=None, mode=None, **kwargs)```.

- Added tests for new behavior: replaced ```test_download``` with ```test_download_with_auth``` and ```test_download_without_auth```.

## Side Effects

None visible.

## Ticket

[https://openscience.atlassian.net/browse/OSF-3628](https://openscience.atlassian.net/browse/OSF-3628)

### DIFF
--- a/tests/providers/osfstorage/test_provider.py
+++ b/tests/providers/osfstorage/test_provider.py
@@ -18,6 +18,7 @@ from waterbutler.core.path import WaterButlerPath
 from waterbutler.providers.osfstorage import OSFStorageProvider
 from waterbutler.providers.osfstorage.settings import FILE_PATH_COMPLETE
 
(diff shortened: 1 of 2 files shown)


felliott:
Thanks, @darioncassel! LGTM. Merged with minor fixes.

PR #170, opened 2016-10-02
[Feature] Environment Variable Settings Support

icereval:
Support for settings overrides w/ Environment Variables for simple values.

[Twelve-Factor App Config](https://12factor.net/config)

Discovered need when simplifying docker-compose integration w/ OSF.


felliott:
Merged into develop.

PR #172, opened 2016-10-13
Create empty folders in zip download [OSF-6125]

darioncassel:
## Purpose
When DAZing a folder, subfolders with no file descendants are omitted. This PR allows for empty folders to be included in the zip archive.

## Changes
- Recognize an empty folder and pass it to the ZipLocalFile method as a empty file
- In ZipLocalFile, check to see if the file is a directory, if it is then set the directory bitflag

## Side Effects
- Required import of types.MethodType for creating temporary empty file object

## Ticket
[https://openscience.atlassian.net/browse/OSF-6125](https://openscience.atlassian.net/browse/OSF-6125)

### DIFF
--- a/waterbutler/core/streams/zip.py
+++ b/waterbutler/core/streams/zip.py
@@ -122,14 +122,18 @@ class ZipLocalFile(MultiStream):
     """
     def __init__(self, file_tuple):
         filename, stream = file_tuple
-        filename = filename.strip('/')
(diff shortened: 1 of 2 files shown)


felliott:
Thanks, @darioncassel!  This looks good.  I added a couple of commits to move the folder response stream into its own class and bypass compression for folders.  Merged.

PR #176, opened 2016-11-14
Add undeletable_contents metadata field

alexschiller:
## Purpose
In OSFStorage has two kinds of undeletable files, preprints and checked out files. Currently to find out if a folder is deletable, every nested file and folder must be explored. By adding this field, the fangorn logic can be simplified and also handle situations where the folder has not been fully loaded.

## Changes
This adds `undeletable_contents` to the osfstorage extra metadata field. The field returns a list of ids of folder contents that are undeletable

## Side effects
None known


alexschiller:
Closing. Current course of action changed, this may come up again post django-osf.

PR #178, opened 2017-01-04
Remove setuptools ... a lot

cwisecarver:
The osf docker-compose script was giving some people `__version__` errors still. I believe this fixes them. I built a docker image locally and was able to start it and run the server. The uninstall and delete is probably overkill but why not delete it as many times as possible. h/t @felliott Your commit comment gave me enough to figure it out.


felliott:
Merged as a hotfix, v0.29.1!  Thanks, @cwisecarver!

PR #187, opened 2017-02-02
Update config and README for local docker development

TomBaxter:
## Purpose:
[SVCS-245](https://openscience.atlassian.net/browse/SVCS-245)
Update local development Docker config and doc for WB

## Changes:
Update Documentation and `API_URL` to reflect use of 192.168.168.167 instead of localhost

## Side effects
None

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,8 @@ invoke celery
 
 ### Configuring
 
+### For Docker setup use 192.168.168.167 instead of localhost
+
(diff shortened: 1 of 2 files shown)


felliott:
Hey @TomBaxter, 

This is not the correct place for this change.  WB may be used outside the OSF, and so should default to using localhost instead of our arbitray IP address.  Instead this setting should be overridden where the dockerized OSF/WB/MFR config is defined, in osf.io. It looks like it [already has been](https://github.com/CenterForOpenScience/osf.io/blob/develop/.docker-compose.wb.env#L12), so the question is why the setting isn't making it to WB.  Perhaps the env var is misnnamed?

Cheers,
@felliott

PR #191, opened 2017-02-09
[SVCS-246] Gracefully reject unsupported methods

Johnetordoff:
# Purpose

Fix error message when unsupported method is used so it sends helpful information.

# Changes

Adds new exception with appropriate message. New message `{"code": 500, "message": "Method 'patch' not supported, currently supported methods are PUT, POST, GET, HEAD, DELETE"}
`

# Side Effect

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-246


waterbutler/core/exceptions.py
```diff
@@ -52,6 +52,13 @@ class AuthError(PluginError):
     """
 
 
+class UnsupportedMethodError(PluginError):
```
felliott: Minor: this error should return status code 405 (Method Not Allowed).  In general 5xx is "we messed up", 4xx is "you messed up".

waterbutler/core/exceptions.py
```diff
@@ -52,6 +52,13 @@ class AuthError(PluginError):
...
+    """WaterButler related errors raised
+    from a :class:`waterbutler.core.auth` should
+    inherit from AuthError
```
felliott: Please update this docstring.

felliott:
Err, please add some unit tests, too.  We don't always need tests for provider code, but this is constrained enough to be worthwhile.

tests/providers/osfstorage/test_handler.py
```diff
@@ -0,0 +1,23 @@
+from waterbutler.auth.osf.handler import OsfAuthHandler
+from waterbutler.server.settings import DOMAIN
+import unittest
+import requests
+
+
+class TestOsfAuthHandler(unittest.TestCase):
```
Johnetordoff: @felliott I've added these unittests but can't set them up properly so they don't fail in Travis. They seem to work fine loccally.

tests/providers/osfstorage/test_handler.py
```diff
@@ -0,0 +1,23 @@
+from waterbutler.auth.osf.handler import OsfAuthHandler
```
felliott: The OSF auth handler is a completely different beast from the osfstorage provider.  This test should live in a new file, 'tests/auth/osf/test_handler.py`.

tests/providers/osfstorage/test_handler.py
```diff
@@ -0,0 +1,23 @@
...
+    def setUp(self):
+        self.handler = OsfAuthHandler
+
+    def test_supported_methods(self):
+
+        supported_methods = ['put', 'post', 'get', 'head', 'delete']
+        assert all(method in self.handler.ACTION_MAP.keys() for method in supported_methods)
+
+        for method in supported_methods:
+            assert requests.request(method, DOMAIN + '/v1/resources/test/providers/test/').status_code == 404
```
felliott: I suspect this is failing because you're issuing a live request across the network (even if it's just localhost).  You don't need to do that.  The auth handler is just a regular python class that makes HTTP requests.  You can create an instance of it, then call `.get(resource, provider, request)` on it manually and assert that it ties properly.

felliott: Oh, and you can mock `make_request` to return a dummy repsonse, so that supported methods won't fail.

tests/auth/osf/test_handler.py
```diff
@@ -0,0 +1,38 @@
...
+    async def test_supported_methods(self):
+
+        supported_methods = ['put', 'post', 'get', 'head', 'delete']
+        assert all(method in self.handler.ACTION_MAP.keys() for method in supported_methods)
+
+        for method in supported_methods:
+            self.request.method = method
+
+            with pytest.raises(AuthError):
+                await self.handler.get("test", "test", self.request)
```
felliott: This is trying to issue a live http request.  It's getting all the way to `auth.osf.handler.make_request` which tries to connect to `localhost:5000`.  You should mock `make_request` to return a fake successful response, and test against that.  See https://github.com/CenterForOpenScience/waterbutler/blob/develop/tests/server/api/v1/test_fuzzing.py#L106  for an example (though that mocks the entire `.get()` method, not just `make_request()`).

tests/auth/osf/test_handler.py
```diff
@@ -0,0 +1,38 @@
+from waterbutler.auth.osf.handler import OsfAuthHandler
+from waterbutler.auth.osf import settings
+from waterbutler.core.exceptions import AuthError, UnsupportedMethodError
+from tests.utils import HandlerTestCase
+import pytest
+import tornado
```
felliott: Nitpick: please fix the import order per CONTRIBUTING.md

waterbutler/core/exceptions.py
```diff
@@ -52,6 +52,13 @@ class AuthError(PluginError):
     """
 
 
+class UnsupportedMethodError(PluginError):
+    """An unsupported method was used
```
felliott: Nitpick: "An unsupported HTTP method was used"

waterbutler/auth/osf/handler.py
```diff
@@ -107,11 +107,20 @@ def build_payload(self, bundle, view_only=None, cookie=None):
             # View only must go outside of the jwt
             view_only = view_only[0].decode()
 
+        try:
```
felliott: Forgot one!  Please move this test up to the top of the method.  If we're going to raise an exception, no point in doing all the other work first.

tests/auth/osf/test_handler.py
Johnetordoff: CONTRIBUTING.md isn't 100% percent clear on this, it basically say use PEP8, but pay attention to line length then provides an example which isn't alphabetical or entirely based on line length. I'm confused.

felliott: PEP8 style is to group imports into three groups:  standard libs, external libs, internal imports.  Within each group, sort by line length.

felliott:
Merged with fd5f36bc7960e36494bbf7c4195644b8d4b555cb.  Thanks, @Johnetordoff!

PR #192, opened 2017-02-10
Update intra_move and intra_copy to return folder's children.

TomBaxter:
## Purpose:

[SVCS-229](https://openscience.atlassian.net/browse/SVCS-229)
WB API: OSFStorage intra_move/_copy does not return children

## Changes:

update  waterbutler/providers/osfstorage/provider.py
- def intra_move update to return children
- def intra_copy update to return children

## Side effects:

None

[#SVCS-229]


felliott:
Ace!  Merged as b09b786476b8ec6c23436c769e43e246c6482f2c.  Thanks, @TomBaxter!

PR #193, opened 2017-02-13
[SVCS-236] Fix broken revision handling for Github

Johnetordoff:
# Purpose 

Stop the 404s that occur when attempting to view github revisions.

# Changes

Astonishingly simple change stops WB from searching whole tree when it only needs single commit. The problem is caused by unnecessarily grabbing the tree for revisions where the metadata for a single commit is all that's needed.

# Side Effects

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-236

# Related PR

#190

### DIFF
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -79,9 +79,12 @@ class GitHubProvider(provider.BaseProvider):
 
         branch_ref, ref_from = None, None
         if kwargs.get('ref'):
-            branch_ref = kwargs.get('ref')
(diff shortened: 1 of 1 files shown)


Johnetordoff:
I think this function is running in a different context than you think, two things stick out at me:

1. Using the raw 'ref' which should be the commit sha works fine with no 404, it should need the tree sha, so something is wrong with out assumptions.
2. In your pseudo-code the root path '/' can't be found in the tree and crashes gives a 404.

PR #195, opened 2017-02-21
[SVCS-21] Add exception to prevent accessing gdoc files without file extension

Johnetordoff:
# Purpose
Google Doc files stored via the google drive WB provider are currently api-accessible both with and without extension. This forces you to use an extension.

# Changes
Adds new exception that's called when extension-less gdoc is called.

# Side Effects
None that I know of.

# Tickets
https://openscience.atlassian.net/browse/SVCS-21

### DIFF
--- a/waterbutler/providers/googledrive/provider.py
+++ b/waterbutler/providers/googledrive/provider.py
@@ -412,6 +412,10 @@ class GoogleDriveProvider(provider.BaseProvider):
                         "and trashed = false " \
                         "and mimeType != 'application/vnd.google-apps.form' " \
                         "and mimeType != 'application/vnd.google-apps.map' " \
+                        "and mimeType != 'application/vnd.google-apps.document' " \
+                        "and mimeType != 'application/vnd.google-apps.drawing' " \
+                        "and mimeType != 'application/vnd.google-apps.presentation' " \
+                        "and mimeType != 'application/vnd.google-apps.spreadsheet' " \
                         "and mimeType {} '{}'".format(
                             clean_query(part_name),
                             '=' if part_is_folder else '!=',
@@ -444,11 +448,6 @@ class GoogleDriveProvider(provider.BaseProvider):
                 throws=exceptions.MetadataError,
             ) as resp:
                 ret.append(await resp.json())
-
-        content_type = ret[-1]['mimeType']
-        if not ext and 'application/vnd.google-apps' in content_type and content_type != 'application/vnd.google-apps.folder':
-            raise exceptions.NotFoundError(path=path)
-
         return ret
 
     async def _resolve_id_to_parts(self, _id, accum=None):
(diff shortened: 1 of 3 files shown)


waterbutler/providers/googledrive/provider.py
```diff
@@ -445,6 +445,10 @@ def _build_upload_metadata(self, folder_id, name):
             ) as resp:
                 ret.append(await resp.json())
 
+        content_type = ret[-1]['mimeType']
+        if not ext and 'application/vnd.google-apps' in content_type and content_type != 'application/vnd.google-apps.folder':
+            raise exceptions.ExtensionNotIncluded(provider=self.NAME)
```
felliott: This doesn't need a bespoke error, a standard 404 Not Found is fine.

waterbutler/providers/googledrive/provider.py
```diff
@@ -445,6 +445,10 @@ def _build_upload_metadata(self, folder_id, name):
...
+
```
felliott: This isn't quite what we want.  First, there's no need to throw a custom exception, a regular 404 Not Found is fine. Second, this will throw the exception even if there's a non-gdoc file with the name and no extension.  Example:

image

I've got a Google Doc named `unshared`, and a Google Doc, Google Sheet, and plain text file all named `shared`.  Here's what I'd expect if I tried to get metadata for each of these:

````
Request                                         Result
GET .../googledrive/unshared.gdoc?meta=         200 OK + metadata
GET .../googledrive/unshared?meta=              404 Not Found
GET .../googledrive/shared.gdoc?meta=           200 OK + gdoc metadata
GET .../googledrive/shared.gsheet?meta=         200 OK + gsheet metadata
GET .../googledrive/shared?meta=                200 OK + text file metadata
````

Johnetordoff: the above scenario is what happens by default  already.

felliott: The last case is the one failing for me.  The file I uploaded to GDrive is called `shared`, not `shared.txt`, so I should be able to access it via `GET .../googledrive/shared?meta=`.  If I do that with this PR, I still get:
````
{
  "code": 404,
  "message": "Could not retrieve file or directory /shared"
}
````

Johnetordoff: if I upload a txt file with no extension to gdrive it gives it and extension automatically, but if I rename it (using gdrive UI) to remove the extension then run these queries it works correctly as described above. 

How exactly are you getting extensionless files into gdrive? are you renaming them?

felliott: Nope, just uploading them via the GDrive web interface.

Johnetordoff: I realized this is actually part of a larger bug, I created a text file with nano (pages was hiding the extension) and I can't upload it to googledrive on production. It also gives the wrong error message "Error <br> Cannot upload folders." I'm not sure how to go about fixing this without making users always require an extension.

felliott:
Merged as de340a7adb63d57fc94a63d3d99557a4d30bbd91.

PR #196, opened 2017-02-22
[SVCS-55] Stop Github uploads which cause empty commits

Johnetordoff:
# Purpose
When a "revision" is uploaded with no changes, waterbutler treats this like a real revision and creates an empty commit. This PR compares the new and old files to see if they are the same and skips creating a commit if they are the same.

# Changes
Changes code in _fetch_content to include content is a way which is comparable to old content then compared the two and immediately returns metadata if they are the same. Edit: There's now new methods for fetching and checking git info.

# Side Effects
Though I couldn't see any immediate difference I think this could could cause performance decreases, because it forces another api call.

# Ticket
https://openscience.atlassian.net/browse/SVCS-55

### DIFF
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -2,7 +2,6 @@ import copy
 import json
 
 import furl
-import base64
 
 from waterbutler.core import streams
 from waterbutler.core import provider
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -216,6 +216,10 @@ class GitHubProvider(provider.BaseProvider):
         return any(new_blob['sha'] == blob['sha'] and path.path == blob['path'] for blob in blob_tree['tree'])
 
 
+    async def is_tree_in_trees(self, new_tree, branch_ref):
+        blob_tree = await self.get_blob_tree(branch_ref)
+        return new_tree['sha'] == blob_tree['sha']
+
     async def upload(self, stream, path, message=None, branch=None, **kwargs):
         assert self.name is not None
         assert self.email is not None
@@ -930,12 +934,15 @@ class GitHubProvider(provider.BaseProvider):
         new_tree = await self._create_tree({'tree': old_tree})
 
         # Create a new commit which references our top most tree change.
-        new_head = await self._create_commit({
-            'tree': new_tree['sha'],
-            'parents': [old_head['sha']],
-            'committer': self.committer,
-            'message': commit_msg,
-        })
+        if await self.is_tree_in_trees(new_tree, branch_ref):
+            return None
+        else:
+            new_head = await self._create_commit({
+                'tree': new_tree['sha'],
+                'parents': [old_head['sha']],
+                'committer': self.committer,
+                'message': commit_msg,
+            })
 
         # Update repository reference, point to the newly created commit.
         # No need to store data, rely on expects to raise exceptions
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -202,6 +202,20 @@ class GitHubProvider(provider.BaseProvider):
 
         return streams.ResponseStreamReader(resp, size=data.size)
 
+    async def get_blob_tree(self, branch_ref):
+
+        resp = await self.make_request(
+            'GET',
+            self.build_repo_url('git', 'trees') + '/{}:?recursive=99999'.format(branch_ref),
+        )
+
+        return await resp.json()
+
+    async def is_blob_in_tree(self, new_blob, path):
+        blob_tree = await self.get_blob_tree(path.branch_ref)
+        return any(new_blob['sha'] == blob['sha'] and path.path == blob['path'] for blob in blob_tree['tree'])
+
+
     async def upload(self, stream, path, message=None, branch=None, **kwargs):
         assert self.name is not None
         assert self.email is not None
@@ -241,13 +255,12 @@ class GitHubProvider(provider.BaseProvider):
             }]
         })
 
-        if exists:  # Avoids empty commits
-            if [x for x in tree['tree'] if x['path'] == path.name and blob['sha'] == x['sha']]:
-                return GitHubFileTreeMetadata({
-                    'path': path.path,
-                    'sha': blob['sha'],
-                    'size': stream.size,
-                }, ref=path.branch_ref), not exists
+        if exists and await self.is_blob_in_tree(blob, path):  # Avoids empty commits
+            return GitHubFileTreeMetadata({
+                'path': path.path,
+                'sha': blob['sha'],
+                'size': stream.size,
+            }, ref=path.branch_ref), not exists
 
         commit = await self._create_commit({
             'tree': tree['sha'],
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -234,8 +234,8 @@ class GitHubProvider(provider.BaseProvider):
         """
 
         blob_tree = await self.get_blob_tree(path.branch_ref)
-        return any(new_blob['sha'] == blob['sha']
-                   and path.path == blob['path'] for blob in blob_tree['tree'])
+        return any(new_blob['sha'] == blob['sha'] and
+                   path.path == blob['path'] for blob in blob_tree['tree'])
 
     async def upload(self, stream, path, message=None, branch=None, **kwargs):
         assert self.name is not None
@@ -952,7 +952,7 @@ class GitHubProvider(provider.BaseProvider):
 
         # Create a new commit which references our top most tree change.
 
-        if new_tree['sha'] == old_tree['sha']: # prevents empty commits
+        if new_tree['sha'] == old_tree['sha']:  # prevents empty commits
             return None
         else:
             new_head = await self._create_commit({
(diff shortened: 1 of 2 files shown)


waterbutler/providers/github/provider.py
```diff
@@ -241,6 +241,15 @@ def can_intra_copy(self, other, path=None):
...
 
+        if exists:  # Avoids empty commits
+            old_blob = await self._fetch_contents(path, path.branch_ref)
+            if old_blob['sha'] == blob['sha']:
+                return GitHubFileTreeMetadata({
+                    'path': path.path,
+                    'sha': blob['sha'],
+                    'size': stream.size,
+                }, ref=path.branch_ref), not exists
+
```
felliott: I think this can be done without the extra API call.  You already have the previous tree sha for the last commit (`latest_sha`, I hate past-me's naming skills).  The call to `_create_tree` gets a new tree sha with the changes applied.  If `old_tree_sha == new_tree_sha` then by definition, the tree contents are the same.  You can then skip the calls to `_create_commit` and `_update_ref`.

Example:  if you clone my test repo, https://github.com/felliott/wb-testing, to your local machine and look at the contents of commits b150f812 and bb8e7048:

````bash
fitz@teodor:~/code/services/wb-test-repos/wb-testing$ git cat-file commit b150f81
tree f02e193693e8826d336a6340d6abac592a02abbc
parent bbe4fa427641df82b350fbd4de5e959ddd1570ee
author Fitz Elliott <felliott@fiskur.org> 1487963691 -0500
committer Fitzfork <3rzwk@osf.io> 1487963691 -0500

File updated on behalf of WaterButler
fitz@teodor:~/code/services/wb-test-repos/wb-testing$ git cat-file commit bb8e7
tree f02e193693e8826d336a6340d6abac592a02abbc
parent b150f812fb23362e4ab4b10d18858b42a6c87ca3
author Fitz Elliott <felliott@fiskur.org> 1487963733 -0500
committer Fitzfork <3rzwk@osf.io> 1487963733 -0500

File updated on behalf of WaterButler
````

The trees are the same, only the parent commit is different.

Comment (felliott) [inline waterbutler/providers/github/provider.py]: Also, the same thing happens in the copy code.  If I have a file `/aaa-01.txt` and a duplicate of it in a subfolder `/duplicate/aaa-01.txt`, then if I copy one over the other, I'll get an empty commit.

````
POST .../github/duplicate/aaa-01.txt
{
  "action": "copy",
  "path": "/",
  "conflict": "replace"
}
````

The solution should be the same as for `upload`, just check the new tree sha against the old tree sha and skip creating commit and updating the branch if they are the same.  I've got an example of that in my wb-testing repo too:

````bash
fitz@teodor:~/code/services/wb-test-repos/wb-testing$ git cat-file commit 621dbc
tree 968001e8f154b105b74185407593e00466738999
parent bf41ea30686700d32c0079ee7c5c9682f19f63d2
author Fitz Elliott <felliott@fiskur.org> 1487963978 -0500
committer Fitzfork <3rzwk@osf.io> 1487963978 -0500

Copied on behalf of WaterButler
fitz@teodor:~/code/services/wb-test-repos/wb-testing$ git cat-file commit bf41ea3
tree 968001e8f154b105b74185407593e00466738999
parent ca587ef997ed5182200eef7053544be248aff735
author Fitz Elliott <felliott@fiskur.org> 1487963870 -0500
committer Fitzfork <3rzwk@osf.io> 1487963870 -0500

File uploaded on behalf of WaterButler
````

File: waterbutler/providers/github/provider.py [hunk 2]
```diff
@@ -241,6 +241,14 @@ def can_intra_copy(self, other, path=None):
             }]
         })
 
+        if exists:  # Avoids empty commits
+            if [x for x in tree['tree'] if x['path'] == path.name and blob['sha'] == x['sha']]:
```
felliott: Nice job here.  I gave you bad advice.  I was thinking the previous root tree sha was available, but it doesn't look like it is.  I guess we do pull it down in `validate_v1_path`, so saving that might be a good future optimization (probably after v0's retirement).

felliott: The Copy/move code needs to be updated to do the same thing.  That code lives [here](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/github/provider.py#L765), [here](https://github.com/Johnetordoff/waterbutler/blob/e326d89b13f880e664217af561ba53ff76738106/waterbutler/providers/github/provider.py#L781) and [here](https://github.com/Johnetordoff/waterbutler/blob/e326d89b13f880e664217af561ba53ff76738106/waterbutler/providers/github/provider.py#L797)

I've done some pre-digging to make sure (this time) I'm not giving you bad advice.  For these, the old root tree sha IS available in the `src_tree` and `dest_tree` objects, so all of these can probably be fixed in `_commit_tree_and_advance_branch` if the first parameter is updated to take the full tree metadata object.

Johnetordoff: The copy/move code can't create empty commits because any move is going to affect the tree, you can't trigger a move without changing the tree.

felliott: Move's can't avoid changing the tree, but copying can.  When you copy, you can overwrite an existing file.  If you overwrite a file with identical content, that doesn't change the tree.

See https://github.com/felliott/wb-testing/commit/b653aac04e9ef016f0cc7d17304572c19d900583  That was caused by copying https://github.com/felliott/wb-testing/blob/master/duplicate/aaa-01.txt overtop https://github.com/felliott/wb-testing/blob/master/aaa-01.txt.  If you check the trees for `b653aac0` and `b653aac0^`, they're the same (`968001e8f154b105b741`).

Johnetordoff: I don't really understand what copying is in this context (as a opposed to moving), can you do it from the UI? When does it happen?

felliott: Copying is a hidden feature of Fangorn / public feature of the v1 API.  In Fangorn if you hold down `option` while moving a file, it'll do a copy instead, which leaves the original file in place.  In the API this is done like so:
````
POST http://localhost:7777/v1/resources/shdpa/providers/github/duplicates/aaa-01.txt
{
  "action": "copy",
  "path": "/",
  "conflict": "replace"
}
````

This will copy `/duplicates/aaa-01.txt` over `/aaa-01.txt` without modifying or removing `/duplicates/aaa-01.txt`.

File: waterbutler/providers/github/provider.py [hunk 3]
```diff
@@ -202,6 +202,24 @@ def can_intra_copy(self, other, path=None):
 
         return streams.ResponseStreamReader(resp, size=data.size)
 
+    async def get_blob_tree(self, branch_ref):
```
felliott: Please add a docstring to this, similar to `_commit_tree_and_advance_branch`.

waterbutler/providers/github/provider.py
```diff
@@ -202,6 +202,24 @@ def can_intra_copy(self, other, path=None):
...
+
+        resp = await self.make_request(
+            'GET',
+            self.build_repo_url('git', 'trees') + '/{}:?recursive=99999'.format(branch_ref),
+        )
```
felliott: This `make_request` needs an `expects` parameter.

waterbutler/providers/github/provider.py
```diff
@@ -202,6 +202,24 @@ def can_intra_copy(self, other, path=None):
...
+
+        return await resp.json()
+
+    async def is_blob_in_tree(self, new_blob, path):
```
felliott: Please add a docstring.

waterbutler/providers/github/provider.py
```diff
@@ -202,6 +202,24 @@ def can_intra_copy(self, other, path=None):
...
+        blob_tree = await self.get_blob_tree(path.branch_ref)
+        return any(new_blob['sha'] == blob['sha'] and path.path == blob['path'] for blob in blob_tree['tree'])
```
felliott: Minor: line wrap at 100 chars.

waterbutler/providers/github/provider.py
```diff
@@ -907,12 +934,15 @@ def _prune_subtrees(self, tree):
         new_tree = await self._create_tree({'tree': old_tree})
 
         # Create a new commit which references our top most tree change.
-        new_head = await self._create_commit({
-            'tree': new_tree['sha'],
-            'parents': [old_head['sha']],
-            'committer': self.committer,
-            'message': commit_msg,
-        })
+        if await self.is_tree_in_trees(new_tree, branch_ref):
```
felliott: There's no need to make another api call.  All of the places that call `_commit_tree_and_advance_branch` already know the old tree sha, they just need to be updated to pass it in.

waterbutler/providers/github/provider.py
```diff
@@ -202,6 +202,24 @@ def can_intra_copy(self, other, path=None):
...
+
+        return await resp.json()
+
+    async def is_blob_in_tree(self, new_blob, path):
+        blob_tree = await self.get_blob_tree(path.branch_ref)
+        return any(new_blob['sha'] == blob['sha'] and path.path == blob['path'] for blob in blob_tree['tree'])
+
+
+    async def is_tree_in_trees(self, new_tree, branch_ref):
+        blob_tree = await self.get_blob_tree(branch_ref)
```
felliott: Not needed.  See below.

felliott:
Thanks, @Johnetordoff!  Merged as de340a7adb63d57fc94a63d3d99557a4d30bbd91.

PR #199, opened 2017-03-14
[SVCS-229] Add children to folder metadata for intra_move and intra_copy

TomBaxter:
## Purpose:

[SVCS-229](https://openscience.atlassian.net/browse/SVCS-229)
WB API: OSFStorage intra_move/_copy does not return children

## Changes:

update  waterbutler/providers/osfstorage/provider.py
- def intra_move update to return children
- def intra_copy update to return children

## Side effects:

None

[#SVCS-229]

### DIFF
--- a/tests/providers/osfstorage/test_provider.py
+++ b/tests/providers/osfstorage/test_provider.py
@@ -1,6 +1,7 @@
 import os
 import io
 import time
+import json
 import asyncio
 from http import client
 from unittest import mock
@@ -16,6 +17,7 @@ from waterbutler.core import exceptions
 from waterbutler.server import settings
 from waterbutler.core.path import WaterButlerPath
 from waterbutler.providers.osfstorage import OSFStorageProvider
+from waterbutler.providers.osfstorage.metadata import OsfStorageFolderMetadata
 from waterbutler.providers.osfstorage.settings import FILE_PATH_COMPLETE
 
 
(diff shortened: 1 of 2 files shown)


TomBaxter:
I spent some time looking at possible tests for this. If I understand what I was looking at, there is a waterbutler.io API for prod osfstorage. I'm not sure how to properly test against that.

TomBaxter:
Test created. It turns out that a call to intra_copy does do one call each to src_prov._children_metadata and src.validate_v1_path, so the test is a little different then you requested.

tests/providers/osfstorage/test_provider.py
```diff
@@ -291,6 +334,49 @@ def file_lineage():
 
     assert aiohttpretty.has_call(method='GET', uri=url, params=params)
 
+@pytest.mark.asyncio
+@pytest.mark.aiohttpretty
+async def test_intra_copy_folder(provider_and_mock, provider_and_mock2,
```
felliott: Since `intra_copy` and `intra_move` use different code paths, can you add a test for `intra_move`, too.

tests/providers/osfstorage/test_provider.py
```diff
@@ -291,6 +334,49 @@ def file_lineage():
...
+    })
+    url, _, params = src_provider.build_signed_url('POST',
+                                        'https://waterbutler.io/hooks/copy/',
+                                        data=data)
+
+    body = {'path': '/folder1/', 'id': 'folder1', 'kind': 'folder', 'name': 'folder1'}
+    aiohttpretty.register_json_uri('POST', url, params=params, status=201,
+                                   body=body)
+
+    folder_meta, _ = await src_mock.intra_copy(dest_mock, src_path, dest_path)
```
felliott: Please add a test for the result of `exists` (the second return arg from `intra_copy`) to avoid the bug below.

tests/providers/osfstorage/test_provider.py
```diff
@@ -291,6 +334,49 @@ def file_lineage():
...
+@pytest.mark.aiohttpretty
+async def test_intra_copy_folder(provider_and_mock, provider_and_mock2,
+                                 children_response):
+    src_provider, src_mock = provider_and_mock
+    src_mock.intra_copy = src_provider.intra_copy
+
+    dest_provider, dest_mock = provider_and_mock2
+    dest_mock.nid = 'abcde'
+    dest_mock._children_metadata = utils.MockCoroutine(
+                                       return_value=children_response)
```
felliott: Not a big deal, but your editor's indent heuristics are *weird*.

tests/providers/osfstorage/test_provider.py
TomBaxter: Done.

tests/providers/osfstorage/test_provider.py
TomBaxter: Nah, just me. Fixed?

tests/providers/osfstorage/test_provider.py
TomBaxter: Done.

waterbutler/providers/osfstorage/provider.py
```diff
@@ -195,7 +199,11 @@ def can_intra_move(self, other, path=None):
         if data['kind'] == 'file':
             return OsfStorageFileMetadata(data, str(dest_path)), dest_path.identifier is None
 
-        return OsfStorageFolderMetadata(data, str(dest_path)), dest_path.identifier is None
+        folder_meta = OsfStorageFolderMetadata(data, str(dest_path))
+        dest_path = await dest_provider.validate_v1_path(data['path'])
+        folder_meta.children = await dest_provider._children_metadata(dest_path)
+
+        return folder_meta, dest_path.identifier is None
```
TomBaxter: Done.

PR #200, opened 2017-03-16
[SVCS-259] Files with conflicting hash maps return oldest modified date

Johnetordoff:
## Purpose

Files with the same hashes all have the same dates. They have the date of the first that was uploaded.

## Changes

This problem was apparently caused by the fact that files with the same hash inherit a lot of the metadata for thier twin, much of the metadata is over-written with new metadata, but date modified is left out.

## Side Effects
This requires another PR in osf.io to be fully effective. 
https://github.com/CenterForOpenScience/osf.io/pull/6986

## Ticket
https://openscience.atlassian.net/browse/SVCS-259

### DIFF
--- a/tests/providers/osfstorage/test_provider.py
+++ b/tests/providers/osfstorage/test_provider.py
@@ -515,6 +515,20 @@ class TestUploads:
         inner_provider.move.return_value = (utils.MockFileMetadata(), True)
         inner_provider.metadata.side_effect = exceptions.MetadataError('Boom!', code=404)
 
+        signed_url, _, params = provider.build_signed_url('GET', provider.build_url(path.identifier, revision=''))
+
+        metadata_resp = {
+            'name': 'OtherTest',
+            'path': '/OtherTest',
+            'kind': 'file',
+            'version': 10,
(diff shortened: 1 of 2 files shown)


felliott:
I'm not sure I understand this fix.  Changing `update_metadata` to unconditionally set the modified date to now() seems overly broad.  Why are we sending it the wrong modified date in the first place?  Shouldn't we fix it so that we send the right metadata to the OSF when we create the metadata record (here: https://github.com/CenterForOpenScience/waterbutler/pull/200/files#diff-927878a162eb1a1110dfc88d4cf2373dR330).  The metadata we're sending here is correct for some properties ('size', 'contentType'), but it's using the old `modified`.

Please let me know if there's something I'm overlooking here.  This one feels like it might be a bit tricky.

Cheers,
Fitz

Johnetordoff:
This line was the culprit: https://github.com/CenterForOpenScience/waterbutler/blob/f1d5f58f93ae7daccc1355ceb1f914a3df0ed074/waterbutler/providers/filesystem/provider.py#L118

PR #208, opened 2017-04-06
[SVCS-137] stop passing query params as kwargs

Johnetordoff:
## Purpose

Stops Waterbutler from passing query params as kwargs, this prevents 500 when params (like 'path') overwrite real params.

## Changes 

Minor refactoring

## Side Effects 

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/SVCS-137

### DIFF
--- a/waterbutler/server/api/v1/provider/__init__.py
+++ b/waterbutler/server/api/v1/provider/__init__.py
@@ -62,7 +62,7 @@ class ProviderHandler(core.BaseHandler, CreateMixin, MetadataMixin, MoveCopyMixi
 
         self.auth = await auth_handler.get(self.resource, provider, self.request)
         self.provider = utils.make_provider(provider, self.auth['auth'], self.auth['credentials'], self.auth['settings'])
-        self.path = await self.provider.validate_v1_path(self.path, **self.arguments)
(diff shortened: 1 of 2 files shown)


waterbutler/server/api/v1/provider/__init__.py
```diff
@@ -65,7 +65,7 @@ class ProviderHandler(core.BaseHandler, CreateMixin, MetadataMixin, MoveCopyMixi
         if method != 'post':
             self.auth = await auth_handler.get(self.resource, provider, self.request)
             self.provider = utils.make_provider(provider, self.auth['auth'], self.auth['credentials'], self.auth['settings'])
-            self.path = await self.provider.validate_v1_path(self.path, **self.arguments)
+            self.path = await self.provider.validate_v1_path(self.path, params=self.arguments)
```
cslzchen: This change requires that we change 1) the signature for `.validate_v1_path()` and `.validate_path()` for core and each provider, and 2) usage of `kwargs` in the two method.

birdbrained: use `query`

birdbrained:
Waiting on v0 removal.

PR #212, opened 2017-04-13
[SVCS-132] Factor out move/copy code

Johnetordoff:
## Purpose

If WB core provider the move and copy code repeats itself unnecessarily this factors it out and cleans up the functions substantially.

## Changes 
Moves the metrics, name handling and intra_move/copy code into a decorator which wraps the move and copy functions, also adds/updates some docstrings.

## Side Effects
None that I know of.

## Ticket

https://openscience.atlassian.net/browse/SVCS-132

### DIFF
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -65,6 +65,43 @@ def build_url(base, *segments, **query):
     return url.url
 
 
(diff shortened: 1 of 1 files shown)


birdbrained:
Need to make it not a decorator

PR #213, opened 2017-04-17
[SVCS-294] Add rename validations

Johnetordoff:
## Purpose

Currently you can rename a file or folder whatever you like, this creates a variety of error behaviors for the different providers. For Dropbox filenames with slashes create bad folders and files, for other providers the rename will just fail with no explanation, this fix alerts the user they are using an illegal character. 

## Changes

Add two static variables for the base provider FORBIDDEN_FILENAME_CHARS and FORBIDDEN_FOLDERNAME_CHARS which are used by validate_rename to bar the use of invalid characters. 

Also cleans up some superfluous code in movecopy.py

## Side Effects 

None that I know of.

## Tickets

https://openscience.atlassian.net/browse/SVCS-294

### DIFF
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -78,6 +78,8 @@ class BaseProvider(metaclass=abc.ABCMeta):
     """
 
     BASE_URL = None
+    FORBIDDEN_FILENAME_CHARS = []
+    FORBIDDEN_FOLDERNAME_CHARS = ['/']
 
     def __init__(self, auth, credentials, settings, retry_on={408, 502, 503, 504}):
         """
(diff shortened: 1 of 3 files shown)


waterbutler/core/provider.py
```diff
@@ -83,6 +83,8 @@ class BaseProvider(metaclass=abc.ABCMeta):
     """
 
     BASE_URL = None
+    FORBIDDEN_FILENAME_CHARS = []
+    FORBIDDEN_FOLDERNAME_CHARS = ['/']
```
birdbrained: Make this an abc property

birdbrained:
TODO: Double check this isn't supported, and clarify pr message a little bit.

PR #215, opened 2017-04-28
[SVCS-259] Sends correct date modified in to fangorn after upload

Johnetordoff:
## Purpose 

Currently metadata sent back in a response to fangorn after an upload reflects the date modified coming from the file system, not the actual revision metadata from osfstorage.

## Changes 

Updates metadata to return correct info in the response.

## Side Effects 

None that I know of.

## Tickets/ Related PRs
https://openscience.atlassian.net/browse/SVCS-259
https://github.com/CenterForOpenScience/osf.io/pull/7155

### DIFF
--- a/tests/providers/osfstorage/test_provider.py
+++ b/tests/providers/osfstorage/test_provider.py
@@ -159,6 +159,7 @@ def upload_response():
             'checkout': None,
             'md5': 'abcdabcdabcdabcdabcdabcdabcd',
             'sha256': '123123123123123123',
+            'modified': '4-28-2017',
         }
     }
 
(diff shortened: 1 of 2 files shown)


tests/providers/osfstorage/test_provider.py
```diff
@@ -159,6 +159,7 @@ def upload_response():
             'checkout': None,
             'md5': 'abcdabcdabcdabcdabcdabcdabcd',
             'sha256': '123123123123123123',
+            'modified': '4-28-2017',
```
chrisseto: Would it be more correct for this to be a ISO-8601 date, with time and timezone information?

TomBaxter: @chrisseto  Was looking at your comment on PR7155, will modified be correct? You seemed to be saying it won't be.

TomBaxter: NM, I see what you did.

PR #216, opened 2017-05-01
[SVCS-295] Add new docs for query params

Johnetordoff:
# Purpose
Waterbulter's use of query params is currently undocumented this fixes that.

# Changes 
Adds new info to .rst docs

# Side Effect
None

# Ticket

https://openscience.atlassian.net/browse/SVCS-295

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
 
 To delete a file or folder send a DELETE request to the delete link. Nothing will be returned in the response body. As a precaution against inadvertantly deleting the root folder, the query parameter ``confirm_delete`` must be set to ``1`` for root folder deletes. In addition, a root folder delete does not actually delete the root folder. Instead it deletes all contents of the folder, but not the folder itself.
 
-
+======================
 Magic Query Parameters
-----------------------
+======================
 
-* ``direct``: issuing a download request with a query parameter named ``direct`` indicates that WB should handle the download, even if a redirect would be possible (e.g. osfstorage and s3).  In this case, WB will act as a middleman, downloading the data from the provider and passing through to the requestor.
+
+========================
+Provider Handler Params
+========================
+
+These are the query parameters that apply to all providers along with the request method this is the where Waterbutler get the infomation it needs to know what operation to perform, whether to upload, download, move, rename .etc.
+
+meta
+----
+
+Indicates that WB should return metadata about the file or folder and not download the contents.
+
+**Type**: flag
+
+**Expected on**: GET requests against file paths
+
+**Interactions**:
+
+* revisions / versions: meta takes precedence.  File metadata is returned, not a revision list.
+
+* revision / version: These are honored and passed on the the metadata method.  version takes precendence over revision.
+
+**Notes**: The meta query param is not required to fetch folder metadata; a bare GET folder request suffices. To download a folder a zip param should be included.
+
+
+zip
+---
+This parameter tells Waterbutler to download a folder's content into a zip for download.
+
+**Type**: flag
+
+**Expected on**: GET requests against folder paths
+
+**Interactions**:
+
+* All other query parameters will be ignored.
+
+**Notes**: A GET request folder with no query params will return metadata, but the same request on a file will download it.
+
+
+kind
+----
+This param is used on uploads and indicates whether an item to upload is a file or a folder.
+
+**Type**: string ('file' or 'folder') defaults to file.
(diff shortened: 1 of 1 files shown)


docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
 Magic Query Parameters
-----------------------
+======================
+
+
+========================
+Provider Handler Params
+========================
+
+These are the query parameters that apply to all providers along with the request method this is the where Waterbutler get the infomation it needs to know what operation to perform, whether to upload, download, move, rename .etc.
```
TomBaxter: Period between "providers along"
Comma between "method this"
"gets" instead of "get" after "Waterbutler"

docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
+
+meta
+----
+
+Indicates that WB should return metadata about the file or folder and not download the contents.
```
TomBaxter: Is "meta=" correct for folders?

docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
+
+**Type**: flag
+
+**Expected on**: GET requests against file paths
+
+**Interactions**:
+
+* revisions / versions: meta takes precedence.  File metadata is returned, not a revision list.
+
+* revision / version: These are honored and passed on the the metadata method.  version takes precendence over revision.
```
TomBaxter: IIRC we are trying to deprecate either revision(s) or versions(s). I forget which. If so, we may not want to document the deprecated param.

docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
+
+**Expected on**: GET for file paths
+
+**Interactions**:
+
+* is overridden by the 'meta' query param, but shouldn't be used with other params.
+
+**Notes**:
+
+* Revision and version can be used interchangeably, comments within the code indicate version is preferred, but no reason is supplied.
```
TomBaxter: Not sure of the reason. But makes sense that we only use one. Whichever one we go with, we should probably only document that param. Baby steps towards deprecation.

docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
+
+**Expected on**: GET  file paths
+
+**Interactions**:
+
+* This is parameter of the download provider method.
+
+**Notes**:
+
+* Not used for all providers, currently only used for OwnCloud, Cloudfiles and S3
```
TomBaxter: Might be worth while noting that Cloudfiles = osfstorage(in production)

docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
+* Is only used for the osfstorage provider.
+
+**Notes**:
+
+* currently only used with MFR.
+
+
+confirm_delete
+--------------
+Certain providers; Figshare, Dropbox, Box, Github, S3, Google Drive and osfstorage need to include the parameter confirm_delete as equal 1 in order to delete a root folder. This is done to prevent inadvertant root deletion.
```
TomBaxter: Good work hunting down which providers actually implement this. This should be implemented for any provider that allows the deletion of folders. As such we should document this as required for all providers.

docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
+
+
+
+Auth Handler Params
+===================
+These query params are used to decide autorization and permissions for the user.
+
+cookie
+------
+This gives user the user's cerdentials to Waterbutler
```
TomBaxter: Not following? "gives user the user's"?

docs/api.rst
```diff
@@ -152,8 +152,231 @@ If a moved/copied file is overwriting an existing file, a 200 OK response will b
...
+**Type**: flag
+
+**Expected on**: GET for files or folders
+
+Notes: Only used internally for the Open Science Framework.
+
+
+Github Provider Params
+======================
+Github, unlike other non-git providers requires special params to indicate the requested resource specfically the Git related infomation such as branch or commit shas which non-git providers don't need.
```
TomBaxter: Might be more readable to shorten this up. No need to belabor the point that this is specific to github. The header tells the tale.

PR #218, opened 2017-05-01
[SVCS-236] Fix broken revision handling for Github

Johnetordoff:
# Purpose 

Stop the 404s that occur when attempting to view Github revisions.

# Changes

Uses Github's comparison API endpoint to tell on what branch a commit sha originated that gives us enough data to validation that path of the revision still exists in the branch's tree. 

# Side Effects

None that I know of.

# Ticket / Related PRs

https://openscience.atlassian.net/browse/SVCS-236
#193 
#190

### DIFF
--- a/tests/providers/github/test_provider.py
+++ b/tests/providers/github/test_provider.py
@@ -156,6 +156,94 @@ def create_folder_response():
         }
     }
 
+@pytest.fixture
+def comparision_metadata():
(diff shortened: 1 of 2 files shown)


abought:
Quick note from our scrum today: if we go with this approach, we will want to be sure to add unit tests before merging.

That said, I haven't reviewed for the basic approach yet. :)

PR #219, opened 2017-05-02
[SVCS-575] Track requests within OSF

Johnetordoff:
# Purpose

Dropbox and Github have a nice feature where they return request identifiers in their response headers. That way, when a request starts acting up you can contact their support with the ID. This enables WB to do the same thing. Now we can verify a particular request made from the browser is the same one in our logs and more easily track inter-service requests if we want.

# Changes

Simply adds a uniquely generated id to the header "X-WATERBUTLER-REQUEST-ID" when the request is being prepared.

# Side Effects
None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-575


felliott:
:+1:  Rebasing and merging.

PR #221, opened 2017-05-09
[SVCS-332][SVCS-292] Make Googledrive give folder children’s metadata on moves and copies

Johnetordoff:
# Purpose

On inter/intra moves/copies googledrive doesn't return the folder's children, this fixes that.

# Changes

Changes item serializer to recursively call the google drive api include children to metadata.

# Side Effects

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-332

### DIFF
--- a/waterbutler/providers/googledrive/provider.py
+++ b/waterbutler/providers/googledrive/provider.py
@@ -138,7 +138,7 @@ class GoogleDriveProvider(provider.BaseProvider):
         ) as resp:
             data = await resp.json()
 
-        return GoogleDriveFileMetadata(data, dest_path), dest_path.identifier is None
+        return await self._serialize_item(dest_path, data), dest_path.identifier is None
 
     async def intra_copy(self, dest_provider, src_path, dest_path):
         self.metrics.add('intra_copy.destination_exists', dest_path.identifier is not None)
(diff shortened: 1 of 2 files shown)


abought:
> Coverage decreased (-0.0003%) to 75.235% when pulling 37949be on Johnetordoff:googledrive-send-children-metadata into 753bdd2 on CenterForOpenScience:develop.

^ @felliott , these emails can be controlled by a setting in coveralls.

waterbutler/providers/googledrive/provider.py
```diff
@@ -178,7 +178,8 @@ def can_intra_copy(self, other, path=None):
             throws=exceptions.IntraMoveError,
         ) as resp:
             data = await resp.json()
-        return GoogleDriveFileMetadata(data, dest_path), dest_path.identifier is None
+
+        return await self._serialize_item(path, data), dest_path.identifier is None
```
felliott: For GoogleDrive `intra_copy()` only works on files, folders have to go through core `copy()` which already includes children.  `can_intra_copy()` tests this in its implementation, but it's easy to overlook.  I've added actual comments to `develop` to make it more visible.

waterbutler/providers/googledrive/provider.py
```diff
@@ -604,10 +622,10 @@ def _build_upload_metadata(self, folder_id, name):
...
                 resp_json = await resp.json()
-                full_resp.extend([
-                    self._serialize_item(path.child(item['title']), item, raw=raw)
-                    for item in resp_json['items']
-                ])
+
+                for item in resp_json['items']:
+                    full_resp.append(await self._serialize_item(path.child(item['title']), item))
+
                 built_url = resp_json.get('nextLink', None)
```
felliott: I think this will recursively include children of folders, which I don't think is our intended behavior.

Johnetordoff:
Instead of resolving the conflicts here, I'm closing this and restarting on this: https://github.com/CenterForOpenScience/waterbutler/pull/244 PR

PR #225, opened 2017-05-18
Add type annotations and mypy compliance

abought:
:squid: Ready for review, but watch for sea monsters (my first WB PR)

# Ticket 
https://openscience.atlassian.net/browse/SVCS-351

# Purpose
- Add type annotations to assist with debugging (core classes and some, but not all, providers)
- Provide an option (currently not on by default) to use mypy, a linter for gradual type hinting- catches weird usages of methods

# Notes for reviewers
To try this out, run `inv mypy`. 

In the future this can be used in Travis CI, but is currently hidden behind a flag by default:
`inv test --types`

Many of these types are a "best guess" and subject to evolve in the future. Python types are informational only and do not affect runtime, but tools such as mypy (and pycharm!) support automatic linting to catch errors during development.

I am new to waterbutler. If there is a better type to use, let me know. The ignore settings in mypy may be too aggressive and are worth revisiting in the future.

## Specific rules that we ignore
Mypy does not appear to support selective ignoring of certain rules. Therefore, ignore statements are scattered throughout the code more than I would wish. See notes and rationale:

- Mypy doesn't seem to do very well with functions that can return more than one type of thing. (so basically every call to metadata gets ignored) See:
`https://github.com/python/mypy/issues/1693`
  - This could be resolved using dependent types / `@overload`, if we some day refactor so that files and folders are represented via different classes.
- Method call signatures or return types are not consistent with parent classes. We routinely violate the Liskov substitution principle, of necessity as there is no uniform abstraction for all providers. See: `https://github.com/python/mypy/issues/1237`
  - There are a lot of ignores for this. It's terrible; I'm sorry.

# Summary of changes
- Add type annotations to code
- Update docstrings
- Update method signatures that did not reflect the common usages across all subclasses

# Testing notes
In theory this is a pure developer facing change. One or two minor changes were made to method call signatures.

Recommend watching for any new errors during common operations. I have done light manual testing consisting of uploads/downloads/renames/cross provider moves (osf storage to google drive).

### DIFF
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -1,16 +1,21 @@
-import os
+import aiohttp
 import http
 import json
+import os
+import typing
 
-from waterbutler.core import streams
-from waterbutler.core import provider
+
 from waterbutler.core import exceptions
+from waterbutler.core import provider
+from waterbutler.core import streams
+
 from waterbutler.core.path import WaterButlerPath
 
 from waterbutler.providers.box import settings
-from waterbutler.providers.box.metadata import BoxRevision
-from waterbutler.providers.box.metadata import BoxFileMetadata
-from waterbutler.providers.box.metadata import BoxFolderMetadata
+from waterbutler.providers.box.metadata import (BaseBoxMetadata,
+                                                BoxFileMetadata,
+                                                BoxFolderMetadata,
+                                                BoxRevision)
 
 
 class BoxProvider(provider.BaseProvider):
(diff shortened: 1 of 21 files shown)


waterbutler/providers/box/provider.py
```diff
@@ -1,16 +1,21 @@
-import os
-import http
+import aiohttp
+from http import HTTPStatus
 import json
+import os
+import typing
```
abought: Updated in 3705f2a

I haven't yet manually tested this; is there a page with suggestions on common operations to verify?

Some tests are passing locally but failing on travis; still investigating why. In particular, as part of the cleanup of abc stub classes, I added `async` keyword to a few methods of the base class (when all child classes were using it). Want to make sure this doesn't cause unintended side effects.

abought:
Initial manual testing. Merge conflicts with googledrive refactor; will update and revalidate before next week.

abought:
Merge conflicts resolved and basic retesting done. Let me know if there are additional checks requested!

PR #234, opened 2017-06-06
[SVCS-319] Fix video for Sarfari 

Johnetordoff:
## Purpose 

Video doesn't stream in the MFR on Sarfari it's simply disappears frame and all.

## Note 

When using docker the url `http://localhost:5000/<guid>/` will not render video with any browser, but `http://192.168.168.167:5000/<guid>/` will work fine.

## Changes 

The problem is that Safari is a very strange browser which only works if you send it the exact number of bytes it needs within arbitrary seeming ranges of data, so instead of just asking for the whole video right off the bat, it requests a range of two bytes, then it requests everything, (unless caching is disabled, this will cause it to keep making arbitrary sized range requests until it gets want it wants.)

The changes implemented do two things

1. Make the content-length header, content-type header, HTTP status codes  being sent more clear and fool-proof, most of it is taken for Tornado's StaticFileHandler which already handles request ranges.

2. Makes us able accept request ranges and deliver streams consistent with those ranges while still maintaining a chunked stream

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/SVCS-319


waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -76,17 +76,51 @@ class MetadataMixin:
...
-            self.set_header('Content-Range', stream.content_range)
-
-        if stream.content_type is not None:
-            self.set_header('Content-Type', stream.content_type)
+        request_range = None
+        range_header = self.request.headers.get("Range")
+        if range_header:
+            # As per RFC 2616 14.16, if an invalid Range header is specified,
+            # the request will be treated as if the header didn't exist.
+            request_range = tornado.httputil._parse_request_range(range_header)
```
AddisonSchiller: Lines 62-64 basically do this exact same thing. Is there some reason here to do it again?

waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -76,17 +76,51 @@ class MetadataMixin:
...
+                # download size; cap the endpoint at the actual file size.
+                end = size
+            # Note: only return HTTP 206 if less than the entire range has been
+            # requested. Not only is this semantically correct, but Chrome
+            # refuses to play audio if it gets an HTTP 206 in response to
+            # ``Range: bytes=0-``.
+            if size != (end or size) - (start or 0):
+                self.set_status(206)  # Partial Content
+                self.set_header("Content-Range",
+                                tornado.httputil._get_content_range(start, end, size))
```
AddisonSchiller: Nitpick: Since all these `if` statements aren't part of the same expression, put spaces between them

waterbutler/server/utils.py
```diff
@@ -77,20 +77,23 @@ class UtilMixin:
...
+        if request_range:
+            start, end = request_range
+            content = (await stream.read(end))[start:]
+        else:
+            content = await stream.read()
+
+        if isinstance(content, bytearray):
+            content = bytes(content)
+
+        chunks = [content[i:i + settings.CHUNK_SIZE] for i in range(0, len(content), settings.CHUNK_SIZE)]
```
AddisonSchiller: Is `self.bytes_downloaded` no longer used at all?

waterbutler/server/utils.py
```diff
@@ -77,20 +77,23 @@ class UtilMixin:
...
+
+        for chunk in chunks:
+            self.write(chunk)
+            try:
```
AddisonSchiller: The old try/catch block was around nearly the whole function.  Will any of these other operations, namely `stream.read` throw the `StreamClosedError` that should be caught by this block?

AddisonSchiller:
Few things to note:

- The case by case comments are great on this. 

- The large amount of ifs is a bit confusing and awkward. If there is anyway to cut back on that slightly, it might be worth looking at (But yeah might be too much of a pain.) So basically, is there a simpler way to do all those ifs/elses

- Also couple questions/things to look at in the comments (redefining request_range when its already given to you etc)

Back to you.

waterbutler/server/utils.py
Johnetordoff: It's used here: https://github.com/CenterForOpenScience/waterbutler/blob/280569c520d1111bfbca57a9bb793474db8f6765/waterbutler/server/api/v1/provider and also for v0

waterbutler/server/utils.py
Johnetordoff: flush is the only message that can produce a StreamClosedError in the manner noted.

waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -62,6 +62,8 @@ class MetadataMixin:
         if 'Range' not in self.request.headers:
             request_range = None
         else:
+            # As per RFC 2616 14.16, if an invalid Range header is specified,
```
cslzchen: I see these changes are ported from tornado itself: https://github.com/tornadoweb/tornado/blob/master/tornado/web.py#L2384

```python
class StaticFileHandler(RequestHandler):
    """A simple handler that can serve static content from a directory.
         ...
    """
    ...
    @gen.coroutine
        def get(self, path, include_body=True):
```

Looks good to me but this is a different implementation from what we have before. Need more input from @felliott and @icereval (including the changes below on `write_stream`.

Please correct me if wrong. The primary change is that we rely on `request_range` and stream read according to the range instead of the predefined `CHUNK_SIZE`. If `request_range` is not available, we set request range to the full length and stream read them all.

cslzchen: A follow up question is why can't we use `request_range` and `chunk` together? For example, if the range of `request_range` is larger than the size of the `chunk`, we still chunk read.

waterbutler/server/utils.py
cslzchen: As discussed with @Johnetordoff , this only tries to "chunk-ify". Without this limitation, the code still works.

waterbutler/server/utils.py
```diff
@@ -77,20 +77,23 @@ class UtilMixin:
...
-                    break
-                # Temp fix, write does not accept bytearrays currently
-                if isinstance(chunk, bytearray):
-                    chunk = bytes(chunk)
-                self.write(chunk)
-                self.bytes_downloaded += len(chunk)
-                del chunk
+    async def write_stream(self, stream, request_range=None):
+
+        if request_range:
```
cslzchen: As mentioned before, is it possible to do chunk read in both `if` and `else` statement?

waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -76,17 +78,45 @@ class MetadataMixin:
...
 
-        if getattr(stream, 'partial', None):
-            # Use getattr here as not all stream may have a partial attribute
-            # Plus it fixes tests
-            self.set_status(206)
-            self.set_header('Content-Range', stream.content_range)
-
-        if stream.content_type is not None:
-            self.set_header('Content-Type', stream.content_type)
+        size = stream.size
```
felliott: Since this code is taken directly from tornado, it would be both polite and helpful to give them credit.  Please include a comment saying where it was taken from, and include a link to the GH page (preferably a tagged version) with the code.

waterbutler/server/utils.py
```diff
@@ -77,20 +77,23 @@ class UtilMixin:
...
-                chunk = await stream.read(settings.CHUNK_SIZE)
-                if not chunk:
-                    break
-                # Temp fix, write does not accept bytearrays currently
-                if isinstance(chunk, bytearray):
-                    chunk = bytes(chunk)
-                self.write(chunk)
-                self.bytes_downloaded += len(chunk)
-                del chunk
+    async def write_stream(self, stream, request_range=None):
```
felliott: The chunking was there to keep from reading the entire downloaded file into memory.  We still want that.  Please add that back for both cases (when `request_range` is present and when it isn't.  When there's a request range, you'll need to do some extra bookkeeping to make sure the last chunk doesn't exceed the total range size.

waterbutler/server/utils.py
felliott: You're no longer updating `bytes_downloaded`, meaning our metrics will always report 0 bytes being downloaded.

waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -76,17 +78,45 @@ class MetadataMixin:
...
+        elif end is not None:
+            content_length = end
+        elif start is not None:
+            content_length = size - start
+        else:
+            content_length = size
 
-        if stream.size is not None:
-            self.set_header('Content-Length', str(stream.size))
+        self.set_header("Content-Length", content_length)
```
felliott: I think `content_length` can still be `None` if a request range isn't provided and the stream size isn't known.  You either need an if guard here or a comment explaining why it couldn't possibly be `None`.

Johnetordoff: I'm not sure we can have a scenario where the stream size isn't know without it failing before it gets to this point. Size is a property here that always seems to always return a number, https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/core/streams/file.py#L18-L24.

felliott:
@Johnetordoff, minor process request: I've removed the Review Phase 2 label and added the Add'l Dev label.  When you're ready for review, can you remove Add'l Dev and add back Review Phase 2?  If you can't update the labels, ping me and I'll figure out something.  Thank you!

Johnetordoff:
@felliott I can't update the labels.

felliott:
Thanks, @Johnetordoff!  This was amazing work, and you did a great job of identifying the root cause.  There turned out to be an underlying bug in WB's Range header handling. I've opted to try to fix that issue in #318 which will supersede this PR.  Closing for now, but may revisit for inspiration on how to deal with providers that don't handle the Range header.

PR #235, opened 2017-06-07
[SVCS-249] Purge failed osfstorage uploads from pending directory

Johnetordoff:
## Purpose

In osfstorage, a copy of a file that fails to be successfully uploaded to the remote data storage is
left behind in /tmp/pending/ and will continue to take up disk space until manually removed. This PR makes sure that file is deleted before the error is thrown to prevent pending files from accumulating.

## Changes

Simple Python magic, and new unit tests.

## Side Effects

Partially uploaded files will be deleted.

## Ticket

https://openscience.atlassian.net/browse/SVCS-249


tests/providers/osfstorage/test_provider.py
```diff
@@ -18,7 +18,7 @@
 from waterbutler.core.path import WaterButlerPath
 from waterbutler.providers.osfstorage import OSFStorageProvider
 from waterbutler.providers.osfstorage.metadata import OsfStorageFolderMetadata
-from waterbutler.providers.osfstorage.settings import FILE_PATH_COMPLETE
+from waterbutler.providers.osfstorage.settings import FILE_PATH_COMPLETE, FILE_PATH_PENDING
```
AddisonSchiller: Nitpick: Import style 
from ... import( thing1,
    thing2)
etc.

tests/providers/osfstorage/test_provider.py
```diff
@@ -568,3 +568,22 @@ def patch_tasks(self, monkeypatch):
...
+
+        path = WaterButlerPath('/newfile', _ids=('rootId', None))
+        url = 'https://waterbutler.io/{}/children/'.format(path.parent.identifier)
+        aiohttpretty.register_json_uri('POST', url, status=201, body=upload_response)
+
+        provider, inner_provider = provider_and_mock
+        inner_provider.metadata = utils.MockCoroutine(return_value=utils.MockFileMetadata())
+        inner_provider.upload.side_effect = Exception()
+
+        with pytest.raises(Exception):
```
AddisonSchiller: Is there a specific exception that should be raised? or a specific case?
Given the context of the change I can understand why there wouldn't be though.

waterbutler/providers/osfstorage/provider.py
```diff
@@ -316,9 +316,13 @@ def signed_request(self, *args, **kwargs):
...
         stream.add_writer('sha256', streams.HashStreamWriter(hashlib.sha256))
 
-        with open(local_pending_path, 'wb') as file_pointer:
-            stream.add_writer('file', file_pointer)
-            await provider.upload(stream, remote_pending_path, check_created=False, fetch_metadata=False, **kwargs)
+        try:
+            with open(local_pending_path, 'wb') as file_pointer:
+                stream.add_writer('file', file_pointer)
+                await provider.upload(stream, remote_pending_path, check_created=False, fetch_metadata=False, **kwargs)
+        except Exception as ex:
```
AddisonSchiller: Is there a specific set of exceptions we should be looking for?
Nitpick:  in WB the general form of this is either `Exception as e` or `Exception as exc`. While this is mostly used in testing, its nice to have uniformity. Pick whichever one you fancy (unless say @felliott  has a preference).

waterbutler/providers/osfstorage/provider.py
```diff
@@ -316,9 +316,13 @@ def signed_request(self, *args, **kwargs):
         stream.add_writer('sha1', streams.HashStreamWriter(hashlib.sha1))
         stream.add_writer('sha256', streams.HashStreamWriter(hashlib.sha256))
 
-        with open(local_pending_path, 'wb') as file_pointer:
-            stream.add_writer('file', file_pointer)
-            await provider.upload(stream, remote_pending_path, check_created=False, fetch_metadata=False, **kwargs)
+        try:
+            with open(local_pending_path, 'wb') as file_pointer:
+                stream.add_writer('file', file_pointer)
+                await provider.upload(stream, remote_pending_path, check_created=False, fetch_metadata=False, **kwargs)
```
AddisonSchiller: since you are already touching this line, could you change this call so it adheres to our Line-Length requirements?

tests/providers/osfstorage/test_provider.py
```diff
@@ -568,3 +568,22 @@ def patch_tasks(self, monkeypatch):
         mock_backup.assert_called_once_with(complete_path, 'versionpk', 'https://waterbutler.io/hooks/metadata/', credentials['archive'], settings['parity'])
         inner_provider.metadata.assert_called_once_with(WaterButlerPath('/' + file_stream.writers['sha256'].hexdigest))
         inner_provider.move.assert_called_once_with(inner_provider, WaterButlerPath('/uniquepath'), WaterButlerPath('/' + file_stream.writers['sha256'].hexdigest))
+
+    @pytest.mark.asyncio
+    @pytest.mark.aiohttpretty
+    async def test_upload_fails(self, monkeypatch, provider_and_mock, file_stream, upload_response, mock_time):
```
AddisonSchiller: Line Length

tests/providers/osfstorage/test_provider.py
```diff
@@ -568,3 +568,22 @@ def patch_tasks(self, monkeypatch):
...
+
+        provider, inner_provider = provider_and_mock
+        inner_provider.metadata = utils.MockCoroutine(return_value=utils.MockFileMetadata())
+        inner_provider.upload.side_effect = Exception()
+
+        with pytest.raises(Exception):
+            await provider.upload(file_stream, path)
+
+        assert not os.path.isfile(FILE_PATH_PENDING + '/uniquepath')
+        inner_provider.upload.assert_called_once_with(file_stream, WaterButlerPath('/uniquepath'), check_created=False, fetch_metadata=False)
```
AddisonSchiller: Line Length

AddisonSchiller:
I keep forgetting to write my top level comment before submitting a review.... anyway
Couple line lengths/formatting changes
One small functionality question relating to exceptions thrown (but probably not super relevant)

LGTM after formatting stuff.

Oh also there is a conflict with develop to fix.

tests/providers/osfstorage/test_provider.py
cslzchen: Agreed. We can't pin down to specific exceptions in this case.

tests/providers/osfstorage/test_provider.py
```diff
@@ -700,20 +699,27 @@ def patch_tasks(self, monkeypatch):
...
-        path = WaterButlerPath('/' + upload_response['data']['name'],
-                               _ids=('Test', upload_response['data']['id']))
+        path = WaterButlerPath('/foopath', _ids=('Test', 'OtherTest'))
         url = 'https://waterbutler.io/{}/children/'.format(path.parent.identifier)
 
         mock_parity = mock.Mock()
         mock_backup = mock.Mock()
         inner_provider.move.return_value = (utils.MockFileMetadata(), True)
         inner_provider.metadata.side_effect = exceptions.MetadataError('Boom!', code=404)
+        body = {'version': 'versionpk',
```
cslzchen: I see a series of changes due to using a hard coded `body` instead of `upload_response`. Is it because the the old way no longer fit our needs. If so, why not update the fixtures `tests.providers.osfstorage.fixtures` by adding an entry?

cslzchen:
Update: consider a `rebase` on the latest develop as well. Thanks.

tests/providers/osfstorage/test_provider.py
Johnetordoff: This was started before we were using fixtures heavily, I'll just change it.

waterbutler/providers/osfstorage/provider.py
```diff
@@ -316,9 +316,15 @@ def signed_request(self, *args, **kwargs):
...
 
-        with open(local_pending_path, 'wb') as file_pointer:
-            stream.add_writer('file', file_pointer)
-            await provider.upload(stream, remote_pending_path, check_created=False, fetch_metadata=False, **kwargs)
+        try:
+            with open(local_pending_path, 'wb') as file_pointer:
+                stream.add_writer('file', file_pointer)
+                await provider.upload(stream, remote_pending_path, check_created=False,
+                                      fetch_metadata=False, **kwargs)
+        except Exception as ex:
```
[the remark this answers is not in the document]
Johnetordoff: @cslzchen For `os.remove()` how do we handle an error for that? A nested try? I'm not sure what the utility of catching is. What would I re-throw it as?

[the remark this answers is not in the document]
cslzchen: Here is a link to the python doc (3.6): https://docs.python.org/3.6/library/os.html#os.remove. The expected error is `OSError`. You made a good point since I realize I don't know what to do when we catch it. I will make this a no-blocker. 

FYI, from the code base, we don't `try ... except ...` similar methods often. My ideal way is to catch these, log only necessary information and throw an tornado or waterbutler level exception.

tests/providers/osfstorage/test_provider.py
```diff
@@ -700,20 +700,25 @@ def patch_tasks(self, monkeypatch):
 
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
-    async def test_upload_and_tasks(self, monkeypatch, provider_and_mock, file_stream,
```
cslzchen: Why do we change this test? Maybe something went wrong during merge conflicts? I reverted this change and removed unnecessary `request_body`.

cslzchen:
Pending secondary PR: https://github.com/Johnetordoff/waterbutler/pull/6

waterbutler/providers/osfstorage/provider.py
```diff
@@ -316,9 +315,19 @@ def signed_request(self, *args, **kwargs):
...
+            with open(local_pending_path, 'wb') as file_pointer:
+                stream.add_writer('file', file_pointer)
+                await provider.upload(stream, remote_pending_path, check_created=False,
+                                      fetch_metadata=False, **kwargs)
+        except Exception as exc:
+            try:
+                os.remove(local_pending_path)
+            except OSError as os_exc:
+                raise exceptions.UploadFailedError(
+                    'Upload failed and attempts to clean pending files failed:\n{}\n{}'.format(exc, os_exc)
```
cslzchen: Use `str(exc)` and `str(os_exc)`.

waterbutler/providers/osfstorage/provider.py
```diff
@@ -316,9 +315,19 @@ def signed_request(self, *args, **kwargs):
...
+                )
+            raise exceptions.UploadError('Upload failed and pending files cleaned:\n{}'.format(exc))
```
cslzchen: Similarly, use`str(exc)`.

felliott:
Looks good, merging with some minor changes.  I made the error messages *less* informative, because they get displayed to the user.  Instead of showing the underlying exception (which they can't fix), they get a message asking them to try again.  I suspect that a lot of these sorts of errors would come from network interruptions, so retrying would be a valid first step.  The debugging information in the exceptions will still be captured by Sentry and the console logs.

PR #236, opened 2017-06-08
[SVCS-314] WaterButler - S3 - Revisions broken

TomBaxter:
## Purpose:
[SVCS-314](https://openscience.atlassian.net/browse/SVCS-314)
Currently S3 revisions always download the latest version.
This PR fixes that.

## Changes:
Updated  waterbutler/providers/s3/provider.py def download accept arg revision instead of version
Updated  tests/providers/s3/test_provider.py def test_download_version use are revision instead of version

https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/files.py#L307  calls revision not version

## Side effects
None

[#SVCS-314]


TomBaxter:
@cslzchen Updated changes section with reference to the call that uses revision instead of version.

waterbutler/providers/s3/provider.py
```diff
@@ -135,7 +135,7 @@ def can_intra_move(self, dest_provider, path=None):
         await resp.release()
         return (await dest_provider.metadata(dest_path)), not exists
 
-    async def download(self, path, accept_url=False, version=None, range=None, **kwargs):
+    async def download(self, path, accept_url=False, revision=None, range=None, **kwargs):
```
cslzchen: For @felliott :

(At first, the variable renaming made no sense to me.) This works since the method is called with the `revision` instead of `version` in [osf/models/files.py](https://github.com/CenterForOpenScience/osf.io/blob/develop/osf/models/files.py#L307):

```python
resp = requests.get(
        self.generate_waterbutler_url(revision=revision, meta=True, _internal=True, **kwargs),
        headers=headers,
    )
```

felliott:
Lovely!  Moving to RTM.

PR #237, opened 2017-06-08
WB S3 provider: Assign resp.headers before releasing resp

TomBaxter:
While working SVCS-314 ran into issue during testing where I was receiving "Key not found: 'X-WATERBUTLER-METADATA'" error from s3/provider.py def _metadata_file . This PR fixes the issue.


TomBaxter:
@felliott this is the side issue from SVCS-314. There is no corresponding JIRA ticket.

Johnetordoff:
This could use a little more info on how to reproduce this, if this fails every time _file_metadata is called how did we not notice immediately?

PR #248, opened 2017-08-23
Adding "undecoded_path" to providers. This should be the first step f…

AddisonSchiller:
…ixing Issues with special characters in names

Pulled the request uri out of the tornado header, then split it based on the provider name and '?'
There may be a few edge cases where this splitting could be a problem, but since its the non-decoded path, it should be okay(shouldn't hit any problems with splitting on '?')

refs: https://openscience.atlassian.net/browse/SVCS-256

## Purpose
We need the un-decoded file name to distinguish between files with special characters. 
ie: "me/ow.txt and me%2Fow.txt" The decoding tornado does makes these files hard to distinguish.

## Summary of changes
Added an "undecoded_path" variable to the base_provider. This gets set in the "prepare" function to be the undecoded_path received from the request. The path is pulled from 'self.request.uri'

## QA
QA will come on individual tickets for providers


waterbutler/server/api/v1/provider/__init__.py
```diff
@@ -62,6 +62,9 @@ class ProviderHandler(core.BaseHandler, CreateMixin, MetadataMixin, MoveCopyMixi
 
         self.auth = await auth_handler.get(self.resource, provider, self.request)
         self.provider = utils.make_provider(provider, self.auth['auth'], self.auth['credentials'], self.auth['settings'])
+        # Find start of provider name, and start of ? marker and pull out file name from between them.
+        self.provider.undecoded_path = self.request.uri[self.request.uri.rfind(self.path_kwargs['provider'] + '/') + (len(self.path_kwargs['provider'])): self.request.uri.rfind("?")]
```
felliott: Line length!

felliott: Please break down the components of this indexing operation into individual variables with descriptive names.  This is hard to skim.

felliott: Indexing issue:

`http://localhost:7777/v1/resources/p8a63/providers/googledrive/me%2fow/provider/googledrive/beef?meta=`

`self.provider.undecoded_path` is `beef`.

felliott: Indexing issue:

`http://localhost:7777/v1/resources/p8a63/providers/googledrive/me%2fow/provider/googledrive/beef`

`self.provider.undecoded_path` is `bee`

AddisonSchiller:
Added requested Changes

Johnetordoff:
Might want to wrap up this conflict.

PR #249, opened 2017-08-29
Increasing coverage for dropbox provider and metadata pytests.

AddisonSchiller:
Wrote tests for intra_move, intra_copy, delete, revisions, various file and folder metadata calls, a few exceptions and a few operation functions.
Added various fixtures to accommodate these tests.
Moved the fixtures into a fixture file and out of the main test file.

refs: https://openscience.atlassian.net/browse/SVCS-437

## Purpose
increase the coverage of tests for the dropbox addon (provider, exceptions, metadata)

## Summary of changes
Wrote tests for intra_move, intra_copy, delete, revisions, various file and folder metadata calls, a few exceptions and a few operation functions.

moved fixture data to its own file. Put json data into its own seperate files under /fixtures. Fixtures are loaded from these files into batches. 

## Testing
Tests should complete, coverage should be much higher for drop box.

### DIFF
--- a/tests/providers/dropbox/fixtures.py
+++ b/tests/providers/dropbox/fixtures.py
@@ -1,0 +1,28 @@
+import pytest
+import os
+import json
+
+
+@pytest.fixture
+def root_provider_fixtures():
+    # fixtures for testing validate_v1_path for root provider
+    with open(os.path.join(os.path.dirname(__file__), 'fixtures/root_provider.json'), 'r') as fp:
+        return json.load(fp)
+
+
+@pytest.fixture
+def revision_fixtures():
+    with open(os.path.join(os.path.dirname(__file__), 'fixtures/revisions.json'), 'r') as fp:
+        return json.load(fp)
+
+
+@pytest.fixture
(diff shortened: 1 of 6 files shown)


Johnetordoff:
Is loading the fixtures as JSON something that's been discussed? I'm see the benefits, but it might be something to do later.

AddisonSchiller:
@Johnetordoff Fitz loaded them this way for one-drive, and he asked me to do the same for dropbox, and I assume the other ones working on. My set up for loading the json may not be exactly like his and It may need a quick spruce up at some point to conform to some sort of standard. 

There are a few others that load them a similar way, but isn't exactly the same.

Johnetordoff:
That's great I had to collect a lot of these JSON response for metadata, here https://osf.io/t8ry5/ might save you some time for Gdrive etc. I'll make you a contributor if you want.

AddisonSchiller:
That should be pretty useful. I'll take a look at it. I probably won't need to be a contributor, but i'll let you know.
Thanks @Johnetordoff !

PR #220, opened 2017-05-05
[SVCS-327][SVCS-292] Make Box give folder children’s metadata on moves and copies

Johnetordoff:
# Purpose

On inter/intra moves/copies Box doesn't return the folder's children, this fixes that.

# Changes

Changes item serializer to include children to metadata.

# Side Effects

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-292

PR #253, opened 2017-09-05
[SVCS-327][SVCS-292] Make Box give folder children’s metadata on moves and copies

Johnetordoff:
# Purpose

On inter/intra moves/copies Box doesn't return the folder's children, this fixes that.

# Changes

Changes item serializer to include children to metadata.

# Side Effects

None that I know of.

# Ticket

https://openscience.atlassian.net/browse/SVCS-292

# Old PR

https://github.com/CenterForOpenScience/waterbutler/pull/220

### DIFF
--- a/tests/providers/box/test_provider.py
+++ b/tests/providers/box/test_provider.py
@@ -594,33 +594,61 @@ class TestIntraCopy:
 
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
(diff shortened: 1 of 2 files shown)


Johnetordoff:
There are no unit tests for this. They are to included in https://github.com/CenterForOpenScience/waterbutler/pull/252 which aren't updated for this behavior, not sure how are should reconcile this, probably not too problematic, but heads up @AddisonSchiller .

PR #261, opened 2017-09-14
[SVCS-447] Add owncloud test coverage

Johnetordoff:
## Purpose 

This PR will bring up our Owncloud provider test coverage significantly, enabling us to better handle changes in the future.

## Changes 

Adds coverage for intra_move/copy and steps up coverage for edge cases in other areas.

## Side Effects 

I found a small bug while running tests, it's small enough I decided to address it in this PR, also it would be difficult to get complete coverage without fixing this first.

## Ticket

https://openscience.atlassian.net/browse/SVCS-447

### DIFF
--- a/tests/providers/owncloud/test_provider.py
+++ b/tests/providers/owncloud/test_provider.py
@@ -10,8 +10,9 @@ from waterbutler.core import exceptions
 from waterbutler.core.path import WaterButlerPath
 
 from waterbutler.providers.owncloud import OwnCloudProvider
-from waterbutler.providers.owncloud.metadata import OwnCloudFileMetadata
-from waterbutler.providers.owncloud.metadata import OwnCloudFolderMetadata
+from waterbutler.providers.owncloud.metadata import OwnCloudFileMetadata, OwnCloudFileRevisionMetadata
+
+from tests import utils
 
 @pytest.fixture
 def auth():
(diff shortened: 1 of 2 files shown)


tests/providers/owncloud/test_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+from waterbutler.providers.owncloud.metadata import OwnCloudFileRevisionMetadata
+
+
+
+class TestFileMetadata:
+
+    def test_file_metadata(self, file_metadata_object):
+        assert file_metadata_object.provider == 'owncloud'
+        assert file_metadata_object.name == 'dissertation.aux'
+        assert file_metadata_object.path == ''
```
felliott: Why are all these paths `''`?  That shouldn't be possible.

tests/providers/owncloud/test_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+        assert file_metadata_object.size == '3011'
+        assert file_metadata_object.etag == '&quot;a3c411808d58977a9ecd7485b5b7958e&quot;'
+        assert not file_metadata_object.created_utc
+        assert file_metadata_object.content_type == 'test-type'
+        assert file_metadata_object.extra == {}
```
felliott: Where are `kind`, `modified`, `modified_utc`, and `materialized`?

tests/providers/owncloud/test_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_revision_metadata(self, revision_metadata_object):
+        assert revision_metadata_object.version_identifier == 'revision'
+        assert revision_metadata_object.version == 'latest'
+        assert revision_metadata_object.modified == 'Sun, 10 Jul 2016 23:28:31 GMT'
+        assert revision_metadata_object.extra == {}
+
+    def test_revision_metadata(self, revision_metadata_object, file_metadata_object):
+
+            revision = OwnCloudFileRevisionMetadata.from_metadata(file_metadata_object)
+            assert revision == revision_metadata_object
```
felliott: Nitpick: these two lines are over-indented.

tests/providers/owncloud/test_metadata.py
```diff
@@ -0,0 +1,105 @@
...
+
+class TestRevisionMetadata:
+
+    def test_revision_metadata(self, revision_metadata_object):
+        assert revision_metadata_object.version_identifier == 'revision'
+        assert revision_metadata_object.version == 'latest'
+        assert revision_metadata_object.modified == 'Sun, 10 Jul 2016 23:28:31 GMT'
+        assert revision_metadata_object.extra == {}
+
+    def test_revision_metadata(self, revision_metadata_object, file_metadata_object):
```
felliott: Wait a second, this test has the same name as the previous test.  That means the latter one is always run and the former is always skipped.

tests/providers/owncloud/test_provider.py
```diff
@@ -269,7 +179,35 @@ class TestCRUD:
         aiohttpretty.register_uri('PROPFIND', url, body=file_metadata, auto_length=True, status=207)
         aiohttpretty.register_uri('PUT', url, body=b'squares', auto_length=True, status=201)
         metadata, created = await provider.upload(file_stream, path)
-        expected = OwnCloudFileMetadata('dissertation.aux','/owncloud/remote.php/webdav/Documents/phile',
+        expected = OwnCloudFileMetadata('dissertation.aux',
```
felliott: Minor: seems like this could use one of the fixtures you created.

tests/providers/owncloud/test_provider.py
```diff
@@ -269,7 +179,35 @@ class TestCRUD:
...
+        url = provider._webdav_url_ + renamed_path.full_path
+
+        aiohttpretty.register_uri('PROPFIND', url, body=file_metadata, auto_length=True, status=207)
+        aiohttpretty.register_uri('PUT',
+                                  provider._webdav_url_ + '/my_folder/phile (1)',
+                                  body=b'squares',
+                                  auto_length=True,
+                                  status=201)
+        metadata, created = await provider.upload(file_stream, path, 'keep')
+        expected = OwnCloudFileMetadata('dissertation.aux',
```
felliott: Can you use a fixture here?

tests/providers/owncloud/test_metadata.py
Johnetordoff: These aren't attributes of OwnCloudFileRevisionMetadata.

felliott: They are attributes of OwnCloudFileMetadata, which is what's being tested here.   I'll fix this up locally.

PR #262, opened 2017-09-15
[SVCS-464] Dropbox upgrade from deprecating endpoints to V2

Johnetordoff:
## Purpose

Dropbox is deprecating some endpoints we use and upgrading them to "_v2" endpoints, this PR changes are code to accommodate the "_v2"s

## Changes 

This PR amounts to a refactor with no major changes.

The effected endpoints were.

- move -> move_v2
- copy -> copy_v2
- delete -> delete_v2
- create_folder -> create_folder_v2

All endpoints were effected in essentially the same way. They had a minor change in metadata responses. Dropbox has changed their metadata responses from this: 

`{ ... }`

to this:

`{ 'metadata': { ... } }`

Not a huge change but just enough to break every function which receives a JSON response. So I had to add a line of code unpacking the JSON for each effected method. 

The unit tests were also refactored and fixtures were added. I left in the old fixtures in case they are needed in the future.

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/SVCS-464


waterbutler/providers/dropbox/provider.py
```diff
@@ -176,7 +176,7 @@ def default_headers(self) -> dict:
                     expects=(200, 201, 409),
                     throws=exceptions.IntraCopyError,
                 )
-                data = data['metadata']
+            data = data['metadata']
```
cslzchen: Is this change intentional? `data` is not defined if the `if` statement is `False` though. Should it be within the `if`?

Another question is why `intra_copy` used `data['metadata']` already before the upgrade but `intra_move` used `data` and `create_folder` did not.

Johnetordoff: The above change is intentional and is necessary to avoid a copy fail. As for why certain things used the metadata key before, I don't know, it seems like an arbitrary decision on Dropbox's behalf.

cslzchen: I think I overlooked the `else` statement that is collapsed during the first pass.

tests/providers/dropbox/test_provider.py
```diff
@@ -509,34 +509,35 @@ class TestIntra:
...
         src_path = WaterButlerPath('/pfile', prepend=provider.folder)
         dest_path = WaterButlerPath('/pfile_renamed', prepend=provider.folder)
 
-        url = provider.build_url('files', 'copy')
+        url = provider.build_url('files', 'copy_v2')
         data = {'from_path': src_path.full_path.rstrip('/'),
                 'to_path': dest_path.full_path.rstrip('/')},
         aiohttpretty.register_json_uri('POST', url, data=data,
-                                       body=root_provider_fixtures['file_metadata'])
+                                       body=intra_copy_fixtures['intra_move_copy_file_metadata_v2'])
```
cslzchen: Interesting. I understand using `intra_copy_fixtures` instead.

Not sure why we used `root_provider_fixtures['file_metadata']` instead of `intra_copy_fixtures['intra_copy_file_metadata_v2']`.

tests/providers/dropbox/test_provider.py
```diff
@@ -587,23 +590,27 @@ class TestIntra:
...
+                                       body=intra_copy_fixtures
+                                       ['intra_move_copy_folder_metadata_v2'])
 
         url = provider.build_url('files', 'list_folder')
         data = {'path': dest_path.full_path}
         aiohttpretty.register_json_uri('POST', url, data=data,
                                        body=root_provider_fixtures['folder_children'], status=200)
 
         result = await provider.intra_copy(provider, src_path, dest_path)
-        expected = DropboxFolderMetadata(root_provider_fixtures['folder_metadata'], provider.folder)
```
cslzchen: `root_provider_fixtures` has a `fodler_metadata_v2`. Can it be merged with `intra_move_copy_folder_metadata_v2`?

Johnetordoff: It probably won't effect current tests, but it's worth having the differences documented IMO, we have similar behavior where the metadata returned my moves/copies isn't automatically the same as metadata returned by a metadata request.

tests/providers/dropbox/test_provider.py
Johnetordoff: These fixtures are "close enough" that they both pass tests, but they are mildly different. I'm differentiating them just so we are aware of this. Metadata returned by move/copies and that returned by metadata requests are often different and this has led to bugs in the past.

PR #265, opened 2017-09-20
[SVCS-397] Copying a root without rename does not 500

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-397
## Purpose
Currently if you try to copy root, it will error out and 500. Copying non root files does not require rename. Code needs to be added so copying root without rename does not 500, and instead throws an appropriate error message

## Summary of changes
Added a check so that if someone is attempting to copy root, rename needs to be set. If it is not set, throw an InvalidParameters error.


Johnetordoff:
This might be a good opportunity to write some tests for this situation.

AddisonSchiller:
@Johnetordoff 
There wasn't a good test file for the copy/move. The file itself had very low coverage. Since I wasn't sure how to properly set up for that sort of testing, it started to feel scope creepy. I should have made a note of it, however.

waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -91,6 +91,9 @@ def build_args(self):
 
             # TODO optimize for same provider and resource
 
+            if action in ['copy'] and not self.json.get('rename') and self.path.is_root:
```
Johnetordoff: `action in ['copy']` not sure why you would do this for a list with a single element.

waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -91,6 +91,9 @@ def build_args(self):
...
+                raise exceptions.InvalidParameters('Rename is required for copying root')
```
Johnetordoff: Be a little more specific with this message, the user may not know they need to add a new parameter, also I'm not sure but I think a full sentence is preferred in cases where we can correct user behavior.

AddisonSchiller: I just followed the format of the other exceptions raised in that same File. Is this something that should be changed for all exceptions in that file?

Johnetordoff: There are no examples of comparisons of this nature in movecopy.py that need to be changed.

Johnetordoff: You should change it to be more specific and helpful, indicate that this is a correctable problem.

AddisonSchiller: https://github.com/AddisonSchiller/waterbutler/blob/945b38a6af5596e328798fe3d60c9c414b982df5/waterbutler/server/api/v1/provider/movecopy.py#L77

As per my previous comment, here is an example of similar nature.
Could you give me an example of what you mean? Just in general my rule of thumb is to try to conform to the standards set in a file. Keep it consistent. If there are no other precedents set, look at surrounding files etc. Then do it your own way.

Johnetordoff: Oh, I got this confused with the previous comment on line 94.

cslzchen:
Replaced by https://github.com/CenterForOpenScience/waterbutler/pull/305 after https://github.com/CenterForOpenScience/waterbutler/pull/265 got merged in.

PR #273, opened 2017-09-25
[SVCS-421] Removing unsizable flag from ResponseStreamReader

### LINKED ISSUE #217: ResponseStreamReader does not handle chunked encoding
We've had an issue with Water Butler being unable to read streams with chunked encoding. 

If you look at  the constructor of ResponseStreamReader(BaseStream) in waterbutler/core/streams/http.py you will see that if size is None and unsizable is False, the attempt to turn size into an int will fail. This seems to happen with chunked encoding. As a workaround for our testing, we just set self._size to None in that circumstance.

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-421
closes #217 
## Purpose
The unsizable flag is breaking chunked encoding responses. It is also no longer used and can be removed.

## Summary of changes
Removed unsizable flag from ResponseStreamReader

## Test Notes/ QA
Git archeology showed that not much had used this flag, or was using it. It seems like googledrive was only ever using it, and that has been changed.

On the Jira ticket, there is a script which tests the difference between the old and new ResponseStreamReader (demonstrating the bug).


Johnetordoff:
LGTM 👍

waterbutler/core/streams/http.py
```diff
@@ -140,12 +140,10 @@ def _make_boundary_stream(self):
 
 class ResponseStreamReader(BaseStream):
 
-    def __init__(self, response, size=None, name=None, unsizable=False):
```
cslzchen: `size=None` and `unsizable=False` are used together, should we remove both? And please make sure that no one calls `ResponseStreamReader` with the removed kwargs.

cslzchen: Or maybe keep those two but improve conditional logic. However, we still need to make sure that `int(size)` does not throw exceptions (or we need to catch them when that happens). Whether `size` is of type `int` or `string` makes a difference.

```python
elif not unsizable and size is not None:
    try:
        elf._size = int(size)
    except:
        elf._size = None
```

Johnetordoff: So after a little research I've found the unsizable argument is never used, and the size argument is used, but seems to usually be None, so I've decided to change the code to match the suspected intention of original writer. When a size is given the stream will read to that size, if it is None, the whole stream will be read.
`    def __init__(self, response, size=None, name=None):
        super().__init__()
        if 'Content-Length' in response.headers:
            self._size = int(response.headers['Content-Length'])
        else:
            self._size = size

        self._name = name
        self.response = response
`

cslzchen:
PR closed. Work continues in https://github.com/CenterForOpenScience/waterbutler/pull/309.

PR #274, opened 2017-09-25
[SVCS-479]  Raise exception for copy/move replace folder that orphans itself.

TomBaxter:
## Purpose:
[SVCS-479](https://openscience.atlassian.net/browse/SVCS-479)
Replace folder from inside folder deletes both.

## Changes:
Updated  waterbutler/core/provider.py
-- Add def replace_will_orphan
-- Update def move calls replace_will_orphan
-- Update def copy calls replace_will_orphan
Update  waterbutler/core/exceptions.py
-- Add class OrphanSelfError
Update tests/providers/osfstorage/test_provider.py
-- Add test test_intra_copy_folder_orphan

## Side effects
Folder move/copies that use conflict=replace may now be more costly

### DIFF
--- a/tests/providers/osfstorage/test_provider.py
+++ b/tests/providers/osfstorage/test_provider.py
@@ -341,7 +341,6 @@ class TestIntraCopy:
         src_mock._children_metadata.assert_not_called()
         src_mock.validate_v1_path.assert_not_called()
 
-
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
     async def test_intra_copy_file(self, provider_and_mock, provider_and_mock2,
--- a/waterbutler/core/exceptions.py
+++ b/waterbutler/core/exceptions.py
@@ -208,6 +208,15 @@ class OverwriteSelfError(InvalidParameters):
                          'folder onto itself is not supported.'.format(path))
 
 
+class OrphanSelfError(InvalidParameters):
+    """Error for conflict=replace sent to methods copy or move, when action would result
+    in removal of a directory containing the file or folder to be copied or moved.
+    """
+    def __init__(self, path):
+        super().__init__('Unable to move or copy \'{}\'. Moving or copying a folder onto parent '
+                         'folder of same name with "replace" is not supported.'.format(path))
+
+
 class UnsupportedOperationError(ProviderError):
     def __init__(self, message, code=HTTPStatus.FORBIDDEN, is_user_error=True):
         if not message:
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -224,6 +224,18 @@ class BaseProvider(metaclass=abc.ABCMeta):
         })
 
         if handle_naming:
+            # In handle_naming so we don't run it multiple times
+            # This does not mean the `dest_path` is a folder, at this point it could just be the parent
+            # of the actual dest_path, or have a blank name
+            if not dest_path.is_file:
+                temp_path = await self.revalidate_path(
+                    dest_path,
+                    rename or src_path.name,
+                    folder=src_path.is_dir
+                )
+                if self.will_self_overwrite(dest_provider, src_path, temp_path):
+                    raise exceptions.OverwriteSelfError(src_path)
+
             dest_path = await dest_provider.handle_naming(
                 src_path,
                 dest_path,
@@ -269,7 +281,29 @@ class BaseProvider(metaclass=abc.ABCMeta):
             'conflict': conflict,
             'got_rename': rename is not None,
         })
+
+        if not dest_path.is_file:
+            temp_path = await self.revalidate_path(
+                dest_path,
+                rename or src_path.name,
+                folder=src_path.is_dir
+            )
+            if self.will_self_overwrite(dest_provider, src_path, temp_path):
+                raise exceptions.OverwriteSelfError(src_path)
+
         if handle_naming:
+            # In handle_naming so we don't run it multiple times
+            # This does not mean the `dest_path` is a folder, at this point it could just be the parent
+            # of the actual dest_path, or have a blank name
+            if not dest_path.is_file:
+                temp_path = await self.revalidate_path(
+                    dest_path,
+                    rename or src_path.name,
+                    folder=src_path.is_dir
+                )
+                if self.will_self_overwrite(dest_provider, src_path, temp_path):
+                    raise exceptions.OverwriteSelfError(src_path)
+
             dest_path = await dest_provider.handle_naming(
                 src_path,
                 dest_path,
@@ -393,10 +427,15 @@ class BaseProvider(metaclass=abc.ABCMeta):
 
         :rtype: :class:`.WaterButlerPath`
         """
+
+        # This function can either take a file path, or a parent path to make a file path out of.
+
         if src_path.is_dir and dest_path.is_file:
             # Cant copy a directory to a file
             raise ValueError('Destination must be a directory if the source is')
 
+        # This is confusing. `dest_path` at this point can refer to the parent or root of
+        # the file we want to move/copy. So even if moving/copying a file, this code will run
         if not dest_path.is_file:
             # Directories always are going to be copied into
             # cp /folder1/ /folder2/ -> /folder1/folder2/
@@ -409,6 +448,22 @@ class BaseProvider(metaclass=abc.ABCMeta):
         dest_path, _ = await self.handle_name_conflict(dest_path, conflict=conflict)
 
         return dest_path
+
+    def will_self_overwrite(self,
+                            dest_provider: 'BaseProvider',
+                            src_path: wb_path.WaterButlerPath,
+                            dest_path: wb_path.WaterButlerPath) -> bool:
+        """ Return wether a move or copy operation will result in a self-overwrite.
+
+       .. note::
+            Defaults to False
+
+        :param dest_provider: ( :class:`.BaseProvider` ) The provider to check against
+        :param  src_path: ( :class:`.WaterButlerPath` ) The move/copy source path
+        :param  dest_path: ( :class:`.WaterButlerPath` ) The move/copy destination path
+        :rtype: :class:`bool`
+        """
+        return False
 
     def can_intra_copy(self,
                        other: 'BaseProvider',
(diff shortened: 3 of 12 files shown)


waterbutler/core/provider.py
```diff
@@ -476,6 +485,25 @@ def can_intra_move(self,
...
         return data, created
 
+    async def replace_will_orphan(self,
+                                  dest_path: wb_path.WaterButlerPath,
+                                  src_path: wb_path.WaterButlerPath) -> bool:
+        """Check if copy/move conflict=replace will orphan src_path.
+
+        This can be an expensive operation. Should be used as last resort. Should be overridden if
+        provider has a cheaper option.
+        """
```
AddisonSchiller: Since this can be expensive, is there a possible check you can do here that will quickly tell you if you need to go down the whole chain, ie if 
```
src_path.name != dest_path.name:
   return False
```
Unsure if this works or if this check is already done.

tests/providers/osfstorage/test_provider.py
```diff
@@ -381,6 +381,22 @@ def file_lineage():
 
 @pytest.mark.asyncio
 @pytest.mark.aiohttpretty
+async def test_intra_copy_folder_orphan(provider_and_mock, provider_and_mock2, children_response):
```
AddisonSchiller: I know I already asked but putting this here for thought anyway.
Any possibility of moving this to core/test_provider?  (feels more complete that way) 
If it is too large of a change, go ahead and ignore this and let Longze or Fitz decide.

waterbutler/core/exceptions.py
```diff
@@ -193,6 +193,12 @@ def __init__(self, path):
                          'folder onto itself is not supported.'.format(path))
 
 
+class OrphanSelfError(InvalidParameters):
+    def __init__(self, path):
+        super().__init__('Unable to move or copy \'{}\'. Moving or copying a folder onto parent '
```
AddisonSchiller: Nitpick: The error message is a bit hard  for a user to understand. Not a deal breaker though.

AddisonSchiller:
Forgot to add message to review.... 

Local testing went fine (just with osfstorage however)
1 change for sure (look at use of async/await)
some rewording and possible moving of test?
1 possible change on orphan functiong.
Otherwise looks good

tests/providers/osfstorage/test_provider.py
TomBaxter: Spoke with Fitz. For now going with this for brevity.

AddisonSchiller: 👍

waterbutler/core/exceptions.py
TomBaxter: Agreed it is a lot to parse, but not sure how to phrase it in a more clear way that still provides all needed information.

waterbutler/core/provider.py
```diff
@@ -476,6 +485,25 @@ def can_intra_move(self,
         await self.delete(src_path)
         return data, created
 
+    async def replace_will_orphan(self,
```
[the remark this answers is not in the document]
TomBaxter: Complete.

waterbutler/core/exceptions.py
AddisonSchiller: 👍

waterbutler/core/provider.py
```diff
@@ -476,6 +485,27 @@ def can_intra_move(self,
...
+                            src_path: wb_path.WaterButlerPath) -> bool:
+        """Check if copy/move conflict=replace will orphan src_path.
+
+        This can be an expensive operation. Should be used as last resort. Should be overridden if
+        provider has a cheaper option.
+        """
+        if not dest_path.kind == 'folder':
+            return False
+        if src_path.name != dest_path.name:
+            return False
```
AddisonSchiller: This line isn't hit by test coverage

waterbutler/core/provider.py
```diff
@@ -240,6 +240,13 @@ def request(self, *args, **kwargs):
         ):
             raise exceptions.OverwriteSelfError(src_path)
 
+        if (
+            self == dest_provider and
+            conflict == 'replace' and
+            dest_provider.replace_will_orphan(dest_path, src_path)
+        ):
+            raise exceptions.OrphanSelfError(src_path)
```
AddisonSchiller: this line isn't hit by test coverage

waterbutler/core/exceptions.py
```diff
@@ -208,6 +208,12 @@ def __init__(self, path):
                          'folder onto itself is not supported.'.format(path))
 
 
+class OrphanSelfError(InvalidParameters):
```
cslzchen: Can you add a doc str for this exception?

waterbutler/core/provider.py
```diff
@@ -476,6 +485,27 @@ def can_intra_move(self,
         await self.delete(src_path)
         return data, created
 
+    def replace_will_orphan(self,
+                            dest_path: wb_path.WaterButlerPath,
+                            src_path: wb_path.WaterButlerPath) -> bool:
+        """Check if copy/move conflict=replace will orphan src_path.
+
+        This can be an expensive operation. Should be used as last resort. Should be overridden if
```
cslzchen: Did a test, this is not as bad as it seems. Only when there is a nested structure like the following will trigger iteration on each level (for both valid and invalid move and copy).

nested_folders

tests/providers/osfstorage/test_provider.py
```diff
@@ -342,6 +342,24 @@ class TestIntraCopy:
         src_mock.validate_v1_path.assert_not_called()
 
 
+    @pytest.mark.asyncio
+    @pytest.mark.aiohttpretty
+    async def test_intra_copy_folder_orphan(self, provider_and_mock, provider_and_mock2,
```
cslzchen: As mentioned by @AddisonSchiller , this test only covers one situation when it is an invalid copy. Can you add tests to cover the following for both `copy` and `move`? So in total 4 * 2 = 8.

* invalid one
* return false if is not folder
* return false if name is not the same
* loop breaks if materialized path is not the same

waterbutler/core/exceptions.py
cslzchen: I tried to rephrase this but failed. 👍

tests/providers/osfstorage/test_provider.py
TomBaxter: Moved tests to core/provider and added move test for invalid.
Added single tests for 'not folder', 'not same name', and 'loop break' since they aren't running through copy/move.
Completed.

waterbutler/core/provider.py
TomBaxter: Completed.

waterbutler/core/provider.py
TomBaxter: Completed.

felliott:
There seems to be a lot of overlap between this and #270 so they may end up getting merged into one super-PR.  Leaving both open for now for reference until a final PR is merged.

waterbutler/core/provider.py
```diff
@@ -405,11 +410,46 @@ def request(self, *args, **kwargs):
                 rename or src_path.name,
                 folder=src_path.is_dir
             )
+            if conflict != 'keep':
+                if self.will_self_overwrite(src_provider, src_path, dest_path):
+                    raise exceptions.OverwriteSelfError(src_path)
+
+        # files and folders shouldn't overwrite themselves
+        # TODO: how is this different from above will_self_overwrite call?
```
birdbrained: Remove the todo? the former deals specifically with the case the destination is a dir, and the original will be kept?

waterbutler/core/provider.py
```diff
@@ -259,18 +255,21 @@ def request(self, *args, **kwargs):
...
+            'got_handle_conflicts': handle_conflicts,
             'conflict': conflict,
             'got_rename': rename is not None,
         })
-        if handle_naming:
-            dest_path = await dest_provider.handle_naming(
+
+        if handle_conflicts:
+
+            dest_path = await dest_provider.handle_conflicts(
```
birdbrained: Extra whitespace?

waterbutler/core/provider.py
```diff
@@ -202,29 +202,31 @@ def request(self, *args, **kwargs):
...
+            'got_handle_conflicts': handle_conflicts,
             'conflict': conflict,
             'got_rename': rename is not None,
         })
 
-        if handle_naming:
-            dest_path = await dest_provider.handle_naming(
+        if handle_conflicts:
+
+            dest_path = await dest_provider.handle_conflicts(
```
birdbrained: Extra whitespace

waterbutler/core/provider.py
```diff
@@ -476,6 +516,28 @@ def can_intra_move(self,
...
+        """Check if copy/move conflict=replace will orphan src_path.
+
+        Assumes src_provider == dest_provider and conflict == 'replace' have already been checked.
+        This can be an expensive operation. Should be used as last resort. Should be overridden if
+        provider has a cheaper option.
+        """
+        if not dest_path.kind == 'folder':
+            return False
+        if src_path.name != dest_path.name:
+            return False
```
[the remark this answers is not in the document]
birdbrained: That looks like its a private method because it's prefixed with an underscore. Should this be private, or can consuming code rely on self._is_folder?

birdbrained:
New PR https://github.com/CenterForOpenScience/waterbutler/pull/333

waterbutler/core/provider.py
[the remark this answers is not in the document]
cslzchen: Let's ignore this for now. It's everywhere in the code base, not only for WB but also for OSF and MFR ...

cslzchen:
This PR is replaced and closed.

waterbutler/core/provider.py
```diff
@@ -405,11 +410,46 @@ def request(self, *args, **kwargs):
...
+        # TODO: how is this different from above will_self_overwrite call?
+        if (
+            self.shares_storage_root(src_provider) and
+            src_path.materialized_path == dest_path.materialized_path
+        ):
+            raise exceptions.OverwriteSelfError(src_path)
+
+        if (
+            self == src_provider and
+            conflict == 'replace' and
```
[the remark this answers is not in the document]
birdbrained: actually no

PR #275, opened 2017-09-26
[SVCS-427] Don't zip zips when zipping.

Johnetordoff:
## Purpose 

Stop the download as zip feature from trying to zip already zipped files.

## Changes 

Add new condition which checks filename for 'zip' extension and disables compression for that file.

## Side Effects

None that I know of.

## Ticket

https://openscience.atlassian.net/browse/SVCS-427


tests/core/streams/test_zip.py
```diff
@@ -127,3 +127,46 @@ class TestZipStreamReader:
...
+        )
+
+        data = await stream.read()
+
+        for file in files:
+            file['handle'].close()
+
+        zip = zipfile.ZipFile(io.BytesIO(data))
+        # Verify CRCs
+        assert zip.testzip() is None
```
AddisonSchiller: Expand this comment a bit more, ie:
```
# testzip() returns `None` if there are no bad files in the zipfile
```

tests/core/streams/test_zip.py
```diff
@@ -127,3 +127,46 @@ class TestZipStreamReader:
 
         for file in files:
             assert zip.open(file['filename']).read() == file['contents']
+
+    @pytest.mark.asyncio
+    async def test_zip_files(self, temp_files):
+        files = []
+        for filename in ['file1.ext', 'zip.zip', 'file2.ext']:
+            path = temp_files.add_file(filename)
+            contents = os.urandom(2 ** 5)
```
AddisonSchiller: Not sure if this matters much, but is there a particular upside to using `urandom` over explicit contents?

AddisonSchiller:
One other small comment: Could you attach some test files to the Jira ticket that you used for this?

waterbutler/core/streams/zip.py
```diff
@@ -131,7 +131,10 @@ def __init__(self, file_tuple):
             date_time=time.localtime(time.time())[:6],
         )
         # If the file is a directory, set the directory flag, turn off compression
```
cslzchen: This comment is for the first `elif`. Move it down.

waterbutler/core/streams/zip.py
```diff
@@ -131,7 +131,10 @@ def __init__(self, file_tuple):
...
-        if self.zinfo.filename[-1] == '/':
+        if self.zinfo.filename.endswith('.zip'):
```
cslzchen: Similar to the compression off for directory, add a comment here for compressed content.

cslzchen: What about `external_attr`?

tests/core/streams/test_zip.py
cslzchen: Please update the comment here though it was `#Verify CRCs` in our existing code base. I had a hard time understanding it before I read @AddisonSchiller 's suggestion. No need to stick to what used to be if they are not as good, and feel free to improve comments, doc strs and code style for new code or existing related code. Gradually, we can make a difference to our code base 🎆 .

waterbutler/core/streams/zip.py
Johnetordoff: To the best of my knowledge external_attr doesn't recognize file-type, but file hierarchy information.

waterbutler/core/streams/zip.py
```diff
@@ -130,8 +130,12 @@ def __init__(self, file_tuple):
             filename=filename,
             date_time=time.localtime(time.time())[:6],
         )
+        # If the file is a zip don't zip it again.
+        if self.zinfo.filename.endswith('.zip'):
```
cslzchen: FYI, here is the permission after I unzip the downloaded file.

```bash
drwxr-xr-x@ 4 longzechen  staff   136 Nov 29 14:05 Test
-rwxr-xr-x@ 1 longzechen  staff  40371 Nov 29  2017 mfr-test-2.zip
-rw-------@ 1 longzechen  staff     8 Nov 29  2017 _mfr-test_111.txt
```

The `external_attr` does play a part in what the permissions the files in side the `.zip` will end up with after being download and extracted.

`rwxr-xr-x` is too much for a `.zip` file I assume. I will use `rw-------` instead.

felliott:
Terrific!  Merging!

PR #277, opened 2017-10-04
[SVCS-475] Support to log metadata GET requests

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-475

## Purpose
Initial support for logging GET requests

## Summary of changes
Removed line stopping metadata GET requests from being logged.
Added set up for logging metadata
Added `metadata` to the callback ignore list. This can be removed once OSF has the correct support for the new logging feature

removed a superfluous return

## Test/QA notes.

https://github.com/CenterForOpenScience/osf.io/blob/0f184000e7517f030a8b920818461279364109bd/addons/base/views.py#L299-L307

Here is where I'm pretty sure the OSF is defining what logs it wants. Since `metadata` does not exist here, if it isn't ignored it throws a lot of annoying error logs/messages in the waterbutler container.


Johnetordoff:
There should be tests for this, I know it there probably none to modify, but whipping up a test to ensure things are logged on_finish is critical. It's kind of a "Who watches the watchman?", situation with logs because use them to debug if there wrong we probably will find out in prod without unittests.

cslzchen:
@Johnetordoff ignore my comments, just talked with @felliott  and we will leave ticket assignment open for now.

cslzchen:
PR closed. Work continues in https://github.com/CenterForOpenScience/waterbutler/pull/307.

PR #278, opened 2017-10-04
[SVCS-499] Adding `size_as_int` property to `BaseFileMetadata`

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-499

## Purpose
Some provider metadata returns size as a string, not an int. Add a property that returns size as an int.

## Summary of Changes
Added `size_as_int` field to `BaseFileMetadata`. 
Added explicit metadata tests for the new property

## Test/QA notes

This change is very minimal and should require little QA. It should be most notable in the tests for now.


waterbutler/core/metadata.py
```diff
@@ -265,6 +265,14 @@ def size(self) -> int:
         """ Size of the file in bytes. """
         raise NotImplementedError
 
+    @property
+    def size_as_int(self) -> int:
+        """ Size of the file in bytes. """
```
Johnetordoff: There should be a comment here as well as in BaseMetadata.size that these to co-exsist to maintain backwards compatibility.

tests/providers/bitbucket/test_metadata.py
```diff
@@ -35,6 +35,8 @@ def test_build_file_metadata(self, file_metadata, owner, repo):
         assert metadata.created_utc is None
         assert metadata.content_type is None
         assert metadata.size == 13
+        assert metadata.size_as_int == 13
+        assert type(metadata.size_as_int) == int
```
Johnetordoff: Generally you want to avoid comparisons of this nature in Python because of how it handles inheritance. Typically we'd use `assert isinstance(metadata.size_as_int, int)`.

Johnetordoff:
LGTM! 👍

waterbutler/core/metadata.py
```diff
@@ -265,6 +265,17 @@ def size(self) -> int:
...
         raise NotImplementedError
 
+    @property
+    def size_as_int(self) -> int:
+        """ Size of the file as an int.
+        Some providers give metadata as an int. Both exist to maintain backwards
+        compatibility.
+        """
+        if self.size is not None:
+            return int(self.size)
```
cslzchen: Is it possible that `self.size` is an empty String? 
```bash
>>> int('')
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: ''
```
I prefer an easier solution by adding a `try ... except ...` around the method.
A less recommended one (more complicated + performance overhead) is to check the type first and handle `string` and `int` separately.

cslzchen: And we should add tests for this case as well.

Johnetordoff: Good catch, this is kind of a weird conditional. It should really be `if self.size:` because bool('') is False. I think with that change it won't need another unit test.

cslzchen: I had thought about `if self.size` but there is one more corner case. If the size is `0`, `bool(0)` is `False` instead of true ... I guess this is the reason why we use `if self.size is not None:` in the first place.

TomBaxter:
Merged this PR into [PR312](https://github.com/CenterForOpenScience/waterbutler/pull/312) and updated with minor logic change.

waterbutler/core/metadata.py
TomBaxter: Completed in https://github.com/CenterForOpenScience/waterbutler/pull/312

cslzchen:
PR Closed. As mentioned, work continues in https://github.com/CenterForOpenScience/waterbutler/pull/312.

PR #281, opened 2017-10-05
[SVCS-496] Fixing filesystem folder metadata name property

AddisonSchiller:
refs: https://openscience.atlassian.net/browse/SVCS-496

## Purpose
Filesystems folder metadata name was returning an empty string. This is a bug and should be fixed.

## Summary of Changes
changed the way the name property is parsed. Fixed test cases so the name value is now correct, (was just the empty string before)

## QA/Test Notes

Nothing to note

### DIFF
--- a/waterbutler/providers/filesystem/metadata.py
+++ b/waterbutler/providers/filesystem/metadata.py
@@ -25,7 +25,7 @@ class FileSystemFolderMetadata(BaseFileSystemMetadata, metadata.BaseFolderMetada
     @property
     def name(self):
         if self.raw['path'].endswith('/'):
-            return os.path.split(self.raw['path'][:-1])[1]
+            return os.path.split(self.raw['path'].rstrip('/'))[1]
         else:
             return os.path.split(self.raw['path'])[1]
 
--- a/waterbutler/providers/filesystem/metadata.py
+++ b/waterbutler/providers/filesystem/metadata.py
@@ -24,7 +24,10 @@ class FileSystemFolderMetadata(BaseFileSystemMetadata, metadata.BaseFolderMetada
 
     @property
     def name(self):
-        return os.path.split(self.raw['path'])[1]
+        if self.raw['path'].endswith('/'):
+            return os.path.split(self.raw['path'][:-1])[1]
+        else:
+            return os.path.split(self.raw['path'])[1]
 
     @property
     def path(self):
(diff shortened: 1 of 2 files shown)


waterbutler/providers/filesystem/metadata.py
```diff
@@ -24,7 +24,10 @@ class FileSystemFolderMetadata(BaseFileSystemMetadata, metadata.BaseFolderMetada
 
     @property
     def name(self):
-        return os.path.split(self.raw['path'])[1]
+        if self.raw['path'].endswith('/'):
```
Johnetordoff: If this is the class for folder metadata why do some paths not with `/` ? It's assumed all `FileSystemFileMetadata` don't end in `/` . Why not make the the opposite assumption here?

waterbutler/providers/filesystem/metadata.py
```diff
@@ -24,7 +24,10 @@ class FileSystemFolderMetadata(BaseFileSystemMetadata, metadata.BaseFolderMetada
...
+            return os.path.split(self.raw['path'][:-1])[1]
```
Johnetordoff: Might want to go `self.raw['path'].rstrip('\')` it's just more recognizable what you're doing.

waterbutler/providers/filesystem/metadata.py
AddisonSchiller: to maintain back compatibility. Tests in `filesystem/test_provider` would fail if this were not the case. I suppose I wasn't sure if those tests were intended to be written that way for some specific reason or not.

Johnetordoff: You change the test behavior then test manually, if things work out we'll keep the changes. If thing don't workout we'll find out why.

AddisonSchiller: I think it may have to do with the fact that its the filesystems provider acting differently than other providers. Considering the way the old tests were written and worked/made sense, but then the logic failed once we went back and added more. Might require a bit more investigation to figure out which is correct.
@felliott ?

Johnetordoff: I've pulled this and `return os.path.split(self.raw['path'].rstrip('/'))[1]` on line 27 passes unittests and seems to work fine locally.

AddisonSchiller: Makes sense after change from [:-1] to rstrip. Good catch!

cslzchen:
PR closed. Work continues in https://github.com/CenterForOpenScience/waterbutler/pull/308.

PR #283, opened 2017-10-12
Cloudfiles

Johnetordoff:
## Ticket
https://openscience.atlassian.net/browse/SVCS-486

## Purpose

Expand the cloudfiles provider to make more like a real provider, 

## Changes

Adds upload, download, fetch metadata, move, copy, folders creation etc to the Cloudfiles provider, also includes unit tests and moves fixtures to new file. Changes some methods to support unicode file names.

## Side effects

None that I know of.

## QA Notes

Should include testing for special characters in filenames.

## Deployment Notes

This should be deployed be for the corrosponding osf.io PR, https://github.com/CenterForOpenScience/osf.io/pull/7781


waterbutler/providers/cloudfiles/provider.py
```diff
@@ -138,61 +155,65 @@ def default_headers(self):
         return metadata, created
 
     @ensure_connection
-    async def delete(self, path, **kwargs):
+    async def delete(self, path, confirm_delete=False):
         """Deletes the key at the specified path
         :param str path: The path of the key to delete
-        :rtype ResponseStreamReader:
+        :param bool confirm_delete: not used in this provider
```
felliott: What happens if the user tries to `DELETE /resources/mst3k/providers/cloudfiles/`?  The user should receive a warning that deleting everything in the root folder is only allowed if `confirm_delete=1` is given.  It also should not attempt to delete the actual root folder.  I don't know how that would play out on S3, but on a dropbox-like provider, that would break the addon as the user would no-longer be connected to their storage root

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +387,89 @@ def _extract_endpoints(self, data):
...
+        query = {'prefix': '{:03x}'.format(len(path.name)) + path.name + '/'}
+        resp = await self.make_request(
+            'GET',
+            functools.partial(self.build_url, '', container=version_location, **query),
+            expects=(200, 204),
+            throws=exceptions.MetadataError,
+        )
+        json_resp = await resp.json()
+        current = (await self.metadata(path)).to_revision()
+        return [current] + [CloudFilesRevisonMetadata(revision_data) for revision_data in json_resp]
```
felliott: Does this response not include the metadata for the latest revision?  If so, that's worthy of a comment.

tests/providers/cloudfiles/fixtures.py
```diff
@@ -0,0 +1,243 @@
+import os
+import io
+import time
+import json
+import pytest
+import aiohttp
+import aiohttpretty
```
felliott: Nitpick: pytest, aiohttp* are third-party and should be separated from the core libs.  unittest is a core lib.

waterbutler/providers/cloudfiles/metadata.py
```diff
@@ -55,39 +55,51 @@ def __init__(self, raw, path):
...
     def etag(self):
-        return self.raw['etag']
+        return self.raw['ETAG']
 
     @property
     def extra(self):
         return {
             'hashes': {
-                'md5': self.raw['etag'].replace('"', ''),
+                'md5': self.raw['ETAG'],
```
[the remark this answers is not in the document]
Johnetordoff: Should quotes be added? They don't appear naturally as part of the metadata response.

waterbutler/providers/cloudfiles/provider.py
Johnetordoff: I added the comment in the docstring.

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -297,7 +330,7 @@ def _extract_endpoints(self, data):
         data = await resp.json()
         return data
 
-    async def _metadata_file(self, path, is_folder=False, **kwargs):
+    async def _metadata_item(self, path):
```
AddisonSchiller: Nitpick: Reason for the rename? Are cloudfiles files refered to as `items` or something like that?

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
                             break
```
AddisonSchiller: Line 390 has a line length issue.

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
...
+        """
+
+        resp = await self.make_request(
+            'PUT',
+            functools.partial(self.sign_url, path, 'PUT'),
+            expects=(200, 201),
+            throws=exceptions.CreateFolderError,
+            headers={'Content-Type': 'application/directory'}
+        )
+        await resp.release()
```
AddisonSchiller: What does Rackspace do if there is a conflict? We should return a specific error if possible. 

ie:
```
        async with self.request(
            'POST',
            self.build_url('folders'),
            data={
                'name': path.name,
                'parent': {
                    'id': path.parent.identifier
                }
            },
            expects=(201, 409),
            throws=exceptions.CreateFolderError,
        ) as resp:
            # Catch 409s to avoid race conditions
            if resp.status == 409:
                raise exceptions.FolderNamingConflict(path.name)
```

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
...
+        return await self._metadata_item(path)
```
AddisonSchiller: Shouldn't this just call `metadata()` and not the specific metadata helper function?
also, `_metadata_item/file()` was originally used for just files. This is creating folders. Is this okay?

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
...
+
+        try:
+            return resp.headers['X-VERSIONS-LOCATION']
+        except KeyError:
+            raise exceptions.MetadataError('The your container does not have a defined version'
+                                           ' location. To set a version location and store file '
+                                           'versions follow the instructions here: '
+                                           'https://developer.rackspace.com/docs/cloud-files/v1/'
+                                           'use-cases/additional-object-services-information/'
+                                           '#object-versioning')
```
AddisonSchiller: Might not be the best to add links in error messages. They get outdated, etc.
@felliott  ?

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
...
+
+    @ensure_connection
+    async def revisions(self, path):
+        """Get past versions of the request file from special user designated version container,
+        if the user hasn't designated a version_location container in raises an infomative error
```
AddisonSchiller: "in raises" -> "it raises"

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
...
+        message. The revision endpoint also doesn't return the current version so that is added to
+        the revision list after other revisions are returned. More info about versioning with Cloud
+        Files here:
+
+        https://developer.rackspace.com/docs/cloud-files/v1/use-cases/additional-object-services-information/#object-versioning
```
AddisonSchiller: line length. Although idk how to fix line length in a docstring

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
...
+            functools.partial(self.build_url, '', container=version_location, **query),
+            expects=(200, 204),
+            throws=exceptions.MetadataError,
+        )
+        json_resp = await resp.json()
+        current = (await self.metadata(path)).to_revision()
+        return [current] + [CloudFilesRevisonMetadata(revision_data) for revision_data in json_resp]
+
+    @ensure_connection
+    async def get_metadata_revision(self, path, version=None, revision=None):
```
AddisonSchiller: This function should follow the same naming scheme as the other metadata helper functions:
eg: `_metadata_<for some thing>`
so this should be called `_metadata_revision`

waterbutler/providers/cloudfiles/provider.py
AddisonSchiller: also revisions should probably be moved next to metadata, and not in the middle of the `_<helper_function>`'s

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -361,13 +394,110 @@ def _extract_endpoints(self, data):
...
-    def _serialize_folder_metadata(self, data):
+    def _serialize_metadata(self, data):
         if data.get('subdir'):
             return CloudFilesFolderMetadata(data)
         elif data['content_type'] == 'application/directory':
             return CloudFilesFolderMetadata({'subdir': data['name'] + '/'})
         return CloudFilesFileMetadata(data)
+
+    @ensure_connection
+    async def create_folder(self, path, **kwargs):
```
AddisonSchiller: create_folder should probably be moved so it is not in the middle of the _<helper_function>'s

waterbutler/providers/cloudfiles/provider.py
Johnetordoff: _metadata_item returns the metadata for that item, _metadata_folder returns the metadata for the content of the folder. So an item isn't necessarily a file.

waterbutler/providers/cloudfiles/provider.py
Johnetordoff: See my other comment.

waterbutler/providers/cloudfiles/provider.py
Johnetordoff: Folder naming conflict are handled in postvalidation. Here: https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/server/api/v1/provider/create.py#L65

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -138,61 +155,73 @@ def default_headers(self):
...
+
+        if path.is_root and not confirm_delete:
+            raise exceptions.DeleteError(
+                'query argument confirm_delete=1 is required for deleting the entire container.',
+                code=400
+            )
+
         if path.is_dir:
-            metadata = await self.metadata(path, recursive=True)
+            metadata = await self._metadata_folder(path, recursive=True)
```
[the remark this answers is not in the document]
Johnetordoff: No it's better to call the specific function you need. There's no reason to expose us to changes that are made in .metadata that aren't relevant here.

waterbutler/providers/cloudfiles/provider.py
AddisonSchiller: 👍

waterbutler/providers/cloudfiles/provider.py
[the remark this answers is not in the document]
AddisonSchiller: We aren't exposing it to anything. Calling the specific metadata function adds variation. The more similar the providers are, the better. Other providers in general follow the same pattern of just calling `metadata()` and not the specific function (but like i said, a few small exceptions exist)

waterbutler/providers/cloudfiles/provider.py
AddisonSchiller: see above comment

waterbutler/providers/cloudfiles/provider.py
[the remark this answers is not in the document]
Johnetordoff: No this is still better, you should never compose functions unnecessarily. All the providers are different. Anything not imposed on them intentionally by their parent class is fair game. Unless you cite a specific rule this stays as is.

waterbutler/providers/cloudfiles/provider.py
```diff
@@ -339,11 +381,12 @@ def _extract_endpoints(self, data):
...
         data = await resp.json()
 
-        # no data and the provider path is not root, we are left with either a file or a directory marker
+        # no data and the provider path is not root, we are left with either a file or a directory
+        # marker
         if not data and not path.is_root:
             # Convert the parent path into a directory marker (file) and check for an empty folder
-            dir_marker = path.parent.child(path.name, folder=False)
-            metadata = await self._metadata_file(dir_marker, is_folder=True, **kwargs)
+            dir_marker = path.parent.child(path.name, folder=path.is_dir)
```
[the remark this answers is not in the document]
Johnetordoff: The function child is to create a new child path with a new name, so you are your father's child, but your sibling is your father's child with a different name.

PR #290, opened 2017-10-24
[SVCS-528] Look for `mfr` query param from mfr in v1 provider

AddisonSchiller:
We can export .gdoc as a pdf if we know its coming from
mfr. Added `alt_export` functions and properties to Gdrive
utils and metadata. Gdrive download now looks for 'mfr' in its
kwargs and will request file as pdf if used properly.

## Ticket
https://openscience.atlassian.net/browse/SVCS-528
MFR counterpart:  https://github.com/CenterForOpenScience/modular-file-renderer/pull/292

## Purpose
Allow mfr to request .gdoc files in other export formats (to .pdf from .docx). However allow the osf to still download in .docx format

This will allow MFR to cut out the expensive unoconv conversion for .gdoc files

## Changes
v1 provider now looks for a 'mfr' query param. if true, the google drive provider will use `alt_export_name` and `alt_export_link()` instead of its normal counterparts. This will cause the file being sent to be a .pdf over a .docx
Added tests for new changes.

## Side effects
There should be none. File downloads from gdrive will not be effected as long as they don't include `...&mfr=true` as a query param

## QA Notes
This needs to be tested with its mfr counterpart to have noticeable changes.

### DIFF
--- a/tests/providers/googledrive/test_metadata.py
+++ b/tests/providers/googledrive/test_metadata.py
@@ -44,6 +44,7 @@ class TestMetadata:
         assert parsed.materialized_path == str(path)
         assert parsed.is_google_doc is False
         assert parsed.export_name == item['title']
+        assert parsed.alt_export_name == 'PART_1420130849837.pdf'
 
     def test_file_metadata_drive_slashes(self, basepath, root_provider_fixtures):
         item = root_provider_fixtures['file_forward_slash']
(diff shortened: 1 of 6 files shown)


Johnetordoff:
Looks like this has been failing Travis putting it in AD.

AddisonSchiller:
👍

cslzchen:
Replaced by https://github.com/CenterForOpenScience/waterbutler/pull/304.

PR #291, opened 2017-10-30
[SVCS-545][ChunkedUploads] Update Box provider to use chunked-uploads

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/SVCS-545

## Purpose

This allows Box to upload very large files using their chunked upload endpoint.

## Changes

Adds several functions to start an upload session, upload the individual chunks and commit the upload into one file. Includes unittests.

## Side effects

None that I know of.

## QA Notes

A file must be over 50MB to trigger a chunked upload and the progress bar is not a good indicator of how much of the download has completed.

## Deployment Notes

Should probably correspond with removing upload limits for dropbox in Fangorn.js in osf.io

### DIFF
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -347,7 +347,7 @@ class BoxProvider(provider.BaseProvider):
         parts = await self._upload_parts(data, session_data)
 
         # Step 4 complete the session and return the upload file's metadata.
-        return await self._complete_session(path, session_data, parts, data_sha)
+        return await self._complete_session(session_data, parts, data_sha)
 
     async def _create_upload_session(self,
                                      path: wb_path.WaterButlerPath,
@@ -372,11 +372,11 @@ class BoxProvider(provider.BaseProvider):
                             session_data: dict):
 
         parts = {'parts': []}
-        upload_size = len(bytearray(data))
+        upload_size = len(data)
 
         for start_bytes in range(0, upload_size, session_data['part_size']):
             chunk = data[start_bytes: start_bytes + session_data['part_size']]
-            chunk_size = len(bytearray(chunk))  # Final chunk size != session_data['part_size']
+            chunk_size = len(chunk)  # Final chunk size != session_data['part_size']
             chunk_sha = base64.standard_b64encode(hashlib.sha1(chunk).digest()).decode()
 
             byte_range = self._build_range_header((start_bytes, start_bytes + chunk_size - 1))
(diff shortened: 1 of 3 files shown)


waterbutler/providers/box/provider.py
```diff
@@ -24,6 +25,7 @@ class BoxProvider(provider.BaseProvider):
 
     NAME = 'box'
     BASE_URL = settings.BASE_URL
+    NONCHUNKED_UPLOAD_LIMIT = 50000000  # 50 MB
```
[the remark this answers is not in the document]
AddisonSchiller: In Cloudfiles, a similar variable is called `SEGEMENT_SIZE` While from the names they seem slightly different, is there some main reason for the distinction?

waterbutler/providers/box/provider.py
```diff
@@ -286,6 +288,14 @@ def default_headers(self) -> dict:
                      path: wb_path.WaterButlerPath,
                      conflict: str='replace',
                      **kwargs) -> typing.Tuple[BoxFileMetadata, bool]:
+
+        if stream.size > self.NONCHUNKED_UPLOAD_LIMIT:
```
AddisonSchiller: In Cloudfiles you have it like this:
```
if stream.size > self.SEGMENT_SIZE:
    return await self.chunked_upload(...)
```
any reason for making various calls in `upload` instead of following the pattern of cloudfiles?

AddisonSchiller:
One more thing I missed.. any particular reason for coverage drop? (unsure if this is due to cloudfiles ticket, or chunked upload one in general)
Sorry for the email spam!

waterbutler/providers/box/provider.py
```diff
@@ -286,40 +288,133 @@ def default_headers(self) -> dict:
...
+        """
+
+        # Step 1 get the sha of all the data you will upload and base 64 encode it.
+        data_sha = base64.standard_b64encode(hashlib.sha1(data).digest()).decode()
+
+        # Step 2 create an upload session with box and recieve session id.
+        session_data = await self._create_upload_session(path, data)
+
+        # Step 3 split the data into chunks and upload them to box.
+        parts = await self._upload_parts(data, session_data)
```
birdbrained: We should catch any errors here and send an abort request. https://developer.box.com/v2.0/reference#abort

birdbrained:
This PR has been replaced by https://github.com/CenterForOpenScience/waterbutler/pull/348

PR #294, opened 2017-11-01
[SVCS-547] [ChunkedUploads] Update S3 provider to use chunked-uploads

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/SVCS-547

## Purpose

This allows very large files to be uploaded via chunks using  the S3 addon.

## Changes

Adds unit tests and three major methods to the S3 provider which (1) creates an upload session (2) uploads parts to the session and (3) "commits" the parts and completes the session.

## Side effects

None that I know of.

## QA Notes

Uploading large files via Fangorn will require a small osf-side fix.

## Deployment Notes

No special deployment notes.


waterbutler/providers/s3/provider.py
```diff
@@ -190,36 +193,159 @@ def can_intra_move(self, dest_provider, path=None):
...
+
+            upload_url = functools.partial(
+                self.bucket.new_key(path.path).generate_url,
+                settings.TEMP_URL_SECS,
+                'PUT',
+                query_parameters=params,
+                headers=headers
+            )
+
+            resp = await self.make_request(
```
birdbrained: Here we make a request but we don't handle what happens if this fails. I think that if this fails it'll just throw a `exceptions.UploadError` and the whole upload returns a 500. (This would be a big pain for particularly large files).

The Error handler does not send s3 an abort request in the event of a faliure, This would cause chunks of failed uploads to remaing existing on s3 and accrue charges, even though wb and the OSF would have no knowledge of them.

waterbutler/providers/s3/provider.py
```diff
@@ -190,36 +193,159 @@ def can_intra_move(self, dest_provider, path=None):
...
+                params=params,
+                expects=(200, 201,),
+                throws=exceptions.UploadError,
+            )
+            await resp.release()
+            parts_metadata.append(resp.headers)
+
+        return parts_metadata
+
+    async def _complete_multipart_upload(self, path, session_data, parts_metadata):
```
birdbrained: Theres an endpoint for getting a list of the uploaded parts. Would it be beneficial to iterate over the uploaded parts to ensure they all exist prior to completing the upload?

waterbutler/providers/s3/provider.py
birdbrained: Also can this for loop be refactored so it runs requests in parallel, It looks like we wait for each part to upload before starting the upload of the next part.

waterbutler/providers/s3/provider.py
```diff
@@ -44,6 +44,9 @@ class S3Provider(provider.BaseProvider):
     """
     NAME = 's3'
 
+    NONCHUNKED_UPLOAD_LIMIT = 128000000  # 128 MB
+    CHUNK_SIZE = 64000000  # 64 MB
```
birdbrained: From AWS https://docs.aws.amazon.com/AmazonS3/latest/dev/mpuoverview.html
> When uploading a part, in addition to the upload ID, you must specify a part number. You can choose any part number between 1 and 10,000. A part number uniquely identifies a part and its position in the object you are uploading. Part number you choose need not be a consecutive sequence (for example, it can be 1, 5, and 14). If you upload a new part using the same part number as a previously uploaded part, the previously uploaded part is overwritten. 

This means limiting chunks by size is not enough, there also has to be less than 10,000 parts. This means files over 640 000 terabytes cannot be uploaded. 👾

birdbrained: How was the size 64MB picked out of curiosity?

felliott: @birdbrained: good point.  @Johnetordoff, if these limits are in the documentation somewhere, please provide a link in a comment. If they're guesses, a comment explaining the rationale would be good (even if it's just "reasonable first guess, will profile and tweak later")

waterbutler/providers/s3/provider.py
felliott: Yeah, if the upload aborts, we should try to do whatever cleanup we can.  But I don't think we can parallelize this, we're reading the data off a stream as it comes in.

birdbrained: It can be parallelized, though I missed line 278, I thought we were just reading off of disk or something.
It probably wouldn't do much in this case but it is a simple change and would make a difference if for example s3 has a lower upload bandwidth than both wb's upload and download bandwidth (if the file was being streamed from a location on our vps or something). https://quentin.pradet.me/blog/using-asynchronous-for-loops-in-python.html

waterbutler/providers/s3/provider.py
birdbrained: (that said I hope nobody is uploading files that big. It would be cheaper to drive a semi full of hard disks.)

Johnetordoff: 64MB was just picked because it seemed like a reasonable amount to upload in a single request. 128 is the absolute limit.

birdbrained: I think it's fine because this isn't user facing as long as we retry failed chunks.

PR #297, opened 2017-11-08
[SVCS-334]Adding custom sentry sanitizer

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-334

## Purpose
Sanitize and censor tokens we get back from Dataverse that show up in local variables

## Changes
Added a sanitizer based that inherits from the default raven one. It has increased functionality to parse through dictionaries, look for dataverse keys, and a larger scope of variable names to sanitize (key, token)

NOTE: `sanitize.py` and `test_sanitize.py`'s location in waterbutler are pretty variable. I put them in server just because that was the easiest place at the time. if they should live somewhere else, that is an easy change, and something that might need to be considered.

## Side effects
Some sentry things that we don't want to sanitize may end up being sanitized.

## QA Notes
To test, you need to trigger an error that will log a Dataverse API token in a local variable or request some how.
An easy way to do this is if locally testing, upgrade your furl to 1.0.1 in Waterbutler with Dataverse attached to your project. You will also need to add a personal sentry account to your Waterbutler (through your raven.config file, or by manually putting it in your settings).

One thing to note: If an API key shows up in the actual error message, this is currently not censored (not sure best way to go about this, but the only time I ever encountered this was with the error :`Bad API token`.. so the token is not valid anymore anyway.

To trigger the above "Bad API token" functionality, attach a dataverse account. Then on dataverse, go refresh your token and refresh the page. This error will trigger until the OSF has to revalidate (Or something like that. This only happened to me once or twice).

There are other manual ways to test locally, such as raising errors in odd places with variables named things like 'key' or having a variable value that looks like a dataverse token etc.

### DIFF
--- a/tests/server/test_sanitize.py
+++ b/tests/server/test_sanitize.py
@@ -1,5 +1,6 @@
-import pytest
 from unittest import mock
+
+import pytest
 
 from waterbutler.server.sanitize import WBSanitizer
 
@@ -71,12 +72,6 @@ class TestWBSanitizer:
             'okay_value': 'bears are awesome'
         }
         result = sanitizer.sanitize('sanitize_dict', sanitize_dict)
-
-        # Sanity check
-        assert result != {
-            'key': 'secret',
-            'okay_value': 'bears are awesome'
-        }
 
         assert result == {
             'key': self.MASK,
--- a/tests/server/test_sanitize.py
+++ b/tests/server/test_sanitize.py
@@ -26,9 +26,35 @@ class TestWBSanitizer:
     def test_value_is_none(self, sanitizer):
         assert sanitizer.sanitize('great hair', None) is None
 
+    def test_key_is_none(self, sanitizer):
+        assert sanitizer.sanitize(None, 'best day ever') is 'best day ever'
+
     def test_sanitize_credit_card(self, sanitizer):
         assert sanitizer.sanitize('credit', '424242424242424') == self.MASK
+        # This string is not censored since it is out of the range of what it considers
+        # to be a credit card
         assert sanitizer.sanitize('credit', '4242424242424243333333') != self.MASK
+
+    def test_none_key_is_sanitized(self, sanitizer):
+        assert sanitizer.sanitize(None, '424242424242424') == self.MASK
+        # This string is not censored since it is out of the range of what it considers
+        # to be a credit card
+        assert sanitizer.sanitize(None, '4242424242424243333333') != self.MASK
+
+    def test_dataverse_secret(self, sanitizer):
+
+        # Named oddly because if you call it `dv_secret` it will get sanitized by a different
+        # part of the sanitizer
+        dv_value = 'aaaaaaaa-bbbb-bbbb-bbbb-cccccccccccc'
+        assert sanitizer.sanitize('dv_value', dv_value) == self.MASK
+
+        dv_value = 'random characters and other things  aaaaaaaa-bbbb-bbbb-bbbb-cccccccccccc'
+        expected = 'random characters and other things  ' + self.MASK
+        assert sanitizer.sanitize('dv_value', dv_value) == expected
+
+    def test_bytes(self, sanitizer):
+        assert sanitizer.sanitize(b'key', 'bossy yogurt') == self.MASK
+        assert sanitizer.sanitize(b'should_be_safe', 'snow science') == 'snow science'
 
     def test_sanitize_dictionary(self, sanitizer):
         value_dict = {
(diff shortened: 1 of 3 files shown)


waterbutler/server/sanitize.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class WBSanitizer(SanitizePasswordsProcessor):
+    """Asterisk out things that look like passwords, keys, etc."""
+
+    # Store mask as a fixed length for security
+    MASK = '*' * 8
+
+    # Token and key added from original. Key is used by Dataverse
+    FIELDS = frozenset([
```
cslzchen: For phase 2 consideration: double check the list.

waterbutler/server/sanitize.py
```diff
@@ -0,0 +1,71 @@
...
+    def sanitize(self, key, value):
+        """Overload the sanitize function of the `SanitizePasswordsProcessor'."""
+        if value is None:
+            return
+
+        # Part of the original method. Looks for credit cards to sanitize
+        if isinstance(value, str) and self.VALUES_RE.search(value):
+            return self.MASK
+
+        if isinstance(value, dict):
```
cslzchen: Need recursion for `dict` + a test fot it.

waterbutler/server/sanitize.py
```diff
@@ -0,0 +1,71 @@
...
+                value = value.replace(match, self.MASK)
+
+        # key can be a NoneType
+        if not key:
+            return value
+
+        # Just in case we have bytes here, we want to turn them into text
+        # properly without failing so we can perform our check.
+        if isinstance(key, bytes):
+            key = key.decode('utf-8', 'replace')
```
cslzchen: Is it guaranteed to be correctly UTF-8 encoded?

AddisonSchiller: Unsure. This line is directly from raven, so I assume so.

waterbutler/server/sanitize.py
AddisonSchiller: complete, and for lists as well

waterbutler/server/sanitize.py
AddisonSchiller: Complete, for lists as well

waterbutler/server/sanitize.py
cslzchen: I checked the code base and we rarely uses `try` en/decoding. Let's leave it this way. In the future we may want to have a cleanup ticket that add `try...except` wrapper on this.

AddisonSchiller: ill add a comment

tests/server/test_sanitize.py
```diff
@@ -0,0 +1,196 @@
+import pytest
+from unittest import mock
```
Johnetordoff: unittest is a built-in library, import order.

tests/server/test_sanitize.py
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        sanitize_dict = {
+            'key': 'secret',
+            'okay_value': 'bears are awesome'
+        }
+        result = sanitizer.sanitize('sanitize_dict', sanitize_dict)
+
+        # Sanity check
+        assert result != {
```
Johnetordoff: Not necessary given the next the statement.

waterbutler/server/app.py
```diff
@@ -45,7 +45,8 @@ def make_app(debug):
         [(r'/status', handlers.StatusHandler)],
         debug=debug,
     )
-    app.sentry_client = AsyncSentryClient(settings.SENTRY_DSN, release=waterbutler.__version__)
+    app.sentry_client = AsyncSentryClient(settings.SENTRY_DSN, release=waterbutler.__version__,
```
cslzchen: Need to update this after bumping `setuptools` to `37.0.0`

cslzchen: @TomBaxter Thanks for catching this. I probably was thinking about MFR. No need to change.

waterbutler/server/sanitize.py
cslzchen: @TomBaxter My suggestion is check if there are other possible keys that may store sensitive or secret information. Here is the default keys from `SanitizePasswordProcessor`: https://github.com/htquach/raven-python/blob/master/raven/processors.py#L72.

```python
    FIELDS = frozenset([
        'password', 'secret', 'passwd', 'authorization', 'api_key', 'apikey',
        'pw', 'cred'
    ])
```

We don't have `'pw'`, `'cred'` and we added `'access_token'`, `'token'`, `'key'` and `'sentry_dsn'`. Other possible ones are `'refresh_token'`, `'code'`, etc. My suggestion is we need to understand what we need and what we don't.

TomBaxter: Understood. I plan on making better use of inheritance so that we don't miss any defaults in the parent. I'll also check the sentry data for Dataverse, for other possible things to sanitize. I'll definitely add  ```refresh token```.

cslzchen:
As mentioned, this PR is closed and continues in https://github.com/CenterForOpenScience/waterbutler/pull/313.

PR #298, opened 2017-11-08
[SVCS-590] Improve WB Move and Copy

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-590

## Purpose

"Actions"  are used in different places having different means. As requested by @icereval , this PR clean up the logic by using `provider_action` and `auth_action`. There is also an `osf_action` already introduced in https://github.com/CenterForOpenScience/waterbutler/pull/288.

### Primary

Replace `action` with `provider_action` and `auth_action`. The former are native actions supported by providers: `move` and `copy`. The latter are actions requested in the API request and used by the auth handler: `move`, `rename` and `copy`. Both `move` and `copy` in `auth_action` are mapped to `move` in `provider_action`.

### Minor

* Use `.get('key', 'default_value') instead of `['key']` for dictionary access.
* Use `http.HTTPStatus` instead of number literals

## Side effects

No

## QA Notes

Backend follow-up update after https://github.com/CenterForOpenScience/waterbutler/pull/288. Everything should word as expected. Testing a few different providers is recommended.

* move/copy/rename within a project
* move/copy/rename between two projects you own
* move/copy/rename between a project you own and a linked public project

## Deployment Notes

No

### DIFF
--- a/waterbutler/server/api/v1/provider/movecopy.py
+++ b/waterbutler/server/api/v1/provider/movecopy.py
@@ -53,19 +53,24 @@ class MoveCopyMixin:
 
     async def move_or_copy(self):
         """Copy, move and rename files and folders.
+
+        Auth actions: 'copy', 'move' and 'rename'
+        Provider actions: 'copy' and 'move'
+        Auth actions comes from the request and are used by the auth handler. They need to be converted into
+        provider actions which the providers use for actual move or copy.
         """
 
         # Force the json body to load into memory
         await self.request.body
 
-        json_action = self.json.get('action', 'null')
-        if json_action not in ('copy', 'move', 'rename'):
-            raise exceptions.InvalidParameters('Action must be copy, move or rename, not {}'.format(json_action))
+        auth_action = self.json.get('action', 'null')
+        if auth_action not in ('copy', 'move', 'rename'):
+            raise exceptions.InvalidParameters('Auth action must be copy, move or rename, not {}'.format(auth_action))
 
-        # Setup of the provider was delayed so the json action could be retrieved from the request body.
+        # Setup of the provider was delayed so the provider action can be updated from the auth action.
         provider = self.path_kwargs.get('provider', '')
-        provider_action = json_action
-        if json_action == 'rename':
+        provider_action = auth_action
+        if auth_action == 'rename':
             if not self.json.get('rename', ''):
                 raise exceptions.InvalidParameters('Rename is required for renaming')
             provider_action = 'move'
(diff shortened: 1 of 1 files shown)


waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -97,7 +105,7 @@ def build_args(self):
                 self.dest_resource,
                 self.json.get('provider', self.provider.NAME),
                 self.request,
-                action=action,
+                action=json_action,
```
icereval: `provider_action` ?

waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -51,43 +52,50 @@ def build_args(self):
...
+                raise exceptions.InvalidParameters('Rename is required for renaming')
+            provider_action = 'move'
 
         self.auth = await auth_handler.get(
-            self.resource, provider,
+            self.resource,
+            provider,
             self.request,
-            action=action,
+            action=json_action,
```
icereval: `provider_action` ?

waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -133,8 +141,8 @@ def build_args(self):
         self.dest_meta = metadata
 
         if created:
-            self.set_status(201)
+            self.set_status(HTTPStatus.CREATED)
```
icereval: :+1:

waterbutler/server/api/v1/provider/movecopy.py
cslzchen: This is part I changed from your code.

`auth_handler` has a `post_action_map` that needs to know `move` `copy` `rename`,so this should be the `json_handler`. See [here](
https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/auth/osf/handler.py#L100).

However, either should work since both actions (`move` and `rename`) are mapped to `upload`. Or you want to change the `post_action_map` in `auth_handler` to only map `move` and `copy` so we can use `provider_action`?

waterbutler/server/api/v1/provider/movecopy.py
cslzchen: Same as above.

waterbutler/server/api/v1/provider/movecopy.py
icereval: consider renaming to `auth_action`

felliott:
:heart:  Moving to RTM.

PR #299, opened 2017-11-10
[SVCS-576] Scrub keen payloads to make them valid

AddisonSchiller:
## Ticket
https://openscience.atlassian.net/browse/SVCS-576

## Purpose
Scrub '.' from keys in keen payloads. 

## Changes
'.' are now removed from keen payloads. If this created a conflict with another key, a (1) will be added to the end of it. Increments based on number of collisions
Added tests for new scrubbing function

## Side effects
Should be none

## QA Notes
To test this locally, you need to make a keen account at https://keen.io/ and then add your keys to
the settings file [here](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/settings.py#L165-L166) (the private keys)

Verify that your key is working by browsing a file or downloading a file on the osf. You should get some data to show up on your Keen.

Next, you need to send a bad payload. The best way to do this is to manually add code [here](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/core/remote_logging.py#L181) something like:
```
payload[test.....in....name] = 'dict value'
```

Once logged to keen, the value should appear as `testinname = 'dict value'`
Name incrementation can be tested like this:

```
payload['test....key] = 'some value'
payload[test.key] = 'other value'
```
once scrubbed, this will create a conflict. it will be solved by one of the keys turning into a new key

```
'testkey' : .... ,
'testkey (1)` value, ...
```

Unsure of best way to test this on staging, or how to send a dummy payload there.


waterbutler/core/remote_logging.py
```diff
@@ -302,6 +302,31 @@ def _build_public_file_payload(action, request, file_metadata):
...
 
+def _scrub_headers_for_keen(payload):
+    """ Scrub unwanted keystring characters like \\.\\ from a keen payload """
+
+    scrubbed_payload = {}
+    for key in payload:
+        scrubbed_key = key
+
+        if '.' in key:
+            scrubbed_key = key.replace('.', '')
```
felliott: Instead of removing the period entirely, can you replace it with a hyphen?  That'll make it easier to read.

waterbutler/core/remote_logging.py
```diff
@@ -302,6 +302,31 @@ def _build_public_file_payload(action, request, file_metadata):
     return public_payload
 
 
+def _scrub_headers_for_keen(payload):
+    """ Scrub unwanted keystring characters like \\.\\ from a keen payload """
+
+    scrubbed_payload = {}
+    for key in payload:
```
felliott: If you sort the keys in the payload, then `foo-bar` will come before `foo.bar`, meaning `foo.bar` is the one that will be renamed if both are present.

waterbutler/core/remote_logging.py
```diff
@@ -302,6 +302,31 @@ def _build_public_file_payload(action, request, file_metadata):
...
+        scrubbed_key = key
+
+        if '.' in key:
+            scrubbed_key = key.replace('.', '')
+
+        # if our new scrubbed key is already in the payload, we need to increment it
+        if scrubbed_key in scrubbed_payload:
+            i = 1
+            incremented_key = scrubbed_key + ' ({})'.format(i)
+            while incremented_key in scrubbed_payload:
```
felliott: As unlikely as it might be, I don't like having an unlimited while loop in the code.  Please add a `MAX_ITERATIONS` flag and just drop the value if you can find a free key after that many iterations.

tests/core/test_remote_logging.py
```diff
@@ -0,0 +1,66 @@
+import pytest
+
+from waterbutler.core import remote_logging
+
+
+class TestScubPayloadForKeen:
```
felliott: Spelling: "Scub" ==> "Scrub"

waterbutler/core/remote_logging.py
AddisonSchiller: complete

waterbutler/core/remote_logging.py
AddisonSchiller: complete

tests/core/test_remote_logging.py
AddisonSchiller: 👍

waterbutler/core/remote_logging.py
AddisonSchiller: 👍

felliott:
🍤 🍤 🍤   Moving to RTM!

PR #300, opened 2017-11-13
[SVCS-353]Look for Dataverse renamed files on upload

AddisonSchiller:
Dataverse 'ingests' certain file types. These file types get renamed.
In upload when Waterbutler tries to find the correct metadata to return,
it will 500 since it  was not looking for the renamed file.

## Ticket
https://openscience.atlassian.net/browse/SVCS-353

## Purpose
Uploading some file types to dataverse would cause a 500 error to be thrown in waterbutler. The file would still upload, but would have a new extension.

## Changes
Allow waterbutler to return the correct metadata based on the current file name, as well as a new property called `file.original_name`. This new property will parse DV raw metadata to figure out what the original file type was. This way WB can return the correct metadata and won't throw a 500.
added tests for new functions.

## Side effects
Currently DV just deletes old files if you try to upload a conflicting file name. (unsure if this is correct functionality or not)
Because DV changes the file names when uploading certain file types (csv, xlsx, and others), it wont be able to delete the original because it will look like it does not exist. Instead DV uses their own renaming scheme <conflicting-file-1>, <conflicting-file-2< ... 
Because of this, when a file that DV renames gets uploaded, if it got renamed with a -1, or a -2 etc, WB won't look for that. instead it will return the metadata of the non-renamed version.
This shouldn't be much of a problem, as it is really only used for updating fangorn. A page refresh negates this error

I had thought of a few ways to negate this, but decided against it as it would add quite a few more DV api calls to the process. If it is determined in CR that we want to do that, then it can be implemented

## QA Notes
The effect file types are as follows: .csv, .RData, .sav, .dta, .por, .xlsx
If you upload one of these files types to DV, it will be converted into a `.tab` file via their `ingestion` process.

Steps: 
* there are some sample file types on the JIRA ticket to use.
* do not batch upload all the files (DV will return a 400 since it will get stuck on one of the ingestion operations)
* One by one, upload each file
* it should complete successfully and show up in fangorn as <name.tab>.
* clicking on it should go to the right file.

Steps for 'side-effects' notes:
* upload a file for the second time
* it will look like 2 of the same file are on the page
* refresh the page and one of them should show up with a -1


AddisonSchiller:
Changes on last commit:

`original_name` -> `original_names` and is now a list
This gets rid of the need for "RData", "rdata" and "Rdata" etc. It now returns a list of possible names to look for metadata.

If you upload a large RData file, it takes awhile for dataverse to convert it. Any uploads during that time will give a 400 back saying ingestion is currently clogged. 
the provider will now look for these messages and display a message to the user that the upload process is clogged and to wait  a few seconds and try again.

To test the above upload an RData file. Once its load bar finishes, immediately upload another ingested file type. It should throw the error.

New tests for this new exception handling etc.

waterbutler/providers/dataverse/utils.py
```diff
@@ -0,0 +1,58 @@
...
+def original_ext_from_raw_metadata(data):
+    """Use the raw metadata to figure out possible original extensions."""
+    label = data.get('originalFormatLabel', None)
+    file_format = data.get('originalFileFormat', None)
+    content_type = data.get('contentType', None)
+
+    if not label or not file_format or not content_type:
+        return None
+
+    for key in ORIGINAL_FORMATS:
```
Johnetordoff: probably better as:

> for key, value in ORIGINAL_FORMATS.items():
> if (label == value['original_label'] and ...

waterbutler/providers/dataverse/provider.py
```diff
@@ -170,15 +171,26 @@ def can_duplicate_names(self):
...
+            else:
+                raise (await exceptions.exception_from_response(resp,
+                                                                error=exceptions.UploadError))
         await resp.release()
 
         # Find appropriate version of file
         metadata = await self._get_data('latest')
         files = metadata if isinstance(metadata, list) else []
-        file_metadata = next(file for file in files if file.name == path.name)
+        file_metadata = next(file for file in files if (file.name == path.name or
```
Johnetordoff: Can we not just use a list comprehension here?

waterbutler/providers/dataverse/utils.py
Johnetordoff: You can apply the above to TestUtils class as well.

PR #301, opened 2017-11-14
[SVCS-30] Allow delete folder that creates empty GitHub repo

TomBaxter:
## Ticket
[SVCS-30](https://openscience.atlassian.net/browse/SVCS-30)

## Purpose
Allow delete folder that creates empty GitHub repo

## Changes
During "def _folder_delete" check to see if folder is the last contents
of the repository. If so, execute a "def _delete_root_folder_contents".

## Side effects
Slight additional overhead when deleting a folder from repository root.
Deletion of last folder in repo will be much faster then other folder deletions.

## QA Notes
Test deleting folders from repository root. with/without contents and with/without other repo contents. This can be tested with postman and/or through osf.io .

### DIFF
--- a/tests/providers/github/test_provider.py
+++ b/tests/providers/github/test_provider.py
@@ -613,6 +613,10 @@ class TestCRUD:
         branch = 'master'
         message = 'deleted'
 
+        metadata_url = provider.build_repo_url('contents', ref='master')
+        aiohttpretty.register_json_uri('GET', metadata_url,
+                                       body=root_provider_fixtures['content_repo_metadata_root'])
+
         branch_url = provider.build_repo_url('branches', 'master')
         tree_url = furl.furl(
             provider.build_repo_url(
@@ -652,6 +656,10 @@ class TestCRUD:
         branch = 'master'
         message = 'deleted'
 
+        metadata_url = provider.build_repo_url('contents', ref='master')
+        aiohttpretty.register_json_uri('GET', metadata_url,
+                                       body=root_provider_fixtures['content_repo_metadata_root'])
+
         branch_url = provider.build_repo_url('branches', 'master')
         tree_url = furl.furl(provider.build_repo_url(
             'git', 'trees',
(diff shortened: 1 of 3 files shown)


waterbutler/providers/github/provider.py
```diff
@@ -365,6 +365,14 @@ def can_intra_copy(self, other, path=None):
         await resp.release()
 
     async def _delete_folder(self, path, message=None, **kwargs):
+        message = message or settings.DELETE_FOLDER_MESSAGE
+        if path.parent.is_root:
+            root_metadata = await self._metadata_folder(path.parent)
```
AddisonSchiller: huge nitpick: Even though we know our path if a folder, it still feels more kosher to call `metadata()` over the `_metadata_folder`.  see `_delete_file`.

TomBaxter: Completed.

TomBaxter:
Comments added, per CR note.

AddisonSchiller:
As per my old comment on `metadata()` vs `_metadata_folder()` or whichever was called, there may be a refactor changing this later, so the change I had you make was probably pointless. Noting for reference. and for @felliott

waterbutler/providers/github/provider.py
```diff
@@ -365,6 +365,16 @@ def can_intra_copy(self, other, path=None):
...
 
     async def _delete_folder(self, path, message=None, **kwargs):
+        message = message or settings.DELETE_FOLDER_MESSAGE
+        # _create_tree fails with empty tree (422 Invalid tree info), so catch it if this folder
+        # is the last contents of this repository and use _delete_root_folder_contents instead.
+        if path.parent.is_root:
+            root_metadata = await self.metadata(path.parent)
+            if len(root_metadata) == 1:
+                if root_metadata[0].path == path.materialized_path:
+                    await self._delete_root_folder_contents(path, message=message, **kwargs)
```
cslzchen: Both `_delete_root_folder_contents(path, ...)` and `_delete_root_folder_contents(path.parent, ...)` works locally for me. ~~Is there a difference between the two?~~ I feel `path.parent` is more intuitive since the method is designed to delete the contents of the `root` folder.

cslzchen: Deleting a folder is way more complicated. Github only provides API for deleting files. What we do here is modify the tree structure 👍 .

cslzchen: I delved deeper and finally understands how `_delete_root_folder_contents()` works. It just creates an empty tree? It uses `path` to find the `branch`, which explains why both `path` and `path.parent` works.

TomBaxter:
@cslzchen  After reading all CR comments, I don't see any change requests. Sending this back to CR.

cslzchen:
@TomBaxter I think I suggested using `_delete_root_folder_contents(path.parent, ...)` instead of `_delete_root_folder_contents(path, ...)`. The implicit assumption is the first argument should be the `root` instead of an arbitrary path. Does this sounds good to you?

cslzchen:
However, I may have misunderstood the function.

tests/providers/github/test_provider.py
```diff
@@ -613,6 +613,10 @@ class TestCRUD:
         branch = 'master'
         message = 'deleted'
 
+        metadata_url = provider.build_repo_url('contents', ref='master')
+        aiohttpretty.register_json_uri('GET', metadata_url,
```
cslzchen: Not sure why this is added and what it does.

cslzchen: Update: verified locally that tests fail without this change. Further investigation shows that we need to register `metadata_url` for fake request since the `_delete_folder ` now makes the "metadata" request.

Johnetordoff:
Code quality: 👍 
Unit tests: 👍 
Works locally: 👍 

It's would be nice if there was more intuitive way of doing this check in git, but there doesn't appear to be.

felliott:
Lovely!  Merged!

PR #303, opened 2017-12-01
[OSF-8573] Change One Drive versioning so it works with all account types

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/OSF-8573

## Purpose

Currently One Drive uses a "View Delta endpoint for versioning docs: [here](https://dev.onedrive.com/items/view_delta.htm), but that is only supported for One Drive for Business user users, see [here](https://docs.microsoft.com/en-us/onedrive/developer/rest-api/getting-started/release-notes). We want everyone to be able to access versioning, so we must change to a new endpoint, [here](https://docs.microsoft.com/en-us/onedrive/developer/rest-api/resources/driveitemversion) which all users can use. 

Unfortunately this endpoint doesn't support versioning so we put in a temporary work around which will just return the most recent version.

## Changes

- tests use new endpoint url
-  revisions use their own id instead of their eTag 
- exception handling for the above mentioned one note versions
- fixtures slightly reorganized

## Side effects

One note version behavior is changed.

## QA Notes

We'll have to run a thorough check of One Note files with registrations etc. They can be created by right clicking in One Drive and going under new. 

## Deployment Notes

None that I know of.

### DIFF
--- a/tests/providers/onedrive/fixtures/download.json
+++ b/tests/providers/onedrive/fixtures/download.json
@@ -1,11 +1,10 @@
 {
     "root_id": "F4D50E400DFE7D4E!103",
     "file_id": "F4D50E400DFE7D4E!291",
-    "file_revision": "aRjRENTBFNDAwREZFN0Q0RSEyOTEuMg",
     "onenote_id": "F4D50E400DFE7D4E!154",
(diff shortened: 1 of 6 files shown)


felliott:
Looks good, will merge to `develop`!  Thanks, @Johnetordoff!

PR #304, opened 2017-12-05
[SVCS-528] Improve gdoc rendering (WB Part)

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-528
MFR counterpart:  https://github.com/CenterForOpenScience/modular-file-renderer/pull/292

## Purpose

This PR replaces https://github.com/CenterForOpenScience/waterbutler/pull/290. Credit goes to @AddisonSchiller 🎆 🎆 .

Allow MFR to request `.gdoc` files in `.pdf` format while keep `.docx` format for downloading. This will allow MFR to cut out the expensive unoconv conversion for .gdoc files

We can export .gdoc as a pdf if we know its coming from mfr. Added `alt_export` functions and properties to Gdrive utils and metadata. Gdrive download now looks for 'mfr' in its kwargs and will request file as pdf if used properly.

## Changes

WB V1 provider now looks for a `mfr=` query param. If `mfr=true`, the google drive provider will use `alt_export_name` and `alt_export_link()` instead of its normal counterparts. This will cause the file being sent to be a `.pdf` instead of `.docx`.

## Side effects

None

## QA Notes

Test that `.gdoc` rendering works as usual.

## Deployment Notes

This PR must be merged/deployed at the same time as its MFR counterpart.


waterbutler/providers/googledrive/provider.py
```diff
@@ -234,9 +234,18 @@ def can_intra_copy(self,
 
         metadata = await self.metadata(path, revision=revision)
 
+        if kwargs.get('mfr', None) and kwargs['mfr'].lower() == 'true':
+            download_url = drive_utils.get_alt_export_link(metadata.raw)  # type: ignore
+            export_name = metadata.alt_export_name
+        else:
+
+            # TODO figure out metadata.raw.get('downloadUrl')
```
cslzchen: This assumes that `metadata.raw.get('downloadUrl')` is always `None`. Why?

waterbutler/providers/googledrive/provider.py
```diff
@@ -234,9 +234,18 @@ def can_intra_copy(self,
...
+            download_url = metadata.raw.get('downloadUrl') or drive_utils.get_export_link(metadata.raw)  # type: ignore
```
cslzchen: This assumes that `metadata.raw.get('downloadUrl')` is always `None`. Why?

Johnetordoff: My understanding of this line is that downloadUrl is always present unless it's a file that that isn't exportable in it's native type, so a .txt will have this, but a .gdoc won't. Google files have frustratingly nonstandard metadata.

waterbutler/providers/googledrive/provider.py
```diff
@@ -234,9 +234,18 @@ def can_intra_copy(self,
 
         metadata = await self.metadata(path, revision=revision)
 
+        if kwargs.get('mfr', None) and kwargs['mfr'].lower() == 'true':
+            download_url = drive_utils.get_alt_export_link(metadata.raw)  # type: ignore
```
cslzchen:
```python
download_url = metadata.raw.get('downloadUrl') or drive_utils.get_alt_export_link(metadata.raw) 
```

waterbutler/providers/googledrive/provider.py
```diff
@@ -234,9 +234,16 @@ def can_intra_copy(self,
 
         metadata = await self.metadata(path, revision=revision)
 
+        if kwargs.get('mfr', None) and kwargs['mfr'].lower() == 'true':
+            download_url = metadata.raw.get('downloadUrl') or drive_utils.get_alt_export_link(metadata.raw)
```
cslzchen: FYI, fixed. Need to check `metadata.raw.get('downloadUrl')` first. There are file types on gdrive (e.g. `.txt`) that do not have `export_links`.

birdbrained: what about 
```
if kwargs.get('mfr', '').lower() == 'true':
```
Also I don't get the `or` here.

waterbutler/providers/googledrive/utils.py
```diff
@@ -37,28 +40,42 @@ def is_docs_file(metadata):
...
+        if format_type.get('ext') == ext:
+            return format_type.get('mime_type')
 
 
 def get_format(metadata):
-    for format in DOCS_FORMATS:
-        if format['mime_type'] == metadata['mimeType']:
-            return format
+    for format_type in DOCS_FORMATS:
+        if format_type.get('mime_type') == metadata.get('mimeType'):
```
birdbrained: metadata.get('mimeType') should be outside the for loop so it doesn't need to be accessed each iteration. Given that this data structure is tiny so I doubt there will be any perf. difference.

waterbutler/providers/googledrive/utils.py
```diff
@@ -37,28 +40,42 @@ def is_docs_file(metadata):
...
+        return export_links.get(format_type.get('alt_type'))
+    else:
+        return export_links.get(format_type.get('type'))
 
 
 def get_export_link(metadata):
-    format = get_format(metadata)
-    return metadata['exportLinks'][format['type']]
+    format_type = get_format(metadata)
+    return metadata.get('exportLinks').get(format_type.get('type'))
```
birdbrained: What if metadata doesn't have an `exportLinks`?

birdbrained:
I think some of the stuff here may be better passed as a query param that waterbutler passes through to gdrive transparently, this pr currently puts code that deals with exporting into mfr, I'd like to look at if theres a way to put that logic in mfr

PR #305, opened 2017-12-11
[SVCS-397] Copy `root` without the `rename` parameter raises 400 instead of 500

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-397

## Purpose

This PR replaces https://github.com/CenterForOpenScience/waterbutler/pull/265 and credit goes to @AddisonSchiller 👍 👍 

Currently, if one tries to copy root, WB will throw `500`. Copying non-root files does not require rename. Code needs to be added so copying root without rename throws an appropriate error instead of `500`.

## Changes

Added a check so that if someone is attempting to copy `root`, the `rename` param needs to be set. Otherwise, throw an `InvalidParameters` error.

## Side effects

- [x] Verify/Update the code after https://github.com/CenterForOpenScience/waterbutler/pull/298 gets merged.

## QA Notes

Please refer to the QA Notes in the ticket.

## Deployment Notes

No

### DIFF
--- a/waterbutler/server/api/v1/provider/movecopy.py
+++ b/waterbutler/server/api/v1/provider/movecopy.py
@@ -107,13 +107,13 @@ class MoveCopyMixin:
             if path is None:
                 raise exceptions.InvalidParameters('"path" field is required for moves or copies')
             if not path.endswith('/'):
-                raise exceptions.InvalidParameters('"path" field requires a trailing slash to '
-                                                   'indicate it is a folder')
+                raise exceptions.InvalidParameters(
+                    '"path" field requires a trailing slash to indicate it is a folder'
+                )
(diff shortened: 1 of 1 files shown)


waterbutler/server/api/v1/provider/movecopy.py
```diff
@@ -112,6 +112,9 @@ def build_args(self):
 
             # TODO optimize for same provider and resource
 
+            if action == 'copy' and not self.json.get('rename') and self.path.is_root:
```
cslzchen: `action` is replaced with either `auth_action` or `provider_action`, please update and test again.

felliott:
:+1:  Merging!

PR #306, opened 2017-12-13
Add 'await' to stream.read in upload, remove md5 verification and add…

pdan93:
… md5 hash in _make_file_placeholder

Because we have independent tasks that compute the md5, it will almost never be computed fast enough for you to verify it right after the upload, that's why lines 568,569 should be deleted.
The md5 hash should still be supplied to us (it's not mandatory though) if you want the MD5 validation task to start. That's why I added that in lines 215, 232, 244.
Last but not least, the bigger problem was the stream read of the file. Without awaiting for it to finish you were sending incomplete data and aiohttp was maintaining the connection open (hence the 504 Gateway Time-outs). So I added a simple 'await' to the stream.read(size) part.

### DIFF
--- a/waterbutler/providers/figshare/provider.py
+++ b/waterbutler/providers/figshare/provider.py
@@ -212,7 +212,8 @@ class BaseFigshareProvider(provider.BaseProvider):
         # Process for creating a file:
 
         # 1. Get file ID
-        file_id = await self._make_file_placeholder(article_id, name, stream.size)
+        file_id = await self._make_file_placeholder(article_id, name, stream.size,
+                                                    stream.writers['md5'].hexdigest)
 
         # 2. Get upload url and file parts info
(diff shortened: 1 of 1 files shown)


felliott:
Thanks, @pdan93!  I'll take a look at this.  I didn't realize the hashing was done independently, so your approach makes sense.

Cheers,
@felliott

pdan93:
Thanks @felliott, and sorry I didn't see before doing the pull request you had tests that also need to be changed.

birdbrained:
@pdan93, Hey just wanted to check in - was curious if there was any work undertaken in regards to the tests mentioned above?

Also I'd be interested to hear a little more on what the problem was that you ran into that led you to this, just for our records.

felliott:
This is the issue with chunked uploads to figshare.  I've got a fix about ready to go that be merged today or tomorrow.

PR #307, opened 2017-12-18
[SVCS-475] Send hook request to OSF for GET requests

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/SVCS-475

## Purpose

The goal of this PR is to send a hook request for all file requests, including GETs so that the osf-side can better track provider metadata. I've also added some unit tests to improve coverage going forward.

## Changes

Modifies the accepted actions for sending hooks and creates groups of parameterized tests so the desired action behavior is more obvious. 

## Side effects

In order to compare LogPayload instances which aren't returned by the logging function, I've have to modify the __eq__ method of the the LogPayload to compare by serialization instead of the typical instance to instance comparison.

## QA Notes

None. This is not a user facing change.

## Deployment Notes

Requires osf-side changes to have any utility.


waterbutler/core/log_payload.py
```diff
@@ -57,6 +57,10 @@ def serialize(self):
 
         return payload
 
+    def __eq__(self, other: object) -> bool:
```
cslzchen: 👍  I can't think of a better solution than `__eq__`.

tests/core/test_remote_logging.py
```diff
@@ -74,3 +98,76 @@ def test_max_iteration(self):
...
+
+    @pytest.mark.asyncio
+    async def test_log_to_callback_move(self,
+                                        log_payload,
+                                        callback_log_payload_move,
+                                        mock_signed_request,
+                                        mock_time):
+
+        with mock.patch('waterbutler.core.utils.send_signed_request', mock_signed_request):
+            await log_to_callback('move', log_payload, log_payload)
```
cslzchen: In `remote_logging.py`, the signature for `log_to_callback()` is:
```python
async def log_to_callback(action, source=None, destination=None, start_time=None, errors=[]):
```
This doesn't look right. I removed the two `log_payload` arguments and tests passed as expected.

tests/server/api/v1/fixtures.py
```diff
@@ -0,0 +1,122 @@
+import time
```
cslzchen: Fix import order and style according to examples I added above.

tests/core/test_remote_logging.py
```diff
@@ -74,3 +98,76 @@ def test_max_iteration(self):
...
+                               "throwing the desired exception, it should take around 50-60 "
+                               "seconds")
+    @pytest.mark.asyncio
+    async def test_log_to_callback_throws_exception(self, log_payload, mock_signed_request):
+
+        with mock.patch('waterbutler.core.utils.send_signed_request', mock_signed_request):
+            with pytest.raises(Exception) as exc:
+                await log_to_callback('upload', log_payload, log_payload)
+                assert exc.message == 'Callback for upload request failed with {},' \
+                                      ' got {"status": "failure"}'.format(MockBadResponse())
```
cslzchen: Curly braces are escaped by using `{{` and `}}` in python.

tests/server/api/v1/fixtures.py
```diff
@@ -0,0 +1,122 @@
...
+    handler.dest_provider = MockProvider()
+    handler.dest_resource = 'test_dest_resource'
+    handler.dest_meta = MockFileMetadata()
+
+    return handler
+
+
+@pytest.fixture
+def source_payload(handler):
+    return LogPayload(handler.resource,
```
cslzchen: Fix style for arguments in function/method call.

cslzchen:
Update: secondary PR pending.

PR #309, opened 2017-12-18
[SVCS-421] Remove unsizable response reader flag

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/SVCS-421
A continuation of https://github.com/CenterForOpenScience/waterbutler/pull/273
## Purpose

The purpose of this PR is the remove the superfluous unsizable argument from the response reader safely and change the stream size behavior to make more sense.

## Changes

- Removes unsizable parameter for StreamResponseReader
- Changes conditional so size parameter directly controls stream size.
- Adds unit tests.

## Side effects

None that I know of.

## QA Notes

Download a file from each of the following providers:
- GoogleDrive
- Github
- Bitbucket
- Box
- Dropbox
- GitLab 

These all use size as a positional argument.

## Deployment Notes

None.


felliott:
:+1:, merging

PR #311, opened 2017-12-20
[SVCS-552] folder_file_op cut down revalidate calls.

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/SVCS-552

## Purpose

folder_file_op makes roughly a trillion metadata calls every time it moves a folder's children this fix makes it easy to get a path from it's metadata object by giving the metadata an id which corresponds with the path identifier. Since we can now convert an metadata object into a path easily we no longer need to re-validate each time.

## Changes

- Add id property to every provider's metadata.
- Changes from_metadata_to_path to uses metadata.id property
- Changes folder_file op to use from_metadata_to_path instead of revalidate_path
- Adds tests id property
- Adds tests for folder_file_op

## Side effects

None that I know of.

## QA Notes

Copy/moves between providers should be unchanged or slightly faster. 

## Deployment Notes

None.


waterbutler/providers/bitbucket/metadata.py
```diff
@@ -24,6 +24,10 @@ def __init__(self, raw, path_obj, owner=None, repo=None):
         self.owner = owner
         self.repo = repo
 
+    @property
+    def id(self):
+        return self.build_path()
```
cslzchen: Since `id` is exactly the same, why not using `return self.path`? This is what other providers do.

waterbutler/providers/googlecloud/metadata.py
```diff
@@ -27,8 +27,11 @@ class BaseGoogleCloudMetadata(metadata.BaseMetadata, metaclass=abc.ABCMeta):
     def provider(self) -> str:
         return 'googlecloud'
 
+    # GoogleCloud is a path-based provider.  However, it never calls ``revalidate_path()`` since it
+    # is the backend storage provider for OSFStorage.
```
cslzchen: Similar to CloudFiles (when it was the storage provider), GoogleCloud never uses ``revalidate_path()``. We should set the `id` to be `None` or empty so we know it is never used accidentally or cause any confusion in the further.

waterbutler/providers/github/metadata.py
```diff
@@ -19,14 +19,16 @@ def __init__(self, raw, commit=None, ref=None):
...
-    @property
-    def id(self):
-        return self.path
-
     @property
     def provider(self):
         return 'github'
 
+    # GitHub is a path-based provider. However, it has its own ``revalidate_path()``.
+    # TODO: Investigate if this is an issue.
```
cslzchen: GitHub may not work because your fix calls (functionally, not literally) the default ``revalidate_path()`` instead of the customized one.

waterbutler/providers/dropbox/metadata.py
```diff
@@ -9,14 +9,16 @@ def __init__(self, raw, folder):
...
-    @property
-    def id(self):
-        return self.raw['id']
-
     @property
     def provider(self):
         return 'dropbox'
 
+    # TODO: Is Dropbox path-based instead of id-based?
+    # TODO: Dropbox does not have a ``revalidate_path()``, should we use path instead of id?
```
cslzchen: There is no `revalidate_path()` for dropbox which makes guess that it is a path based (or at lease both)?

waterbutler/providers/dataverse/metadata.py
```diff
@@ -3,14 +3,18 @@
...
-    def id(self):
-        return str(self.raw['id'])
-
     @property
     def provider(self):
         return 'dataverse'
 
+    # TODO: Should this be in ``DataverseFileMetadata``?
+    # TODO: Does ``self.raw['id']`` exists for ``DataverseDatasetMetadata``?
+    # TODO: DataVerse's ``revalidate_path()`` looks very different. Is this an issue?
```
cslzchen: Similar to GitHub, your fix calls (functionally, not literally) ``revalidate_path()`` differently than the original way. This might be an issue.

waterbutler/providers/cloudfiles/metadata.py
```diff
@@ -5,14 +5,18 @@
...
-    def id(self):
-        return self.path
-
     @property
     def provider(self):
         return 'cloudfiles'
 
+    # CloudFiles is a path-based provider.  However, it never called ``revalidate_path()`` even when
+    # it was the backend storage provider for OSFStorage.  The provider is no longer active until it
+    # is upgraded to an addon provider for the users.
```
cslzchen: See GoogleCloud.

PR #313, opened 2017-12-27
[SVCS-334] Sentry captures dataverse API tokens on certain error types

TomBaxter:
This PR includes merge of https://github.com/CenterForOpenScience/waterbutler/pull/297 plus refactor of sub-classed raven processor.

## TODO
- [x] Move self.FIELDS.union to WBSanitizer __init__

## Ticket
[SVCS-334](https://openscience.atlassian.net/browse/SVCS-334)

## Purpose
Sanitize and censor tokens we get back from Dataverse that show up in local variables

## Changes
Added a sanitizer based that inherits from the default raven one. It has increased functionality to parse through dictionaries, look for dataverse keys, and a larger scope of variable names to sanitize (key, token)

NOTE: sanitize.py and test_sanitize.py's location in waterbutler are pretty variable. I put them in server just because that was the easiest place at the time. if they should live somewhere else, that is an easy change, and something that might need to be considered.

## Side effects
Some sentry things that we don't want to sanitize may end up being sanitized.

## QA Notes
To test, you need to trigger an error that will log a Dataverse API token in a local variable or request some how.
An easy way to do this is if locally testing, upgrade your furl to 1.0.1 in Waterbutler with Dataverse attached to your project. You will also need to add a personal sentry account to your Waterbutler (through your raven.config file, or by manually putting it in your settings).

One thing to note: If an API key shows up in the actual error message, this is currently not censored (not sure best way to go about this, but the only time I ever encountered this was with the error :Bad API token.. so the token is not valid anymore anyway.

To trigger the above "Bad API token" functionality, attach a dataverse account. Then on dataverse, go refresh your token and refresh the page. This error will trigger until the OSF has to revalidate (Or something like that. This only happened to me once or twice).

There are other manual ways to test locally, such as raising errors in odd places with variables named things like 'key' or having a variable value that looks like a dataverse token etc.

### DIFF
--- a/waterbutler/server/app.py
+++ b/waterbutler/server/app.py
@@ -44,7 +44,8 @@ def make_app(debug):
         [(r'/status', handlers.StatusHandler)],
         debug=debug,
     )
-    app.sentry_client = AsyncSentryClient(settings.SENTRY_DSN, release=__version__)
+    app.sentry_client = AsyncSentryClient(settings.SENTRY_DSN, release=__version__,
+                                        processors=('waterbutler.server.sanitize.WBSanitizer',))
     return app
 
 
--- a/waterbutler/server/sanitize.py
+++ b/waterbutler/server/sanitize.py
@@ -1,0 +1,71 @@
+import re
+
+from raven.processors import SanitizePasswordsProcessor
+
+
+class WBSanitizer(SanitizePasswordsProcessor):
+    """Asterisk out things that look like passwords, keys, etc."""
+
+    # Store mask as a fixed length for security
+    MASK = '*' * 8
+
+    # Token and key added from original. Key is used by Dataverse
+    FIELDS = frozenset([
+        'password',
+        'secret',
+        'passwd',
+        'authorization',
+        'api_key',
+        'apikey',
+        'sentry_dsn',
+        'access_token',
+        'key',
+        'token',
+    ])
+
+    # Credit card regex left intact from original processor
+    # While we should never have credit card information, its still best to perform the check
+    # and keep old functionality
+    VALUES_RE = re.compile(r'^(?:\d[ -]*?){13,16}$')
+
+    # Should specifically match Dataverse secrets. Key format checked on demo and on Harvard
+    DATAVERSE_SECRET_RE = re.compile(r'[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]'
+                                                '{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}')
+
+    def sanitize(self, key, value):
+        """Overload the sanitize function of the `SanitizePasswordsProcessor'."""
+        if value is None:
+            return
+
+        # Part of the original method. Looks for credit cards to sanitize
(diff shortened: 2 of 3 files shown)


waterbutler/server/sanitize.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Should specifically match Dataverse secrets. Key format checked on demo and on Harvard
+    DATAVERSE_SECRET_RE = re.compile(r'[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]'
+                                                '{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}')
+
+    def sanitize(self, key, value):
+        """Subclass the sanitize function of the `SanitizePasswordsProcessor'."""
+
+        # As of raven version 6.4 this attribute name has been changed from FIELDS to KEYS.
+        # Will need to be updated when we upgrade.
```
cslzchen: 👍

waterbutler/server/sanitize.py
```diff
@@ -0,0 +1,43 @@
...
+        self.FIELDS = self.FIELDS.union(['key', 'token', 'refresh_token'])
```
cslzchen: +1 for using `.union` on `frozenset`. I had thought this doesn't work but then I learned that `.union` on `frozenset` updates the set and returns a `frozenset`.

Is there a specific reason why we don't do this in `__init__()`? The side-effect of doing it in`sanitize()` is that the `.union` is called every time the function is called. (Note that we have recursion here).

waterbutler/server/app.py
```diff
@@ -44,7 +44,13 @@ def make_app(debug):
         [(r'/status', handlers.StatusHandler)],
         debug=debug,
     )
-    app.sentry_client = AsyncSentryClient(settings.SENTRY_DSN, release=__version__)
+
+    app.sentry_client = AsyncSentryClient(
```
cslzchen: FYI, I fixed merge conflicts when updating the PR directly on GitHub.

waterbutler/server/sanitize.py
TomBaxter: Good catch, no, no good reason. I will move when ticket next goes to ```additional development```

cslzchen: Sounds good. Back to `Add'l Dev` and you.

TomBaxter: Completed.

waterbutler/server/app.py
```diff
@@ -44,7 +44,8 @@ def make_app(debug):
         [(r'/status', handlers.StatusHandler)],
         debug=debug,
     )
-    app.sentry_client = AsyncSentryClient(settings.SENTRY_DSN, release=__version__)
+    app.sentry_client = AsyncSentryClient(settings.SENTRY_DSN, release=__version__,
```
cslzchen: @felliott If you prefer, you can update the style during merge. Thanks.

birdbrained:
https://github.com/CenterForOpenScience/waterbutler/pull/331

PR #276, opened 2017-10-02
[WIP] Update googledrive provider to use googledrive v3 API

TomBaxter:
WIP
[#SVCS-426 ]

PR #314, opened 2018-01-03
[SVCS-595] Clean up existing type annotation failures

TomBaxter:
## Ticket
[SVCS-595](https://openscience.atlassian.net/browse/SVCS-595)

## Purpose
Update box and google providers to pass mypy

## Changes
Fix type hints and/or signatures to pass mypy
Ignore two lines I was unable to resolve

## Side effects

None

## QA Notes

Only GoogleDrive and Box providers touched. 

- Box:
-- Intra copy/move
-- Folder metadata
- GoogleDrive:
-- Revisions

### DIFF
--- a/waterbutler/providers/box/provider.py
+++ b/waterbutler/providers/box/provider.py
@@ -194,7 +194,7 @@ class BoxProvider(provider.BaseProvider):
                          dest_provider: provider.BaseProvider,
                          src_path: wb_path.WaterButlerPath,
                          dest_path: wb_path.WaterButlerPath) \
-            -> typing.Tuple[typing.Union[BoxFileMetadata, BoxFolderMetadata], bool]:
+            -> typing.Tuple[BaseBoxMetadata, bool]:
         if dest_path.identifier is not None:
             await dest_provider.delete(dest_path)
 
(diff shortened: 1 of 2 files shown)


Johnetordoff:
LGTM! Worth noting however that there are a few `# type: ignore` statements [like here](https://github.com/CenterForOpenScience/waterbutler/pull/314/files#diff-dc1d5e76ff191c30934d4e6726a96e14R183) on google drive that seems to serve no purpose, but that's not really in scope.

cslzchen:
@Johnetordoff Thanks for the review. These are removed in https://github.com/CenterForOpenScience/waterbutler/pull/276/files.

waterbutler/providers/box/provider.py
```diff
@@ -501,7 +500,8 @@ def _build_upload_url(self, *segments, **query):
             box_path = await self.validate_path(child.path)
             await self.delete(box_path)
 
-    async def _intra_move_copy_metadata(self, path, data: dict) -> BaseBoxMetadata:
+    async def _intra_move_copy_metadata(self, path, data: dict) -> typing.Tuple[BaseBoxMetadata,
```
cslzchen: Is there a reason that `path` does not have a type specified?  Is it a dropbox path (if there is one), waterbutler path or a string?

cslzchen: Update: `path` is of type `WaterButlerPath`.

PR #318, opened 2018-02-15
[SVCS-319] Fix Range-header handling (enables MFR video support on Safari)

felliott:
## Ticket

[SVCS-319](https://openscience.atlassian.net/browse/SVCS-319)

This PR replaces #234 

## Purpose

Fix WB's broken Range header support, incidentally fixing video rendering on Safari.

WB was using an external library for parsing HTTP Range headers.  WB assumed that the tuple returned represented the first and last byte positions requested.  It was actually a list-slice compatible tuple, meaning that the end index was one byte greater than expected.  Given `Range: bytes=0-1`, the library would return `(0, 2)`, which WB would treat as `Range: bytes=0-2`.  This patch series fixes this error, adds code to ensure partial responses are properly closed, and adds Range header support to several providers that did not have it already.

This issue was unearthed by Detective @Johnetordoff in PR #234.  His fix would have worked, but was working around WB's incorrect behavior.  Instead, let's teach WB to DTRT.  This exposed other issues in WB's Range handling, which have been addressed.

## Changes

* Add a new utility method, `waterbutler.server.utils.parse_request_range`, that wraps the tornado version but returns an inclusive range instead of a list-slicing range.
* Manually close partial responses after download.  Partial responses did not exhaust their stream (they read exactly the number of bytes they contain), so the provider-originating response was not being released in `ResponseStreamReader`.  Unreleased aiohttp responses warn loudly on destruction.
* Document the three providers that don't currently support Range headers: Bitbucket, Dataverse, and GitHub.  Pass the Range header anyway, in hopes that they'll turn it on one day, and it will Just Work™.
* Manually implement Range header support for GitLab. We have to buffer the entire response into memory anyway, so we may as well trim it while we got it.
* Start sending the Range header to Figshare.  Figshare supports it, we just weren't passing it on.
* Implement Range header support for the filesystem provider by creating a new version of FileStreamReader that can return only the requested byte range from a file.
* Add tests for all supporting providers to ensure the Range Header is getting set and the returned response is a partial response.  The second part is a bit artificial for all but the GitLab and filesystem providers, since we mock the response.

## Side effects

* Positive: most (all?) providers now have a logger object available in them!  Logging defaults to `debug` level to not clog the terminal. 
* WB now rejects Range headers of the format `bytes=-6` with a 400 error.  Some providers do support this, so we may want to reconsider in the future.

## QA Notes

1. Upload videos (mp4 has good cross-browser support) to all WB-supported providers.
2. Attempt to view each on Safari.  Only Bitbucket, Dataverse, and GitHub should fail to render.
3. Attempt to download the file from each provider.  The file should download successfully for all providers and be playable locally (i.e. not corrupted)
4. *Bonus round*: Upload a plain text file containing the English alphabet to each provider. For each provider, send a WB download request with the header `Range: bytes=5-9`.  You should receive the string `fghij` from all but Bitbucket, Dataverse, and Github.  From those three, you should receive the entire original file.

## Deployment Notes

None needed.

### DIFF
--- a/waterbutler/core/streams/file.py
+++ b/waterbutler/core/streams/file.py
@@ -46,3 +46,55 @@ class FileStreamReader(BaseStream):
         await asyncio.sleep(0.001)
         async for chunk in self.file_gen:
             return chunk
+
+
+class PartialFileStreamReader(FileStreamReader):
+    """Awful class, used to avoid messing with FileStreamReader"""
+
+    def __init__(self, file_pointer, byte_range):
+        super().__init__(file_pointer)
+        self.start = byte_range[0]
+        self.end = byte_range[1]
+        self.bytes_read = 0
+
+    @property
+    def size(self):
+        return self.end - self.start + 1
+
+    @property
+    def total_size(self):
+        cursor = self.file_pointer.tell()
+        self.file_pointer.seek(0, os.SEEK_END)
+        ret = self.file_pointer.tell()
+        self.file_pointer.seek(cursor)
+        return ret
+
+    @property
+    def partial(self):
+        return self.size < self.total_size
+
+    @property
+    def content_range(self):
+        return 'bytes {}-{}/{}'.format(self.start, self.end, self.total_size)
+
+    @agent.async_generator
+    def chunk_reader(self):
+        self.file_pointer.seek(self.start)
+        while True:
+            chunk = self.file_pointer.read(self.read_size)
+            self.bytes_read += self.read_size
+            if not chunk:
+                self.feed_eof()
+                yield b''
+
+            yield chunk
+
+    async def _read(self, size):
+        self.file_gen = self.file_gen or self.chunk_reader()
+        bytes_remaining = self.size - self.bytes_read
+        self.read_size = bytes_remaining if size == -1 else min(size, bytes_remaining)
+        # add sleep of 0 so read will yield and continue in next io loop iteration
+        # asyncio.sleep(0) yields None by default, which displeases tornado
+        await asyncio.sleep(0.001)
+        async for chunk in self.file_gen:
+            return chunk
--- a/waterbutler/providers/gitlab/provider.py
+++ b/waterbutler/providers/gitlab/provider.py
@@ -197,7 +197,7 @@ class GitLabProvider(provider.BaseProvider):
         return [GitLabRevision(item) for item in data]
 
     async def download(self,  # type: ignore
-                       path: GitLabPath, **kwargs):
+                       path: GitLabPath, range: typing.Tuple[int, int]=None, **kwargs):
         """Return a stream to the specified file on GitLab.
 
         There is an endpoint for downloading the raw file directly, but we cannot use it because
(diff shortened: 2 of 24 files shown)


waterbutler/server/api/v1/provider/metadata.py
```diff
@@ -101,6 +107,11 @@ class MetadataMixin:
 
         await self.write_stream(stream)
 
+        if getattr(stream, 'partial', False) and isinstance(stream, ResponseStreamReader):
+            await stream.response.release()
```
cslzchen: I thought that I understood this during discussion but my mind lied. What will happen if we don't release the partial response here? I assume that it must have already been handled somewhere, otherwise the old code should have broken.

cslzchen: Optional: if we must release the partial response, should we do it for V0 as well? I am OK if we don't since V0 is going to be gone soon anyway.

waterbutler/providers/gitlab/provider.py
```diff
@@ -235,9 +238,17 @@ def path_from_metadata(self,  # type: ignore
             # fixed in GL v9.5
             data = self._convert_ruby_hash_to_dict(raw_data)
 
+        mdict_options = {}
+
         raw = base64.b64decode(data['content'])
+        if range is not None:
```
cslzchen: No action needed: as discussed, this manual effort has a side effect that each range request actually makes a full request (and receives the full data). However, this is the only solution to support the header due to the encoding.

waterbutler/server/utils.py
```diff
@@ -28,6 +29,37 @@ def make_disposition(filename):
...
+    if request_range is None:
+        return request_range
+
+    start, end = request_range
+    if start is None and end is None:
+        return None
+    elif start is None or start < 0:
+        raise InvalidHeaderError('Unsupported Range header format: {}'.format(range_header))
+
+    if end is not None:
```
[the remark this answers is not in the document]
cslzchen: Update: please add a test if you decide to accept this change

waterbutler/core/streams/file.py
```diff
@@ -46,3 +46,55 @@ def chunk_reader(self):
         await asyncio.sleep(0.001)
         async for chunk in self.file_gen:
             return chunk
+
+
+class PartialFileStreamReader(FileStreamReader):
```
cslzchen: As discussed, we agreed that this is the way to go. 👍

waterbutler/core/streams/file.py
```diff
@@ -46,3 +46,55 @@ def chunk_reader(self):
...
+    """Awful class, used to avoid messing with FileStreamReader"""
```
cslzchen: Opposite to our discussion, let's keep the docstr. I kind of like the use of "Awful" here.

waterbutler/server/api/v1/provider/metadata.py
felliott: Yeah, I don't think it's worth it for v0, but I'll make a note in the commit message explaining that.  Not releasing the partial response isn't the end of the world, but can cause a lot of error spam that can hide real issues.

PR #324, opened 2018-03-05
[PLAT-218] Change log callback so we can log file views

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/PLAT-218
Plat PR:
https://github.com/CenterForOpenScience/osf.io/pull/8184

## Purpose

Recently we opened up the log_to_callback function to send back data on GET actions, this completes that behavior by sending the file's version info and not sending it when listing revisions.

## Changes

- Minor logic changes to handler.

## Side effects

None that I know of.

## QA Notes

Should be tested as part of PLAT-218 ticket follow notes on that PR.

## Deployment Notes

None that I know of.

### DIFF
--- a/waterbutler/server/api/v1/provider/__init__.py
+++ b/waterbutler/server/api/v1/provider/__init__.py
@@ -150,7 +150,7 @@ class ProviderHandler(core.BaseHandler, CreateMixin, MetadataMixin, MoveCopyMixi
         if any((method in ('HEAD', 'OPTIONS'), status == 202, status > 302, status < 200)):
             return
 
-        if method == 'GET' and 'meta' in self.request.query_arguments:
+        if method == 'GET' and ('meta' in self.request.query_arguments or 'revisions' in self.request.query_arguments):
             return
 
         # Done here just because method is defined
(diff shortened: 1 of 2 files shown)


waterbutler/server/api/v1/provider/__init__.py
```diff
@@ -150,7 +150,7 @@ def on_finish(self):
         if any((method in ('HEAD', 'OPTIONS'), status == 202, status > 302, status < 200)):
             return
 
-        if method == 'GET' and 'meta' in self.request.query_arguments:
+        if method == 'GET' and ('meta' in self.request.query_arguments or 'revisions' in self.request.query_arguments):
             return
 
         # Done here just because method is defined
```
birdbrained: Can we use the fact we're changing code in their immediate vicinity as an excuse to make lines 156-162 a little more pythonic?

felliott:
Merged, but the request metadata structure has been changed slightly.  The metadata is now under the `request_meta` key and includes four fields: `url`, `referrer`, `method`, and `user_agent`.  The full list of headers is no longer being returned.

PR #326, opened 2018-03-09
[SVCS-673] Use Context Manager to Release S3 Responses

birdbrained:
## Ticket

https://openscience.atlassian.net/browse/SVCS-672

## Purpose

In an effort to prepare waterbutler for a potential upgrade to a newer version of aiohttp, make sure that all requests are using a context manager to set up and break down requests.

The scope of this PR is constrained to only the s3 provider.

## Changes

Changes all instances where `aiohttp.ClientResponse.release` is called to use `async with`.

## Side effects

Shouldn't be any.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -64,7 +64,7 @@ def test(ctx, verbose=False, types=False):
     if types:
         mypy(ctx)
 
-    cmd = 'py.test --cov-report term-missing --cov waterbutler tests'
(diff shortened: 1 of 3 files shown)


cslzchen:
Update:
* I found the ticket: https://openscience.atlassian.net/browse/SVCS-672
* As discussed, @felliott and @birdbrained , we will put this ticket on hold until the analysis for `aiohttp` upgrade is done

birdbrained:
Handle this when we do the aiohttp upgrade

PR #292, opened 2017-10-31
[WIP] [SVCS-543] Handle GitHub rate limit during inter provider move/copy

TomBaxter:
## Ticket

[SVCS-543](https://openscience.atlassian.net/browse/SVCS-543)

PR #330, opened 2018-03-26
[ENG-302] Handle GitHub Rate Limiting

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/ENG-302

## Purpose

This PR replaces https://github.com/CenterForOpenScience/waterbutler/pull/292

h/t to @TomBaxter for the core part implement which is fully functional, I just polished the code

GitHub API has a [rate-limiting](https://developer.github.com/v3/#rate-limiting) feature that only allow 5000 requests per hour. The limit is not a **rate** but a **count**.  To handle this limitation, this PR enables WB to rate-limiting GitHub move/copy requests by allowing 1-2 requests per second.

### GitHub VS WB

* GitHub rate-limiting is enforced per user. For example, API requests made to GitHub by non-OSF products users have authorized **count** against the quota.

```
For API requests using Basic Authentication or OAuth, you can make up to 5000 requests
per hour. Authenticated requests are associated with the authenticated user, regardless
of whether Basic Authentication or an OAuth token was used. This means that all OAuth
applications authorized by a user share the same quota of 5000 requests per hour when
they authenticate with different tokens owned by the same user.
```

* WB rate-limiting is enforced per provider instance. However, there are a few features that actively preventing it from hitting the rate-limit cap when multiple instances are present and when users are making a reasonable amount of requests using GitHub API outside OSF and WB.
  * Reserve a portion (default is 10%) of the pool for requests that does not go through WB's rate-limiting and for multiple instances scenario mentioned below.
  * Actively (default is 1 minute) updates the request rate based the response headers GitHub returns.
  * Slow down the request rate to 1 per 100 seconds if reaching the limit.

### What Will Work

If user tries to copy two or more repos of more than 2000 files at the same time, WB rate-limit can still work because none of them can make all the requests at the same time. The request rate limit gets refreshed every minute to reflect the most updated info from GitHub.

### What Will Not Work

As mentioned the request rate limit gets updated every minute. Within in that limit each instance makes request on their own pace assuming they are the only one using the quota. However, if users have other non-OSF apps that suddenly used up the quota within that minute, WB will fail.

## Changes

This is the original commit message (slightly modified) by Tom.

This PR only rate limits the GitHub provider's `.make_requests()` that run in celery tasks. Currently, the only tasks are "move" and "copy". The distinction is made by adding a `task_id` to the providers after they are instantiated in the task definition.

This PR uses the rate limit information returned in **all** GitHub response headers to limit the number of `.make_requests()` calls in such a way that the GitHub rate limit is never reached. In practice this usually results in a rate limit between 1-2 calls a second.

As a bonus, this PR adds a `GitHubRateLimitExceeded` exception, that will be checked for with each `.make_request()`. This exception ~~should never~~ is not expected to be raised during a move/copy celery task unless users suddenly use up the quota outside OSF or WB. This also helps non-celery requests as well.

## Side effects

GitHub move/copy request are rate-limited to 1-2 per second, no matter how many/few files are there. 

## QA Notes

Here is a dedicated [repo](https://github.com/cslzchen/OSF-Addon-Folder/tree/develop/feature-tasks/rate-limit) I created for test. Pease fork it for testing purpose. It has folders containing a variety number of files.

* Stage one, use the first 4 folders (files-[00128|00256|00512|01024]) to test and feel the side effect of speed decrease because of the rate limit.
* Stage two, [files-16384](https://github.com/cslzchen/OSF-Addon-Folder/tree/develop/feature-tasks/rate-limit/files-16384) contains 4 folders, each with 4096 files. Any of them should fail on `Prod` but  should succeed on `Staging`. Locally, it takes 2.5 hours to finish each one.
* Final stage, try copy 2 or more folders at the same time, each with a `different` OSF user. Please expect around 6 - 12 hours and it needs a stable `Staging`. You may want to skip this one.

## Deployment Notes

TBD

### DIFF
--- a/tests/providers/github/test_provider.py
+++ b/tests/providers/github/test_provider.py
@@ -102,7 +102,7 @@ def other_provider(other_auth, other_credentials, other_settings, provider_fixtu
 @pytest.fixture
 def rate_limit_provider(provider):
 
-    provider.task_id = 9876543210
+    provider.is_celery_task = True
     provider.rl_remaining = 500
     provider.rl_reset = 600
     provider.rl_available_tokens = 0
@@ -1465,7 +1465,7 @@ class TestRateLimit:
     async def test_make_request_not_celery(self, mock_time, rate_limit_provider):
 
         mock_provider = rate_limit_provider
-        del mock_provider.task_id
+        mock_provider.is_celery_task = False
 
         aiohttpretty.register_json_uri(
             'GET',
@@ -1549,7 +1549,7 @@ class TestRateLimit:
         """Test that non-celery requests throw rate limit exceeded error"""
 
         mock_provider = rate_limit_provider
-        del mock_provider.task_id
+        mock_provider.is_celery_task = False
 
         aiohttpretty.register_json_uri(
             'GET',
@@ -1579,7 +1579,7 @@ class TestRateLimit:
         """Test that non-celery requests re-throw non-limit exceeded errors"""
 
         mock_provider = rate_limit_provider
-        del mock_provider.task_id
+        mock_provider.is_celery_task = False
 
         aiohttpretty.register_json_uri(
             'GET',
--- a/tests/providers/github/test_provider.py
+++ b/tests/providers/github/test_provider.py
@@ -2,6 +2,7 @@ import io
 import os
 import copy
 import json
+import time
 import base64
 import hashlib
 from http import client
@@ -15,14 +16,13 @@ from waterbutler.core import exceptions
 
 from waterbutler.providers.github import GitHubProvider
 from waterbutler.providers.github.path import GitHubPath
+from waterbutler.providers.github import settings as github_settings
+from waterbutler.providers.github import exceptions as github_exceptions
 from waterbutler.providers.github.metadata import (GitHubRevision,
                                                    GitHubFileTreeMetadata,
                                                    GitHubFolderTreeMetadata,
                                                    GitHubFileContentMetadata,
                                                    GitHubFolderContentMetadata)
-from waterbutler.providers.github import settings as github_settings
-from waterbutler.providers.github.exceptions import GitHubUnsupportedRepoError
-
 
 from tests.providers.github.fixtures import (crud_fixtures,
                                              revision_fixtures,
@@ -100,6 +100,27 @@ def other_provider(other_auth, other_credentials, other_settings, provider_fixtu
     provider._repo = provider_fixtures['repo_metadata']
     provider.default_branch = provider_fixtures['repo_metadata']['default_branch']
     return provider
+
+
+@pytest.fixture
+def rate_limit_provider(provider):
+    provider.task_id = 1234567890
+    provider.rate_limit = 1
+    provider.rate_limit_remaining = 100
+    provider.rate_limit_reset = 200
+    provider.rate_limit_updated = 10
+    provider.rate_limit_tokens = 0
+    provider.rate_limit_tokens_updated = 90
+    return provider
+
+
+@pytest.fixture
+def patch_time(monkeypatch):
+
+    def mytime():
+        return 100
+
+    monkeypatch.setattr(time, 'time', mytime)
 
 
 class TestHelpers:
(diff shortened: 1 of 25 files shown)


waterbutler/tasks/move.py
```diff
@@ -1,28 +1,34 @@
...
-    dest_path, dest_provider = dest_bundle.pop('path'), utils.make_provider(**dest_bundle.pop('provider'))
+    src_path = src_bundle.pop('path')
+    src_provider = utils.make_provider(**src_bundle.pop('provider'))
+    dest_path = dest_bundle.pop('path')
+    dest_provider = utils.make_provider(**dest_bundle.pop('provider'))
+
+    # TODO: should we add `task_id` explicitly as a property for core provider?
+    # Save celery task ID. Used to identify provider object running in celery
+    src_provider.task_id = move.request.id
+    dest_provider.task_id = move.request.id
```
cslzchen: - Providers do not explicitly has a property called `task_id`. Tom uses this ID to decide whether a request should be rate-limited by WB. However, one side-effect is, once the ID is set, all future requests by the same instance will be rate limited. I am fine with it.
- Previously, Tom has `dest_path.task_id` which I believe is a typo. In addition, I don't see that we really need the `task_id` for `dest_provider` for this scenario (copy/register GH repo)

waterbutler/providers/github/provider.py
```diff
@@ -68,6 +76,64 @@ def __init__(self, auth, credentials, settings):
...
+        rate limit remaining, time until next reset, etc.
+        """
+
+        # Only update `expects` when it exists in the original request
+        expects = kwargs.get('expects', None)
+        if expects:
+            kwargs.update({'expects': expects + (int(HTTPStatus.FORBIDDEN), )})
+
+        # Only handle rate limiting for WB Celery tasks
+        if not getattr(self, 'task_id', None):
```
cslzchen: As mentioned, rate-limiting is turned on when requests get kicked off to Celery.

cslzchen:
Update, as mentioned in the ticket, credit also goes to https://quentin.pradet.me/blog/how-do-you-rate-limit-calls-with-aiohttp.html where our solution is inspired.

waterbutler/providers/github/provider.py
```diff
@@ -955,3 +1023,102 @@ def _prune_subtrees(self, tree):
...
+        return True
+
+    def _rl_update_req_rate(self) -> None:
+        """Calculate the rate at which WB make requests.  Instead of maximizing the rate with
+        ``.rl_remaining  / (.rl_reset - now)``, WB sets a lower one by reserving "10% of the total
+        number of remaining requests plus 10 extra ones". This ensures that WB won't run out of
+        tokens when there are a REASONABLE amount of additional requests made for the same user
+        during the move/copy process.
+
+        TODO: should the "10% + 10 extra", and the "1 request per 100 seconds" be configurable?
```
birdbrained: Do we have an analysis ticket for this?

waterbutler/tasks/move.py
```diff
@@ -1,28 +1,34 @@
...
 
     logger.info('Starting moving {!r}, {!r} to {!r}, {!r}'
                 .format(src_path, src_provider, dest_path, dest_provider))
 
-    metadata, errors = None, []  # type: ignore
+    metadata, errors = None, []
```
birdbrained: how come this was ignored before?

tests/providers/github/test_provider.py
```diff
@@ -2,31 +2,27 @@
 import os
 import copy
 import json
-import base64
```
birdbrained: was this stale?

waterbutler/providers/github/provider.py
```diff
@@ -955,3 +1023,102 @@ def _prune_subtrees(self, tree):
...
+            raise exc_to_raise
+
+    async def _rl_check_available_tokens(self) -> None:
+        """Checks if there are any available tokens.  If so, consume one and return.  Otherwise,
+        continue to make attempts to add new tokens.  Each failed attempt must wait for a short
+        period of time before making another one.  The wait time is defined by `.RL_TOKEN_ADD_DELAY`
+        and its default is 1 second.
+        """
+
+        # Keep waiting and making attempts to add new tokens if there is no one available.
```
birdbrained: pedantic: "is not" or "isn't"

waterbutler/providers/github/provider.py
cslzchen: No. This **TODO** is for review. I think I will make them settings during responding to CR.

tests/providers/github/test_provider.py
cslzchen: It wasn't used before this PR.

waterbutler/providers/github/provider.py
```diff
@@ -1311,7 +1311,8 @@ def _rl_add_more_tokens(self) -> None:
             # remaining requests and the time between now and next limit reset. If no other major
             # requests (i.e. another copy/move) are being made at the same time, this reference
             # request rate keeps increasing slightly.
-            rl_req_rate = (self.rl_remaining - self.rl_reserved) / (self.rl_reset - time.time())
+            seconds_until_reset = max((self.rl_reset - time.time()), 1)
```
cslzchen: I remembered that I did run into negative `rl_req_rate` during testing. I thought I had fixed it and apparently either I forgot or the fixe didn't work at all. Great catch!

cslzchen:
Forgot to mention that 1) I changed the target branch from `develop` to [release/sequoia](https://github.com/CenterForOpenScience/waterbutler/tree/release/sequoia) so only GH RL commits are shown for diff and 2) I reviewed commit by commit instead of the full diff :)

waterbutler/core/provider.py
```diff
@@ -180,15 +180,13 @@ def build_headers(self, **kwargs) -> dict:
...
                 response = await aiohttp.request(method, non_callable_url, *args, **kwargs)
-                self.provider_metrics.append('requests.verbose',
-                                             ['OK', response.status, non_callable_url])
+                self.provider_metrics.incr('requests.tally.ok')
                 if expects and response.status not in expects:
                     raise (await exceptions.exception_from_response(response, error=throws, **kwargs))
                 return response
             except throws as e:
-                self.provider_metrics.append('requests.verbose', ['NO', e.code, non_callable_url])
+                self.provider_metrics.incr('requests.tally.nok')
```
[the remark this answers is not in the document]
felliott: It's not the performance or cost, but the size.  Keen will reject metrics payloads over 900kb.  This is one place where the metrics payload can grow *really fast*.  An alternative would be to delete keys matching `/\.verbose$/`, but I decided to take this out as YAGNI.  And so far, I don't think we have.

waterbutler/providers/github/provider.py
felliott: You did fix it, but I undid it in a prior commit.  When I realized what I had done, I decided to re-add it in a slightly different way.  Instead of bumping the value of `self.rl_reset` to `self.rl_reset + 3600`, I opted to set a minimum value for `self.reset - time.time()`.  They should be approximately the same thing, but I think this is a little bit clearer.

PR #320, opened 2018-02-16
[SVCS-595] Update Box and GoogleDrive to Pass `mypy`

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-595
Replaces: https://github.com/CenterForOpenScience/waterbutler/pull/314

## Purpose

Credits goes to @TomBaxter 👍.

Fix (or ignore) the failed `mypy` tests:

```bash
waterbutler/providers/googledrive/provider.py:380: error: Value of type "Union[Dict[Any, Any], BaseGoogleDriveMetadata, List[Union[BaseGoogleDriveMetadata, Dict[Any, Any]]]]" is not indexable
waterbutler/providers/googledrive/provider.py:380: error: No overload variant of "__getitem__" of "list" matches argument types [builtins.str]
waterbutler/providers/box/provider.py:220: error: Incompatible return value type (got "BaseBoxMetadata", expected "Tuple[Union[BoxFileMetadata, BoxFolderMetadata], bool]")
waterbutler/providers/box/provider.py:247: error: Incompatible return value type (got "BaseBoxMetadata", expected "Tuple[BaseBoxMetadata, bool]")
waterbutler/providers/box/provider.py:453: error: Too few arguments for "_serialize_item" of "BoxProvider"
waterbutler/providers/box/provider.py:511: error: Incompatible return value type (got "Tuple[Union[BoxFileMetadata, BoxFolderMetadata], bool]", expected "BaseBoxMetadata")
waterbutler/providers/box/provider.py:514: error: Item "BoxFileMetadata" of "Union[BoxFileMetadata, BoxFolderMetadata]" has no attribute "_children"
waterbutler/providers/box/provider.py:515: error: Incompatible return value type (got "Tuple[Union[BoxFileMetadata, BoxFolderMetadata], bool]", expected "BaseBoxMetadata")
```

## Changes

TBD

## Side effects

TBD

## QA Notes

No QA

## Deployment Notes

No special deploy notes

PR #332, opened 2018-04-05
[SVCS-595] Clean Up Existing Type Annotation Failures

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-595

## Purpose

Clean up existing type annotation failures for several providers:

* Bitbucket
* Box
* Dataverse
* figshare
* File System
* GitHub
* Google Drive

This PR replaces: https://github.com/CenterForOpenScience/waterbutler/pull/320 and https://github.com/CenterForOpenScience/waterbutler/pull/314

h/t to @TomBaxter for the initial efforts

### A 🐛 In Box

 In addition to type annotation mentioned above, this PR fixed a bug in Box where `._intra_move_copy_metadata()` calls a buggy `._get_folder_meta()` that did
  not call `._serialize_item()` with correct arguments.

### A 🐞 In `path`

I guess a long time ago `path` is of type `str`, which is probably the reason why there are many places where `path` is annotated or documented as `str` while  it should be `WaterBulterPath` or its sub classes. I only updated where I fixed the typing.

### Minimal Style Update

Type annotation leads to longer lines, especially our code base. I try not to do a refactor by only changing a few things and only where I fixed the typing.

* `from ... import settings` -> `from ... import settings as pd_settings` for all providers affected in this PR. I hate "shadowing" and pd stands for provider
* `import typing` -> `from typing import List, Sequence, Tuple, Union ...`.  This change significantly reduces the length of method definition line.
* Similarly, for one or two providers, `import ...metadata` -> `from ...metadata import` and `import ...path as wb_path` -> `from ...path import WaterButlerPath, WaterButlerPathPart ...` 
* Combine imports from `waterbutler.core`. No need for  them to be separate
* If we use type annotation for a method, we should use it for all eligible args and for return type.

## Side effects

No

## QA Notes

No QA

## Deployment Notes

Nothing special


waterbutler/providers/box/provider.py
```diff
@@ -444,19 +427,15 @@ def default_headers(self) -> dict:
...
+                               folder: bool=False) -> Union[dict, List[BoxFolderMetadata]]:
         if folder:
             async with self.request(
                 'GET', self.build_url('folders', path.identifier),
                 expects=(200, ), throws=exceptions.MetadataError,
             ) as resp:
                 data = await resp.json()
-                # FIXME: Usage does not match function call signature!  Dead code or bug?
-                return data if raw else self._serialize_item(data)
+                return data if raw else self._serialize_item(data, path)
```
cslzchen: The bug that was discovered by @TomBaxter, which also failed `mypy` test.

waterbutler/providers/github/provider.py
```diff
@@ -289,9 +289,9 @@ def can_intra_copy(self, other, path=None):
         else:
             await self._delete_file(path, message, **kwargs)
 
-    async def metadata(self, path, **kwargs):
+    async def metadata(self, path: GitHubPath, **kwargs):  # type: ignore
         """Get Metadata about the requested file or folder
-        :param str path: The path to a file or folder
+        :param GitHubPath path: The path to a file or folder
```
cslzchen: Fixed the wrong type.

waterbutler/providers/dataverse/provider.py
```diff
@@ -113,16 +111,17 @@ def can_duplicate_names(self):
...
         return sum(self._metadata_cache.values(), [])
 
-    async def download(self, path, revision=None, range: typing.Tuple[int, int]=None, **kwargs):
+    async def download(self, path: WaterButlerPath, revision: str=None,  # type: ignore
+                       range: Tuple[int, int] = None, **kwargs) -> streams.ResponseStreamReader:
         """Returns a ResponseWrapper (Stream) for the specified path
         raises FileNotFoundError if the status from Dataverse is not 200
 
-        :param str path: Path to the file you want to download
+        :param WaterButlerPath path: Path to the file you want to download
```
cslzchen: Fixed another wrong type.

felliott:
Love it!  Thanks, @cslzchen and @TomBaxter!

PR #335, opened 2018-04-11
Fix all import ordering

birdbrained:


felliott:
Sorry, I won't be accepting sweeping changes to fix minor style issues.  It just adds an extra layer of obfuscation when doing code archaeology.

PR #337, opened 2018-04-11
 [PLAT-639] Send along the user ID when asking for children metadata from the osf

erinspace:
## Ticket
https://openscience.atlassian.net/browse/PLAT-639

## Purpose
When checking to see if a user has seen a file, it'd be lovely to have the requesting user's _id in the request itself -- this isn't usually necessary because the request has already been authorized and so shouldn't need to send any auth along. Now, since we're planning on highlighting files that a user hasn't yet seen, we need to know which user is asking for the file tree to do the appropriate query to find the files they haven't seen yet!

## Changes
Send the requesting user's id in the query params of the request for children metadata. This allows the user ID to be accessed OSF-side in `osfstorage_get_children` to use to see if the current user has seen the files yet or not!

## Side effects
None anticipated

## QA Notes
This is a small internal change, so should not need any QA!

## Deployment Notes
Nope!

### DIFF
--- a/waterbutler/providers/osfstorage/provider.py
+++ b/waterbutler/providers/osfstorage/provider.py
@@ -493,7 +493,7 @@ class OSFStorageProvider(provider.BaseProvider):
     async def _children_metadata(self, path):
         async with self.signed_request(
             'GET',
-            self.build_url(path.identifier, 'children'),
+            self.build_url(path.identifier, 'children', user_id=self.auth['id']),
             expects=(200, )
         ) as resp:
             resp_json = await resp.json()


waterbutler/providers/osfstorage/provider.py
```diff
@@ -493,7 +493,7 @@ def signed_request(self, *args, **kwargs):
     async def _children_metadata(self, path):
         async with self.signed_request(
             'GET',
-            self.build_url(path.identifier, 'children'),
+            self.build_url(path.identifier, 'children', user_id=self.auth['id']),
```
cslzchen: From the OSF side PR [here](https://github.com/CenterForOpenScience/osf.io/pull/8295/files#diff-2112d9755e4f6cef9f80f95f90a4be92R169), I see the `user_id` is a must. We may want to double check the case when it is a logged-out user viewing a public project (if this is in the scope of this ticket). If this turns out to be an issue, we need to modify the `osfstorage_get_children` view on OSF.

waterbutler/providers/osfstorage/metadata.py
```diff
@@ -84,6 +84,7 @@ def extra(self):
             'version': self.raw['version'],
             'downloads': self.raw['downloads'],
             'checkout': self.raw['checkout'],
+            'latestVersionSeen': self.raw.get('latestVersionSeen', None),
```
cslzchen: For me, `latestVersionSeen` means the latest **version** that has been seen. However, I cannot think of a better name. A comment here for what it is should be helpful.

waterbutler/providers/osfstorage/provider.py
erinspace: Yes, this makes sense! I will absolutely do this, thank you for thinking of that!

waterbutler/providers/osfstorage/metadata.py
erinspace: Yes, it should be this, the latest version! Had a lot of trouble naming this one. Ideally, we'd be able to know the version that's being accessed via the request as well,  but I'm not sure if that's possible at the moment. So, as we have it, this will reflect if the most recent version of the file has been seen.

cslzchen: Sounds good!

erinspace:
@cslzchen Comments addressed! Thanks so much for the through review! Changed `latestVersionSeen` to a dict, and but kept it as `None` for a user who has never seen any version of the file.  So, if a user has seen a previous version of a file and the most recent version, `latestVersionSeen` is `{"user" : "user_id", "seen": True}`, if they have never seen a version of that file, `latestVersionSeen` is `None`. Does this make sense?

As for tests, yes a PR would be just fine! Or if you'd prefer, I am more than happy to take a swing at adding tests myself (tho I might bug you with questions), so whichever you'd prefer is ok by me.

cslzchen:
And rebase before Fitz rebase it 🚔

PR #338, opened 2018-04-18
[SVCS-693] Fix `resp.json()` throwing warnings about `mimetype`

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-693

## Purpose

`resp.json()` will throw the warning "Attempt to decode JSON with unexpected mimetype: `%s`" if the "Content-Type" is not of a `json` type.  This is a correct and expected warning.

However, providers (such as S3 and GoogleCloud) using XML API endpoint will pollute the WB celery log with this warning when checking file existences before each copy/move.  In addition, some 404s responses have a `str` or `bytes` body regardless of the "Content-Type".

Current fix is to skip parsing the response body for HTTP HEAD requests.  We cannot simply rely on the header "Content-Type" to decide whether to call `.json()` or `.read()`.  The warning should be still logged at a level that it is useful but not annoying.

## Changes

Skip parsing response body for HTTP HEAD requests.

## Side effects

Fix an issue where response are not released.
Update an outdated comments for Google Cloud when moving from JSON to XML.

## QA Notes

QA is not needed.

## Deployment Notes

No

### DIFF
--- a/waterbutler/core/exceptions.py
+++ b/waterbutler/core/exceptions.py
@@ -257,26 +257,45 @@ class InvalidProviderConfigError(ProviderError):
 
 
 async def exception_from_response(resp, error=UnhandledProviderError, **kwargs):
-    """Build and return, not raise, an exception from a response object
+    """Build and return, not raise, an exception from a response object.
 
-    :param Response resp: An aiohttp.Response stream with a non 200 range status
+    Quirks:
+
+    ``resp.json()`` will throw warning "Attempt to decode JSON with unexpected mimetype: %s'" if the
+    "Content-Type" is not of a json type.  The warning is an expected one.  However, providers such
+    as S3 and GoogleCloud which use XML API endpoint pollutes the WB celery log with this warning
+    when checking file existences before each copy/move.  In addition, some 404s responses have a
+    string body regardless of the `Content-Type`.
+
+    Current fix is skip parsing the body for HEAD requests.  We cannot simply rely on the header
+    "Content-Type" to decide whether to call ``.json()`` or ``.read()``.  The warning should be
+    logged at a level that it is useful but not annoying.
+
+    :param ClientResponse resp: An ``aiohttp.ClientResponse`` stream with a non-200 range status
     :param Exception error: The type of exception to be raised
-    :param dict \*\*kwargs: Additional context to extract information from
-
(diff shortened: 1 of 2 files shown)


waterbutler/core/exceptions.py
```diff
@@ -257,26 +257,45 @@ def __init__(self, provider_name, message=None):
...
     try:
+        # Try to parse response as String.
         data = await resp.read()
         return error({'response': data.decode('utf-8')}, code=resp.status)
     except TypeError:
         pass
 
-    # When all else fails return the most generic return message
+    # When all else fails, return the most generic return message.
+    await resp.release()
```
cslzchen: When all else fails, the response needs to be released before throwing the error. Fortunately, this hasn't been an issue  because the two `try`s catch almost all the cases.

waterbutler/core/exceptions.py
```diff
@@ -257,26 +257,45 @@ def __init__(self, provider_name, message=None):
...
+        await resp.release()
+        return error(DEFAULT_ERROR_MSG.format(response=resp), code=resp.status)
+
     try:
-        # Try to make an exception from our received json
+        # Try to parse response body as JSON.
         data = await resp.json()
         return error(data, code=resp.status)
-    except Exception:
+    except (TypeError, json.JSONDecodeError):
```
cslzchen: Now catches specific exceptions. Do we prefer nested `try .. except ...` or using `pass`?

waterbutler/core/exceptions.py
```diff
@@ -257,26 +257,45 @@ def __init__(self, provider_name, message=None):
...
+    :param ClientResponse resp: An ``aiohttp.ClientResponse`` stream with a non-200 range status
+    :param Exception error: The type of exception to be raised
+    :param dict \*\*kwargs: Additional context are ignored
     :rtype :class:`WaterButlerError`:
     """
+
+    if resp.method.upper() == 'HEAD':
+        # HEAD requests must have an empty body
+        await resp.release()
+        return error(DEFAULT_ERROR_MSG.format(response=resp), code=resp.status)
```
cslzchen: Do we want to log request URL when it is a signed request? This is what WB does now. I am OK with it since the URL has an expiration and the URL ends up with exception.

PR #343, opened 2018-05-04
[SVCS-701] OSFStorage: Only Enable Intra Move/Copy For Same Region

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-701

## Purpose

Only allow intra move/copy within the same region (for the same bucket) for OSFStorage.

## Changes

* Add a dedicated method `is_same_region()` to `core.provider` and `osfstorage.provider`. Currently implementation is based on the assumption that we use a 1-to-1 mapping between buckets and regions. If we change this assumption in the future, only `is_same_region()` needs to be updated.

* `can_intra_copy()` and `can_intra_move()` now calls the aforementioned `is_same_region()`

* Update provider tests for OSFStorage. (No refactor, just minimal changes that are necessary.)

* Update a comment in GCloud to reflect the latest change in OSFStorage settings. 

## Side effects

No

## QA Notes

This feature cannot be tested until the OSF has region.

## Deployment Notes

No

### DIFF
--- a/tests/providers/osfstorage/fixtures.py
+++ b/tests/providers/osfstorage/fixtures.py
@@ -200,7 +200,23 @@ def provider_and_mock2(monkeypatch, auth, credentials, settings):
 
 @pytest.fixture
 def provider(auth, credentials, settings):
+    settings.update({
+        'storage': {
+            'bucket': 'mocke_bucket_1',
+        }
+    })
     return OSFStorageProvider(auth, credentials, settings)
+
+
+@pytest.fixture
+def provider_other(auth, credentials, settings):
+    settings_other = dict(settings)
+    settings_other.update({
+        'storage': {
+            'bucket': 'mocke_bucket_2',
+        }
+    })
+    return OSFStorageProvider(auth, credentials, settings_other)
 
 
 @pytest.fixture
--- a/waterbutler/providers/osfstorage/provider.py
+++ b/waterbutler/providers/osfstorage/provider.py
@@ -145,11 +145,21 @@ class OSFStorageProvider(provider.BaseProvider):
     def can_duplicate_names(self):
         return True
 
+    def is_same_region(self, other):
+        assert isinstance(other, self.__class__), 'Cannot compare region for providers of ' \
+                                                  'different provider classes.'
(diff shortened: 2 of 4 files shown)


waterbutler/providers/osfstorage/provider.py
```diff
@@ -145,11 +145,21 @@ def make_provider(self, settings):
     def can_duplicate_names(self):
         return True
 
+    def is_same_region(self, other):
```
cslzchen: Add a dedicated methods for check region.

waterbutler/core/provider.py
```diff
@@ -411,6 +411,18 @@ def request(self, *args, **kwargs):
 
         return dest_path
 
+    def is_same_region(self, other: 'BaseProvider') -> bool:
```
cslzchen: Similar to `can_intra_copy()` and `can_intra_move()`, add `is_same_region()` to the core provider. This might be redundant. I can remove this if you prefer.

tests/providers/osfstorage/fixtures.py
```diff
@@ -200,9 +200,25 @@ def provider_and_mock2(monkeypatch, auth, credentials, settings):
...
+            'bucket': 'mocke_bucket_1',
+        }
+    })
     return OSFStorageProvider(auth, credentials, settings)
 
 
+@pytest.fixture
+def provider_other(auth, credentials, settings):
+    settings_other = dict(settings)
+    settings_other.update({
```
cslzchen: I decide not to refactor the full tests. Just make necessary changes to update the code.

waterbutler/core/provider.py
felliott: In addition, while it's called `is_same_region`, the first and only implementation compares the bucket instead of the region.  The fact that we use one bucket per region is an osfstorage-specific detail that could be misleading for someone who only reads the implementation of `osfstorage`'s `intra_copy` code.

Yeah, I don't think this should be in the base provider.  At least not until we really need it.  I could see it being included in a `BucketBasedProvider` mixin down the road, but that's out of scope for this ticket.

PR #348, opened 2018-06-04
[SVCS-545] Box Chunked Upload

birdbrained:
This replaces an older pr https://github.com/CenterForOpenScience/waterbutler/pull/291

# Ticket

https://openscience.atlassian.net/browse/SVCS-545

# Purpose

This allows Box to upload very large files using their chunked upload endpoint.

# Changes

Adds several functions to start an upload session, upload the individual chunks and commit the upload into one file. Includes unittests.

# Side effects

None that I know of.

# QA Notes

A file must be over 50MB to trigger a chunked upload and the progress bar is not a good indicator of how much of the download has completed.

# Deployment Notes

Should probably correspond with removing upload limits for dropbox in Fangorn.js in osf.io


waterbutler/providers/box/provider.py
```diff
@@ -1,15 +1,19 @@
...
 from http import HTTPStatus
 from typing import List, Tuple, Union
 
-import aiohttp
-
 from waterbutler.core.path import WaterButlerPath
 from waterbutler.core import exceptions, streams, provider
 
-from waterbutler.providers.box import settings as pd_settings
+from waterbutler.providers.box.settings import (
```
cslzchen: Fix the import style.

waterbutler/providers/box/provider.py
```diff
@@ -55,7 +60,7 @@ def __init__(self, auth, credentials, settings):
         response = await self.make_request(
             'get',
             self.build_url(files_or_folders, obj_id, fields='id,name,path_collection'),
-            expects=(200, 404,),
+            expects={200, 404},
```
cslzchen: Similarly to S3 and Dropbox, revert the `[] () {}` membership check.

waterbutler/providers/box/provider.py
```diff
@@ -270,50 +273,194 @@ def default_headers(self) -> dict:
...
             throws=exceptions.DownloadError,
         )
         logger.debug('download-headers:: {}'.format([(x, resp.headers[x]) for x in resp.headers]))
 
         return streams.ResponseStreamReader(resp)
 
-    async def upload(self,  # type: ignore
-                     stream: streams.BaseStream, path: WaterButlerPath, conflict: str='replace',
-                     **kwargs) -> Tuple[BoxFileMetadata, bool]:
+    async def upload(
```
cslzchen: In this method, `if stream.size > self.NONCHUNKED_UPLOAD_LIMIT:` has been checked twice. The code for chunked upload are duplicated with slightly different method names. I guess its from a bad merge or conflicts. Please go through the code and fix the code.

waterbutler/providers/box/provider.py
```diff
@@ -270,50 +273,194 @@ def default_headers(self) -> dict:
...
         if path.identifier and conflict == 'keep':
             path, _ = await self.handle_name_conflict(path, conflict=conflict, kind='folder')
             path._parts[-1]._id = None
 
-        stream.add_writer('sha1', streams.HashStreamWriter(hashlib.sha1))
+        if stream.size > self.NONCHUNKED_UPLOAD_LIMIT:
+            data = await stream.read()
+            entry = await self._chunked_upload(path, data)
+        else:
+            stream.add_writer('sha1', streams.HashStreamWriter(hashlib.sha1))
```
cslzchen: Similar to S3 and Dropbox, add a `_contiguous_upload()`.

waterbutler/providers/box/provider.py
```diff
@@ -270,50 +273,194 @@ def default_headers(self) -> dict:
...
-        data_stream = streams.FormDataStream(
-            attributes=json.dumps({
-                'name': path.name,
-                'parent': {
-                    'id': path.parent.identifier
-                }
-            })
-        )
-        data_stream.add_file('file', stream, path.name, disposition='form-data')
+            async with self.request(
```
cslzchen: I am good with `async with`. 👍

waterbutler/providers/box/provider.py
```diff
@@ -270,50 +273,194 @@ def default_headers(self) -> dict:
...
+            entry = data['entries'][0]
+            if stream.writers['sha1'].hexdigest != entry['sha1']:
+                raise exceptions.UploadChecksumMismatchError()
+
+        created = path.identifier is None
+        path._parts[-1]._id = entry['id']
+
+        return BoxFileMetadata(entry, path), created
+
+    async def _chunked_upload(
```
cslzchen: Minor style, not a blocker, just for style consistence, though I prefer yours.

```python
async def _chunked_upload(self,
                          path: WaterButlerPath,
                          data: bytes) -> dict:

File: waterbutler/providers/box/provider.py [hunk 7]
```diff
@@ -270,50 +273,194 @@ def default_headers(self) -> dict:
...
+        session_data = await self._create_chunked_upload_session(path, data)
+
+        try:
+            # Step A.3 split the data into chunks and upload them to box.
+            parts = await self._upload_chunks(data, session_data)
+
+            # Step A.4 complete the session and return the upload file's metadata.
+            return await self._complete_chunked_upload_session(session_data, parts, data_sha)
+
+        except exceptions.UploadError:
```
cslzchen: 👍 for exception handling

waterbutler/providers/box/provider.py
```diff
@@ -270,50 +273,194 @@ def default_headers(self) -> dict:
...
+                'Digest': 'sha={}'.format(data_sha)
+            },
+            expects={201},
+            throws=exceptions.UploadError,
+        ) as resp:
+            entry = (await resp.json())['entries'][0]
+
+        return entry
+
+    async def _abort_chunked_upload(
```
cslzchen: Does box have similar issues to S3 that abort may only be partially successful?

birdbrained: It doesn't say anything about it on the website. Doesn't seem to be breaking though

waterbutler/providers/box/provider.py
birdbrained: Should be good noew

waterbutler/providers/box/provider.py
```diff
@@ -270,28 +283,54 @@ def default_headers(self) -> dict:
...
+            entry = await self._chunked_upload(path, data)
+        else:
+            await self._contiguous_upload(stream, path)
+
+        created = path.identifier is None
+        path._parts[-1]._id = entry['id']
+
+        return BoxFileMetadata(entry, path), created
+
+    async def _contiguous_upload(
```
cslzchen: This method should return the `entry`.

waterbutler/providers/box/provider.py
```diff
@@ -311,9 +350,149 @@ def default_headers(self) -> dict:
...
+
+        # Step 2 create an upload session with box and recieve session id.
+        session_data = await self._create_chunked_upload_session(path, data)
+
+        try:
+            # Step A.3 split the data into chunks and upload them to box.
+            parts = await self._upload_chunks(data, session_data)
+
+            # Step A.4 complete the session and return the upload file's metadata.
+            while True:
```
cslzchen: Don't do `while True`. Set a limit for retries.

cslzchen: For exceptions, please take a look at the update S3 PR https://github.com/CenterForOpenScience/waterbutler/pull/325/files#diff-74bbac431e282383494483a4d0af3031R249 for better handling for both the developers and the users.

waterbutler/providers/box/provider.py
cslzchen: I see. So the "abort" either fails or succeeds? We need to record both cases in log and  inform the users as well. My suggestion is that "abort " returns a `bool` type so the caller knows whether it succeeds or fails. Please take a look at S3's updated PR: https://github.com/CenterForOpenScience/waterbutler/pull/325/files#diff-74bbac431e282383494483a4d0af3031R330

waterbutler/providers/box/provider.py
```diff
@@ -311,9 +350,149 @@ def default_headers(self) -> dict:
...
+            headers={
+                'Content-Type:': 'application/json',
+                'Digest': 'sha={}'.format(data_sha)
+            },
+            expects={201},
+            throws=exceptions.UploadError,
+        ) as resp:
+            entry = (await resp.json())['entries'][0]
+
+        return entry
```
cslzchen: What's in this `entry`? When error happens, this `entry` is returned all the way back to the main `upload()` without triggering any exception ...

waterbutler/providers/box/provider.py
cslzchen: Looks good.

waterbutler/providers/box/provider.py
```diff
@@ -270,28 +283,54 @@ def default_headers(self) -> dict:
...
+        conflict: str='replace',
+        **kwargs
+    ) -> typing.Tuple[BoxFileMetadata, bool]:
+
         if path.identifier and conflict == 'keep':
             path, _ = await self.handle_name_conflict(path, conflict=conflict, kind='folder')
             path._parts[-1]._id = None
 
+        if stream.size > self.NONCHUNKED_UPLOAD_LIMIT:
+            data = await stream.read()
```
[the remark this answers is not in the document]
birdbrained: It looks like we need the SHA before we can start the upload to Box.

waterbutler/providers/box/provider.py
birdbrained: I think it shouldn't matter - wb will return an error in both these cases. the upload is only aborted during an error, and an error with the abort is still an UploadError

waterbutler/providers/box/provider.py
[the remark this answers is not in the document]
birdbrained: Hm it looks like we dont need the sha till the file is fully uploaded, Is there a way to get the full file after the stream has been consumed or should it just write the file as it's streamed so we can get the sha at the end?

PR #351, opened 2018-06-13
Allow arbitrary host/port/encryption

birdbrained:


birdbrained:
Yep that's true, its related to s3-compat.

Replacing the calls with client library is because the signature stuff in boto2 is delusional. the signature stuff in boto3 works as intended.

It is going to be synchronous as written now, but there's two options -
    1. Just wrap it in a `await loop.run_in_executor()`
    2. There's an asyncio wrapper around boto3

PR #353, opened 2018-07-18
[No-Ticket] Improve The Command `invoke test`

cslzchen:
## Ticket

No

## Purpose

Improve The Command `invoke test`

## Changes

- Add an option `--provider=` to test a specific provider only
- Add an option `--path=` to test a specific file or folder only
- Add an option `--nocov=` to disable coverage

## Side effects

No

## QA Notes

No

## Deployment Notes

No

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -59,14 +59,36 @@ def mypy(ctx):
 
 
 @task
-def test(ctx, verbose=False, types=False):
+def test(ctx, verbose=False, types=False, nocov=False, provider=None, path=None):
+    """Run full or customized tests for WaterButler.
+
+    :param ctx: the ``invoke`` context
+    :param verbose: the flag to increase verbosity
+    :param types: the flag to enable ``mypy`` test
+    :param nocov: the flag to disable coverage
+    :param provider: limit the tests to the given provider only
+    :param path: limit the tests to the given path only
+    :return: None
+    """
+
     flake(ctx)
+
     if types:
         mypy(ctx)
 
-    cmd = 'py.test --cov-report term-missing --cov waterbutler tests'
-    if verbose:
-        cmd += ' -v'
+    # `--provider=` and `--path=` are mutually exclusive options
+    assert not (provider and path)
+    if path:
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -59,14 +59,36 @@ def mypy(ctx):
...
+
     flake(ctx)
+
     if types:
         mypy(ctx)
 
-    cmd = 'py.test --cov-report term-missing --cov waterbutler tests'
-    if verbose:
-        cmd += ' -v'
+    # `--provider=` and `--path=` are mutually exclusive options
```
cslzchen: To make things simple, I make `--provider=` and `--path=` mutually exclusive options.

cslzchen: To make things simple, I make `--provider=` and `--path=` mutually exclusive options.

tasks.py
```diff
@@ -59,14 +59,36 @@ def mypy(ctx):
...
-def test(ctx, verbose=False, types=False):
+def test(ctx, verbose=False, types=False, nocov=False, provider=None, path=None):
+    """Run full or customized tests for WaterButler.
+
+    :param ctx: the ``invoke`` context
+    :param verbose: the flag to increase verbosity
+    :param types: the flag to enable ``mypy`` test
+    :param nocov: the flag to disable coverage
+    :param provider: limit the tests to the given provider only
+    :param path: limit the tests to the given path only
```
cslzchen: To be more specific, `tests/` is the base directory for `path`.

tasks.py
```diff
@@ -59,14 +59,36 @@ def mypy(ctx):
...
-    if verbose:
-        cmd += ' -v'
+    # `--provider=` and `--path=` are mutually exclusive options
+    assert not (provider and path)
+    if path:
+        path = '/{}'.format(path) if path else ''
+    elif provider:
+        path = '/providers/{}/'.format(provider) if provider else ''
+    else:
+        path = ''
```
cslzchen: For the few lines above, I reused the var `path` simplicity (normally we prefer not to overwrite the arguments.)

PR #356, opened 2018-08-14
Kill archive and parchive tasks

birdbrained:
This PR needs someone that knows the desired setup for osfstorage better than I do to look at it and make sure this is doing the right thing.

## Ticket

https://openscience.atlassian.net/browse/SVCS-702

## Purpose

We dont use them anymore, so this removes the celery tasks, and the code
inside the osfstorage provider that calls them.

## Changes

Removes the celery tasks, and the code
inside the osfstorage provider that calls them.


felliott:
Looks good!  I've remove a few more things and forced-pushed this branch.  Will merge now.  Thanks, @birdbrained!

PR #357, opened 2018-08-14
[SVCS-281] Escape response headers

birdbrained:
## Ticket

https://openscience.atlassian.net/browse/SVCS-281

## Purpose

Response headers should not have control characters in them because they
can mess up the headers returned in the response. 

## Changes

This change escapes them so they don't mess up the headers.


birdbrained:
See the PR for content disposition headers for a solution that encodes filenames as utf-8 in the header properly.

PR #358, opened 2018-08-20
[SVCS-895] Upgrade tornado from 4.3 to 5.1

ExProbitasFiducia:
## Ticket

https://openscience.atlassian.net/browse/SVCS-895

## Purpose

The latest version of tornado has updates to use the event loop properly
that we need for aiohttp context managers to work properly. The version
of tornado that we use doesn't use the asyncio event loop, but its own
event loop, so other libraries that use the event loop sometimes don't
work properly. Like aiohttp.

## Changes

Tornado to 5.1

### DIFF
--- a/waterbutler/server/app.py
+++ b/waterbutler/server/app.py
@@ -49,8 +49,6 @@ def make_app(debug):
 
 
 def serve():
-    tornado.platform.asyncio.AsyncIOMainLoop().install()
-
     app = make_app(server_settings.DEBUG)
 
     ssl_options = None
--- a/waterbutler/server/app.py
+++ b/waterbutler/server/app.py
@@ -1,8 +1,8 @@
 import os
-# import signal
+import signal
 import asyncio
 import logging
-# from functools import partial
+from functools import partial
 
 import tornado.web
 import tornado.platform.asyncio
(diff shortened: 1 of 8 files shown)


waterbutler/server/app.py
```diff
@@ -19,13 +19,15 @@
 
 
 def sig_handler(sig, frame):
-    io_loop = tornado.ioloop.IOLoop.instance()
+    io_loop = tornado.ioloop.IOLoop.current()
```
cslzchen: As discussed in person, we should use `.instance()` but please double check. See https://stackoverflow.com/questions/34554247/python-tornado-i-o-loop-current-vs-instance-method.

waterbutler/server/app.py
```diff
@@ -71,5 +71,6 @@ def serve():
     logger.info("Listening on {0}:{1}".format(server_settings.ADDRESS, server_settings.PORT))
 
     signal.signal(signal.SIGTERM, partial(sig_handler))
-    asyncio.get_event_loop().set_debug(server_settings.DEBUG)
-    asyncio.get_event_loop().run_forever()
+    tornado.ioloop.IOLoop.current().start()
```
cslzchen: I am for 👍replacing

```python
tornado.platform.asyncio.AsyncIOMainLoop().install()
asyncio.get_event_loop().run_forever()
```
with

```python
tornado.ioloop.IOLoop.current().start()
```

See https://www.tornadoweb.org/en/stable/asyncio.html where it says:

> Deprecated since version 5.0: While the code in this module is still used, it is now enabled automatically when asyncio is available, so applications should no longer need to refer to this module directly.

cslzchen: However, need to double check if we should use `.instance()` as well.

cslzchen:
Forgot to mention that I changed the target to `develop` and remove the `next` label.

cslzchen:
PR replaced by https://github.com/CenterForOpenScience/waterbutler/pull/372.

PR #359, opened 2018-08-22
Remove cute trick to deal with bufffer size

birdbrained:
## Purpose

Wrapping around a socket is cute, but it's not really the right way to
do it. Put an actual limit on the buffer size and wait for it to get
written.


birdbrained:
The test failing shouldn't matter because we're removing v0.

PR #360, opened 2018-08-23
[WIP][SVCS-892] Support cross-region moves in osfstorage.

felliott:
## Ticket

https://openscience.atlassian.net/browse/SVCS-892

## Purpose

Cross-region moves in osfstorage are wonky b/c they can't use `intra_move` (which won't copy the data) and they can't use the default `move` (which will create all-new file metadata entries on the OSF, losing guids and versions in the process).   This PR adds an osfstorage-specific `move` method which will both copy the data to the new region and update the file metadata entry on the OSF.

## Changes

* Factor out duplicate code in osfstorage's `intra_move` and `intra_copy` to make it easier to reason about.  Rewrite associated tests to improve coverage and accuracy.
* Split out the "upload-to-data-storage" and "upload-to-metadata-provider" code from `upload` so that they can be called independently of each other.
* Add an osfstorage-specific `move` method that will upload to the new region's data storage, then call `intra_move` to preserve and update the metadata entry.
* Send the file path and requested version to the OSF when requesting auth.  This will allow the OSF to send the correct settings and credentials for the region the file version is saved in.  This depends on CenterForOpenScience/osf.io#8643 to function correctly for old versions, but will not break requests on current versions or old versions in the same region as current.

## Side effects

goodness gracious yes

## QA Notes

* Need to test move and copy behavior for osfstorage, both for same-region and cross-region move/copies.  After move, file guids and version history should be preserved.  All previous versions should be downloadable.

## Deployment Notes

Shouldn't need any deployment changes on the WB-side.

### DIFF
--- a/tests/providers/osfstorage/test_provider.py
+++ b/tests/providers/osfstorage/test_provider.py
@@ -14,8 +14,10 @@ from waterbutler.providers.osfstorage.metadata import (OsfStorageFileMetadata,
                                                        OsfStorageRevisionMetadata)
 
 from tests import utils
-from tests.providers.osfstorage.fixtures import (auth, credentials, settings, provider,
-                                                 provider_and_mock, provider_and_mock2,
+from tests.providers.osfstorage.fixtures import (auth, credentials, settings,
+                                                 settings_region_one, settings_region_two,
+                                                 provider_one, provider_two,
+                                                 provider_and_mock_one, provider_and_mock_two,
                                                  file_stream, file_like, file_content,
                                                  file_lineage, file_metadata,
                                                  file_metadata_object, file_path,
@@ -24,7 +26,7 @@ from tests.providers.osfstorage.fixtures import (auth, credentials, settings, pr
                                                  revisions_metadata, revision_metadata_object,
                                                  download_response, download_path,
                                                  upload_response, upload_path, root_path,
-                                                 mock_time, provider_other, )
+                                                 mock_time, mock_inner_provider,)
 
 
 def build_signed_url_without_auth(provider, method, *segments, **params):
@@ -48,16 +50,17 @@ class TestCreateFolder:
 
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
-    async def test_create_folder(self, folder_path, provider, folder_metadata, mock_time):
+    async def test_create_folder(self, folder_path, provider_one, folder_metadata, mock_time):
 
+        provider = provider_one
         data = json.dumps(folder_metadata)
-        url, params = build_signed_url_without_auth(provider, 'POST',
+        url, params = build_signed_url_without_auth(provider_one, 'POST',
                                                     folder_path.parent.identifier,
                                                     'children', data=data)
 
         aiohttpretty.register_json_uri('POST', url, body=folder_metadata, status=201, params=params)
 
-        resp = await provider.create_folder(folder_path)
+        resp = await provider_one.create_folder(folder_path)
 
         assert isinstance(resp, OsfStorageFolderMetadata)
 
@@ -66,10 +69,10 @@ class TestDownload:
 
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
-    async def test_download_with_auth(self, provider_and_mock, download_response, download_path,
-                                      mock_time):
+    async def test_download_with_auth(self, provider_and_mock_one, download_response,
+                                      download_path, mock_time):
 
-        provider, inner_provider = provider_and_mock
+        provider, inner_provider = provider_and_mock_one
 
         uri, params = build_signed_url_with_auth(provider, 'GET', download_path.identifier,
                                                  'download', version=None, mode=None)
@@ -90,9 +93,9 @@ class TestDownload:
 
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
-    async def test_download_without_auth(self, provider_and_mock, download_response, download_path,
-                                         mock_time):
-        provider, inner_provider = provider_and_mock
+    async def test_download_without_auth(self, provider_and_mock_one, download_response,
+                                         download_path, mock_time):
+        provider, inner_provider = provider_and_mock_one
 
         provider.auth = {}
         url, params = build_signed_url_without_auth(provider, 'GET', download_path.identifier,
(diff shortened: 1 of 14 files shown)


waterbutler/providers/osfstorage/provider.py
```diff
@@ -463,3 +360,168 @@ def signed_request(self, *args, **kwargs):
...
+        return data, created
+
+    async def move(self,
+                   dest_provider: provider.BaseProvider,
+                   src_path: WaterButlerPath,
+                   dest_path: WaterButlerPath,
+                   rename: str=None,
+                   conflict: str='replace',
+                   handle_naming: bool=True) -> typing.Tuple[BaseMetadata, bool]:
+
```
birdbrained: This could use a docstring

waterbutler/providers/osfstorage/provider.py
```diff
@@ -466,3 +468,60 @@ def signed_request(self, *args, **kwargs):
...
+        if src_path.is_dir:
+            meta_data, created = await self._folder_file_op(self.move, *args, **kwargs)  # type: ignore
+            await self.delete(src_path)
+        else:
+            download_stream = await self.download(src_path)
+            if getattr(download_stream, 'name', None):
+                dest_path.rename(download_stream.name)
+
+            await dest_provider._send_to_storage_provider(download_stream, dest_path, **kwargs)
+            meta_data, created = await self.intra_move(dest_provider, src_path, dest_path)
```
cslzchen: I forgot to mention that what if the `src_path` and `dest_path` are the same? It could be an error and we should just make a metadata call instead of calling the `move`.

waterbutler/providers/osfstorage/provider.py
```diff
@@ -463,3 +360,168 @@ def signed_request(self, *args, **kwargs):
...
+
+        if src_path.is_dir:
+            meta_data, created = await self._folder_file_op(self.move, *args, **kwargs)  # type: ignore
+            await self.delete(src_path)
+        else:
+            download_stream = await self.download(src_path)
+            if getattr(download_stream, 'name', None):
+                dest_path.rename(download_stream.name)
+
+            await dest_provider._send_to_storage_provider(download_stream, dest_path, **kwargs)
```
birdbrained: Would it be possible, with bad settings or something, to get a recursive call between `_send_to_storage_provider` and `move`?

waterbutler/providers/osfstorage/provider.py
```diff
@@ -466,3 +468,60 @@ def signed_request(self, *args, **kwargs):
...
+        ):
+            raise exceptions.OverwriteSelfError(src_path)
+
+        self.provider_metrics.add('move.can_intra_move', False)
+        if self.can_intra_move(dest_provider, src_path):
+            self.provider_metrics.add('move.can_intra_move', True)
+            return await self.intra_move(*args)
+
+        if src_path.is_dir:
+            meta_data, created = await self._folder_file_op(self.move, *args, **kwargs)  # type: ignore
```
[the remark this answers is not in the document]
felliott: This code is c-n-p'd from [`waterbutler.core.provider.BaseProvider.move`](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/core/provider.py#L250), which also uses `meta_data`.  I'd like keep them the same for now, to make the two methods more diffable.  But at some point it would be good to refactor `BaseProvider.move` (and `copy`) to move some of their logic into methods we can call here.  When we do that, `metadata`++!

waterbutler/providers/osfstorage/provider.py
```diff
@@ -466,3 +468,60 @@ def signed_request(self, *args, **kwargs):
...
+                   dest_provider: provider.BaseProvider,
+                   src_path: WaterButlerPath,
+                   dest_path: WaterButlerPath,
+                   rename: str=None,
+                   conflict: str='replace',
+                   handle_naming: bool=True) -> typing.Tuple[BaseMetadata, bool]:
+
+        # when moving to non-osfstorage, default move is fine
+        if dest_provider.NAME != 'osfstorage':
+            super().move(dest_provider, src_path, dest_path, rename=rename, conflict=conflict,
```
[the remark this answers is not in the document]
felliott: Great catch!  You're absolutely right.

waterbutler/providers/osfstorage/provider.py
felliott: Nope.  `_send_to_storage_provider` only calls `upload`, `validate_*_path`, `metadata`, and `delete` on the storage provider.

waterbutler/providers/osfstorage/provider.py
felliott: That's covered by the `OverwriteSelfError` stanza around L505.

waterbutler/providers/osfstorage/provider.py
birdbrained: https://github.com/CenterForOpenScience/waterbutler/pull/360/files/5a33de669e0190238438c1996cc9c9cbf5005472#diff-927878a162eb1a1110dfc88d4cf2373dR431

felliott: Derp. My bad, but still no.  That `provider` is a data storage provider, which the OSF is not.

tests/providers/osfstorage/test_provider.py
```diff
@@ -426,35 +431,52 @@ class TestIntraMove:
...
-
-        src_path = WaterButlerPath('/file', _ids=['RootId', 'fileId'], folder=False)
-        dest_path = WaterButlerPath('/folder1/', folder=True)
+        src_path = WaterButlerPath('/test_file', _ids=['RootId', 'fileId'], folder=False)
+        dest_path = WaterButlerPath('/folder1/test_file',
+                                    _ids=['RootId', 'folder1Id', 'doomedFile'],
+                                    folder=False)
 
-        data = json.dumps({
+        data=json.dumps({
```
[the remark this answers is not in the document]
felliott: Yeah, looks the ghost of c-n-p past has come back to haunt us.  Will fix.

waterbutler/providers/osfstorage/provider.py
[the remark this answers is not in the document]
cslzchen: Ah, I missed this. Let's keep them the same for better diff and better commit.

waterbutler/auth/osf/handler.py
```diff
@@ -136,7 +137,9 @@ def build_payload(self, bundle, view_only=None, cookie=None):
             self.build_payload({
                 'nid': resource,
                 'provider': provider,
-                'action': osf_action
+                'action': osf_action,
+                'path': path,
+                'version': version,
```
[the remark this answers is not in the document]
cslzchen: Yep. `version` now has a foreign key `region`. See here: https://github.com/CenterForOpenScience/osf.io/pull/8643/files#diff-511d556f4513d9e41850aa51126eceabR296

waterbutler/providers/osfstorage/provider.py
birdbrained: would it be ok to do something like `assert provider != self` inside `_send_to_storage_provider`?

PR #361, opened 2018-08-30
[SVCS-939] Upgrade WB base python to 3.6

ExProbitasFiducia:
This PR is pointing at the next branch.

Can't upgrade to 3.7 yet because our old version of aiohttp is blocking
it. async becomes a reserved word, but the old version of aiohttp uses
that.

This pr will require another pr on osf.io to update `docker-compose`

## Ticket

https://openscience.atlassian.net/browse/SVCS-939

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,72 @@
+version: "3"
+
+services:
+  wb_requirements:
+    #image: quay.io/centerforopenscience/wb:develop
+    #image: quay.io/centerforopenscience/wb:next
+    image: wb:local
+    command:
+      - /bin/bash
+      - -c
+      - invoke install --develop &&
+        (python -m compileall /usr/local/lib/python3.6 || true) &&
+        rm -Rf /python3.6/* &&
+        cp -Rf -p /usr/local/lib/python3.6 /
+    restart: 'no'
+    volumes:
+      - wb_requirements_vol:/python3.6
+      - wb_requirements_local_bin_vol:/usr/local/bin
+      - wb-sync:/code:nocopy
+
+  wb:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,72 @@
+version: "3"
```
cslzchen: * Personally, I still prefer this to be a local file for developers who need it.
* In addition, this PR shouldn't include this change since it is not related with Python upgrade.

requirements.txt
```diff
@@ -11,7 +11,7 @@ pyjwe==1.0.0
 pyjwt==1.4.0
 python-dateutil==2.5.3
 pytz==2017.2
-raven==5.27.0
+raven==6.9.0
```
cslzchen: Is the `raven` upgrade blocking `python` upgrade? I can't find docs supporting this claim. I think we can have a separate PR for upgrading `raven.`

waterbutler/core/utils.py
```diff
@@ -55,7 +55,7 @@ def as_task(func):
 
     @functools.wraps(func)
     def wrapped(*args, **kwargs):
-        return asyncio.async(func(*args, **kwargs))
+        return asyncio.ensure_future(func(*args, **kwargs))
```
cslzchen: 👍 fyi, here is the ref: https://docs.python.org/3.6/library/asyncio-task.html#asyncio.ensure_future

> `asyncio.async(coro_or_future, *, loop=None)`
A deprecated alias to ensure_future().
Deprecated since version 3.4.4.

cslzchen:
In addition, please create the a ticket or link one. Thanks.

felliott:
Done in d03a84c7046baed176. Closing.

PR #347, opened 2018-05-31
Ft/dropbox chunked uploads

birdbrained:

PR #364, opened 2018-09-10
[HotFix] [SVCS-913] Fix Conflict Handling For Dropbox Contiguous Upload

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-913

## Purpose

A bug was introduced during dropbox chunked upload where the argument `conflict` is not passed along from `_upload()`to `_contiguous_upload()` and `_chunked_upload()`. This bug led to all requests made to the dropbox upload API missing the `mode` parameter in the required header `Dropbox-API-Arg`.

See https://github.com/CenterForOpenScience/waterbutler/pull/347/files#diff-c3a345e63f19827819cde3a30cae89d6L262 and 
https://github.com/CenterForOpenScience/waterbutler/pull/347/files#diff-c3a345e63f19827819cde3a30cae89d6L255

## Changes

* `_contiguous_upload()` now takes in the `conflict` argument from `_upload()`
* `_chunked_upload()` now takes in the `conflict` argument from `_upload()` and passed along to the `_complete_session()`
* Move `path_arg` from `_upload()` to `_contiguous_upload()`
* Update existing `upload_args` in `_chunked_upload()`
* Update the DocStrings

## Side effects

No

## QA Notes

Should this be Dev QA? @felliott 

## Deployment Notes

No

### DIFF
--- a/waterbutler/providers/dropbox/provider.py
+++ b/waterbutler/providers/dropbox/provider.py
@@ -260,31 +260,39 @@ class DropboxProvider(provider.BaseProvider):
         multipart upload endpoints will be used.
         """
         path, exists = await self.handle_name_conflict(path, conflict=conflict)
-        path_arg = {"path": path.full_path}
-        if conflict == 'replace':
-            path_arg['mode'] = 'overwrite'
 
         if stream.size > self.CONTIGUOUS_UPLOAD_SIZE_LIMIT:
-            data = await self._chunked_upload(stream, path)
+            data = await self._chunked_upload(stream, path, conflict=conflict)
         else:
-            data = await self._contiguous_upload(stream, path)
+            data = await self._contiguous_upload(stream, path, conflict=conflict)
 
         return DropboxFileMetadata(data, self.folder), not exists
 
-    async def _contiguous_upload(self, stream: streams.BaseStream, path: WaterButlerPath) -> dict:
+    async def _contiguous_upload(self,
+                                 stream: streams.BaseStream,
+                                 path: WaterButlerPath,
+                                 conflict: str='replace') -> dict:
         """Upload file in a single request.
 
         API Docs: https://www.dropbox.com/developers/documentation/http/documentation#files-upload
 
+        :param stream: the stream to upload
+        :param path: the WB path of the file
+        :param conflict: whether to replace upon conflict
         :rtype: `dict`
-        :return: A `dict` of metadata about the file just uploaded.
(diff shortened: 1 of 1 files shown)


waterbutler/providers/dropbox/provider.py
```diff
@@ -260,31 +260,39 @@ def default_headers(self) -> dict:
...
+        :param stream: the stream to upload
+        :param path: the WB path of the file
+        :param conflict: whether to replace upon conflict
         :rtype: `dict`
-        :return: A `dict` of metadata about the file just uploaded.
+        :return: A dictionary of the metadata about the file just uploaded
         """
+
+        path_arg = {"path": path.full_path}
+        if conflict == 'replace':
```
cslzchen: For me the `conflict` is more like an boolean flag to allow/disallow overwrite than an option among many choices. However, it was written this way so I decide to preserve it.

cslzchen:
@felliott 

- [x] Apply the fix to chunked upload as well
- [x] Locally tested that chunked upload replacing existing file works as expected
- [x] Created a [ticket](https://openscience.atlassian.net/browse/SVCS-913) for better documentation.

waterbutler/providers/dropbox/provider.py
```diff
@@ -408,22 +416,28 @@ def default_headers(self) -> dict:
...
-        :return: A `dict` of metadata about the file just uploaded.
+        :return: A dictionary of the metadata about the file just uploaded
         """
 
         upload_args = {
             'cursor': {'session_id': session_id, 'offset': stream.size, },
             'commit': {"path": path.full_path, },
         }
+        if conflict == 'replace':
+            upload_args['commit']['mode'] = 'overwrite'
```
cslzchen: Fyi, parameters for [finishing upload session](https://www.dropbox.com/developers/documentation/http/documentation#files-upload_session-finish). Please note that `mode` is a key in `commit`.

```json
{
    "cursor": {
        "session_id": "1234faaf0678bcde",
        "offset": 0
    },
    "commit": {
        "path": "/Homework/math/Matrices.txt",
        "mode": "add",
        "autorename": true,
        "mute": false,
        "strict_conflict": false
    }
}
```

waterbutler/providers/dropbox/provider.py
felliott: :+1:  The other options will be handled in the BaseProvider.

PR #365, opened 2018-09-10
[WIP] Ft/upgrade aiohttp

birdbrained:

### DIFF
--- a/waterbutler/auth/osf/handler.py
+++ b/waterbutler/auth/osf/handler.py
@@ -2,12 +2,19 @@ import datetime
 
 import jwe
 import jwt
-import aiohttp
+import json
+from tornado.httpclient import (
+    AsyncHTTPClient,
+    HTTPRequest
+)
+from tornado.httputil import url_concat
 
 from waterbutler.core import exceptions
 from waterbutler.auth.osf import settings
-from waterbutler.core.auth import (BaseAuthHandler,
-                                    AuthType)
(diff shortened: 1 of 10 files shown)


waterbutler/auth/osf/handler.py
```diff
@@ -22,50 +30,60 @@ class OsfAuthHandler(BaseAuthHandler):
         'delete': 'delete',
     }
 
-    def build_payload(self, bundle, view_only=None, cookie=None):
-        query_params = {}
+    @property
+    def client(self):
```
cslzchen: From the diff, this looks like a PR for replacing `aiohttp` with `tornado` instead of upgrading `aiohttp`.

waterbutler/auth/osf/handler.py
```diff
@@ -102,45 +125,84 @@ def build_payload(self, bundle, view_only=None, cookie=None):
...
+                        auth_type=auth_type
+                    )
+                ),
+                method='GET',
+                headers=self.build_auth_headers(request),
+            )
+            import pdb
+            pdb.set_trace()
+
+            response = await self.client.fetch(auth_request)
```
cslzchen: `.fetch` works for non-streaming requests such as simple file metadata or the auth here. However, it won't work for streaming even with `streaming_callback` as we discussed. The response needs to be `StreamReader` compatible.

PR #366, opened 2018-09-11
[SVCS-894] Cache File During Contiguous Upload For Dropbox

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/SVCS-894

## Purpose

Dropbox's rate limit feature may cause uploading/copying/moving many files to fail with "too_many_requests" & "too_many_write_operations". The former is for too many requests literally and the former is for namespace lock contentions. Unlike GitHub, Dropbox doesn't provide detailed information on how they rate limit requests but simply asks clients to retry according to the "Retry-After" header in the 429 response.

The retry doesn't work in a straightforward way for upload (inter copy and move use upload internally) since the stream will have been consumed when the request is finished. The solution is to cache the stream locally into a temporary file and stream from the file for both the initial request and following 429 retries.

## Changes

TBD

## Side effects

TBD

## QA Notes

TBD

## Deployment Notes

TBD

### DIFF
--- a/waterbutler/providers/dropbox/provider.py
+++ b/waterbutler/providers/dropbox/provider.py
@@ -61,6 +61,7 @@ class DropboxProvider(provider.BaseProvider):
     BASE_URL = pd_settings.BASE_URL
     CONTIGUOUS_UPLOAD_SIZE_LIMIT = pd_settings.CONTIGUOUS_UPLOAD_SIZE_LIMIT
     CHUNK_SIZE = pd_settings.CHUNK_SIZE
+    MAX_429_RETRIES = pd_settings.MAX_429_RETRIES
 
     def __init__(self, auth, credentials, settings):
         super().__init__(auth, credentials, settings)
@@ -304,7 +305,7 @@ class DropboxProvider(provider.BaseProvider):
             chunk = await stream.read()
 
         rate_limit_retry = 0
-        while rate_limit_retry < 2:
+        while rate_limit_retry < self.MAX_429_RETRIES:
             file_stream = streams.FileStreamReader(file_cache)
             resp = await self.make_request(
                 'POST',
--- a/waterbutler/providers/dropbox/provider.py
+++ b/waterbutler/providers/dropbox/provider.py
@@ -1,6 +1,7 @@
 import json
 import typing
 import logging
+import tempfile
 from http import HTTPStatus
 
 from waterbutler.core import provider, streams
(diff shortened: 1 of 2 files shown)


waterbutler/providers/dropbox/provider.py
```diff
@@ -276,26 +277,60 @@ def default_headers(self) -> dict:
...
+                expects=(200, 409, 429, ),
+                throws=core_exceptions.UploadError,
+            )
+            data = await resp.json()
+            if resp.status == 429:
+                rate_limit_retry += 1
+                logger.debug('Retry {} for {}'.format(rate_limit_retry, str(path)))
+                continue
+            elif resp.status == 409:
+                file_cache.close()
```
cslzchen: I try to close the file_cache as soon as possible. Not sure if this is necessary though. Several repetitive occurrences in the `if ... else ...` block.

waterbutler/providers/dropbox/provider.py
```diff
@@ -276,26 +277,60 @@ def default_headers(self) -> dict:
...
+
+        rate_limit_retry = 0
+        while rate_limit_retry < 2:
+            file_stream = streams.FileStreamReader(file_cache)
+            resp = await self.make_request(
+                'POST',
+                self._build_content_url('files', 'upload'),
+                headers={
+                    'Content-Type': 'application/octet-stream',
+                    'Dropbox-API-Arg': json.dumps({'path': path.full_path}),
```
cslzchen: Here is where there will be conflicts with https://github.com/CenterForOpenScience/waterbutler/pull/364, in case you want to fix the conflicts yourself.

cslzchen:
@felliott 

- [x] Rebased to `develop`
- [x] Make max retries a provider setting

waterbutler/providers/dropbox/provider.py
```diff
@@ -276,26 +278,60 @@ def default_headers(self) -> dict:
...
+        different ones.  This makes balancing the requests less effective.
+
+        References: https://www.dropbox.com/developers/reference/data-ingress-guide
+
+        Quirks: Retry Upload Load Requests
+
+        When an upload request finishes, the stream will have been consumed.  In order to retry such a request, WB needs
+        to cache a temporary local file from the incoming stream so both the initial request and retries can stream from
+        this file.
+
```
birdbrained: What a lovely docstring :)

waterbutler/providers/dropbox/provider.py
```diff
@@ -276,26 +278,60 @@ def default_headers(self) -> dict:
...
 
-        data = await resp.json()
-        if resp.status == 409:
-            self.dropbox_conflict_error_handler(data, path.path)
-        return data
+        file_cache = tempfile.TemporaryFile()
+        chunk = await stream.read()
+        while chunk:
+            file_cache.write(chunk)
+            chunk = await stream.read()
```
birdbrained: I think it would be better to get a new stream to consume in the case our upload fails, Maybe this logic is better in the handler/task?

PR #370, opened 2019-01-18
[PLAT-1310][PLAT-709] Add more client request info to Impact metrics / Wiki embed metrics

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/PLAT-1310
https://openscience.atlassian.net/browse/PLAT-709

## Purpose

Pass useful useful info for impact metrics back to the waterbutler auth callback so we can count them perfectly.

## Changes

- adds extra items to payload in auth handler and change test.

## Side effects

None that I know of.

## QA Notes

Not user-facing

## Deployment Notes

Should be deployed before corresponding OSF PR. https://github.com/CenterForOpenScience/osf.io/pull/8918

### DIFF
--- a/waterbutler/auth/osf/handler.py
+++ b/waterbutler/auth/osf/handler.py
@@ -144,6 +144,12 @@ class OsfAuthHandler(BaseAuthHandler):
                 'action': osf_action,
                 'path': path,
                 'version': version,
+                'metrics': {
+                    'referrer': request.headers.get('Referer'),
+                    'user_agent': request.headers.get('User-Agent'),
+                    'origin': request.headers.get('Origin'),
+                    'uri': request.uri,
+                }
             }, cookie=cookie, view_only=view_only),
             headers,
             dict(request.cookies)
--- a/tests/auth/osf/test_handler.py
+++ b/tests/auth/osf/test_handler.py
@@ -78,17 +78,29 @@ class TestOsfAuthHandler(ServerTestCase):
             await self.handler.get(resource, provider, self.request, action=action, auth_type=auth_type)
             if auth_type is AuthType.SOURCE:
                 self.handler.build_payload.assert_called_with({
-                    'nid': resource,
-                    'provider': provider,
+                    'nid': 'test',
+                    'provider': 'test',
                     'action': 'download',
                     'path': '',
                     'version': None,
+                    'metrics': {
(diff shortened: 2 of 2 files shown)


waterbutler/auth/osf/handler.py
```diff
@@ -142,8 +142,11 @@ def build_payload(self, bundle, view_only=None, cookie=None):
                 'nid': resource,
                 'provider': provider,
                 'action': osf_action,
-                'path': path,
-                'version': version,
+                'refer': request.headers.get('Refer'),
```
sloria: Shouldn't this be `Referer`, or is this a special WB thing?

waterbutler/auth/osf/handler.py
```diff
@@ -142,8 +143,11 @@ def build_payload(self, bundle, view_only=None, cookie=None):
                 'nid': resource,
                 'provider': provider,
                 'action': osf_action,
-                'path': path,
-                'version': version,
```
felliott: Are we not using `path` and `version` anymore?  I thought that was needed for version tracking and x-region move/copy?

waterbutler/auth/osf/handler.py
```diff
@@ -142,8 +143,11 @@ def build_payload(self, bundle, view_only=None, cookie=None):
...
+                'referer': request.headers.get('Referer'),
```
felliott: Please use `referrer` as the key name. We don't follow the header misspelling.

waterbutler/auth/osf/handler.py
```diff
@@ -142,8 +143,11 @@ def build_payload(self, bundle, view_only=None, cookie=None):
...
+                'user_agent': request.headers.get('User-Agent'),
+                'origin': request.headers.get('Origin'),
+                'uri': request.uri,
+                'query_args': parse_qs(request.query, keep_blank_values=True)
```
felliott: Will the OSF-side use all of these, or is this a might-gonna-need-it scenario?  If we have an immediate need, then :+1:.  Otherwise, I'd prefer to send only the things we absolutely need.

sloria: Mostly we just need to be able to figure out the origin of the requests (i.e. figure out if requests are coming from Google Scholar). If we can get these from Referer, then that should suffice.

PR #378, opened 2019-06-17
[ENG-380] Fix the web view link for figshare and refactor fileset to dataset

cslzchen:
## Ticket

WB: https://openscience.atlassian.net/browse/ENG-380
By-product: https://openscience.atlassian.net/browse/ENG-649
Future-fix: https://openscience.atlassian.net/browse/ENG-996

### Main

Fix the broken web view link for figshare files. (ENG-380)

* Previously, WB checks whether a figshare article is public or not simply by checking if the `url` field in the JSON response for the article details contains (starts with) the `PRIVATE_IDENTIFIER`: "https://api.figshare.com/v2/account/". This no longer works now since figshare uses the same format for this field for all articles. Fortunately, the JSON response now provides the `is_public` field which is more intuitive and more convenient.

* In addition, the public **HTML** URL we build no longer works either. The aforementioned JSON response also provides two new fields: `url_public_html` and `url_private_html`, which are exactly what we need.

### By-product

* Similarly to figshare file metadata, checking whether a figshare path is public or not is fixed. Please don't confuse `FigsharePath.is_public` with `FigshareFileMetadata.is_public`. In addition, `FigshareBaseProvider.build_url()` now builds private **API** URL only. The main reasons are public **API** URL 1) is broken as it is written and 2) should've been used for collections only. (ENG-380)
* Refactored `Fileset` to `Dataset`. (ENG-649)

### No Fix

* All article types now become folder-like containers for figshare.
* Public API URL for collection containers.
* Comments and DocStr is not fully rewritten due to this is a partial fix.

Please refer to this [JIRA page](https://openscience.atlassian.net/wiki/spaces/SVCS/pages/817037313) for a comprehensive solution for article types and public API endpoints and this [SVCS task](https://openscience.atlassian.net/browse/ENG-996) for implementation for a full fix.

### Figshare File Metadata

fighsare article details request now returns several useful information including: `is_public`, `url_private_html` and `url_public_html`.

* `FigshareFileMetadata.is_public` is now set by the `is_public` key.
* `FigshareFileMetadata.web_view` is now set by either the `url_public_html` or `url_private_html` key, depending on the value of `FigshareFileMetadata.is_public`.

### By-product: Path Validation for Figshare Project Provider

Unlike the article detail request, the project article listing request only returns limited metadata for each article. While it does not provide the `is_public` field, WB can use the `published_date` field instead.

* In both V0 and V1 path validation, `FigsharePath.is_public` is now set by whether the `published_date` field for the current article is a date or `null`.

### By-product: Building URL

For more information on figshare project, collection, article and file, see the aforementioned [JIRA page](https://openscience.atlassian.net/wiki/spaces/SVCS/pages/817037313).

* The fix of `FigsharePath.is_public` reveals another issue that the public API URL WB builds for public project-article and project-article-file no longer works. Actually, figshare doesn't provide such a public API endpoint for them at all. To fix this, simply let the `BaseFigshareProvider.build_url()` build private API URL for project-article and project-article-file all the time, even if the `is_public` argument is `True`.

* Please note that the above fix is a correction instead of a fall-back. `project-article` and `project-article-file` should be accessed with the private API with the credentials of the figshare-OSF auth user. The public API URL was designed for collections only but never tested and enabled. The previous design to make the `FigshareProjectProvider` handle both projects and collections may no longer be feasible due to API differences.

* Please also note that article and article-file are independent of the project they are in. Some may not even belong to any project. figshare provides both public and private API URL at the article level (not the project level. However, we cannot use this public API URL when fetching metadata for an article with a project in the context since the response lacks necessary project information. The article level public API URL should only be used for article and article-file that belongs to collections.

### By-product: `Fileset` -> `Dataset`

Figshare somehow deprecates `Fileset` without any announcement and internally converts `Fileset` to `Dataset`. For example, previous `Fileset` article now becomes `Dataset` on the figshare website. Using API to create a `Fileset` ends up creating a `Dataset`. Fixing this issue makes figshare have folders again so we can test both folder and files. For the OSF side fix, see [OSF-PR#9078](https://github.com/CenterForOpenScience/osf.io/pull/9078).

* When detecting article types (folder or file), we now detects both `Dataset` and `Fileset`.
* When creating folders, we create `Dataset` instead of `Fileset`.
* In addition, added `Dataset` to valid container and folder list.

### Tests

I recommend you review the tests update by following the commits. Personally, I prefer having both a complex refactor for test_metadata` and `test_provider.TestProjectMetadata` and a fixture tweak for the rest. 

- [ ] However, if you prefer the simple fix for all, I will do that.

* Tests in `test_metadata` and `test_provider.TestProjectMetadata` have been fully rewritten and improved using newly generated fixtures. See https://github.com/CenterForOpenScience/waterbutler/pull/378/commits/1d5b97f2a2aee40785daa9525dda5fe08f7db9ba and https://github.com/CenterForOpenScience/waterbutler/pull/378/commits/f97f6de8051e9936ff8b02ada1d777c785e1b083.

* The rest of the tests in the `test_provider` are fixed by tweaking existing fixtures. See https://github.com/CenterForOpenScience/waterbutler/pull/378/commits/896973c9f1cdfb52c73a330d9e537a48776b9aff and https://github.com/CenterForOpenScience/waterbutler/pull/378/commits/2b93c73cf76bd40a1ecf04dde40d6787b9cdfb6a.

## Side effects

No

## QA Notes

First, you should expect to see a **working** figshare provider as you did before, where **working** is WB-wise not figshare-wise, that is to say, previous folders remain to be folders and previous files remain to be files.

As for the web view link, please test that the link works for the following case. Please note that the public and private has nothing to do with whether the OSF projects is public or not.

* When a figshare project is loaded as root.
  * Public file (the one file in a public article that are neither `Fileset` nor `Dataset`)
  * Private file (the one file in a private article  that are neither `Fileset` nor `Dataset`)
  * Public file in public folder (one of the file in a public article that is either `Fileset` or `Dataset`)
  * Private file in private folder (one of the file in a private article that is either `Fileset` or `Dataset`)
* When a figshare article (of folder type) is loaded as root, which can't be tested until the OSF side [OSF-PR#9078](https://github.com/CenterForOpenScience/osf.io/pull/9078) is merged.
  * Public file (the one file in a public article that are neither `Fileset` nor `Dataset`)
  * Private file (the one file in a private article  that are neither `Fileset` nor `Dataset`)

## Deployment Notes

This PR can be merged and deployed independent of the OSF side [OSF-PR#9078](https://github.com/CenterForOpenScience/osf.io/pull/9078) but I would recommend doing them together so QA can test both project root and article root.

### DIFF
--- a/tests/auth/osf/test_handler.py
+++ b/tests/auth/osf/test_handler.py
@@ -104,3 +104,60 @@ class TestOsfAuthHandler(ServerTestCase):
                         'user_agent': None
                     }
                 }, cookie=None, view_only=None)
+
+
+class TestActionMapping:
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize('method, action, auth_type, headers, query_args, expected', [
+        ['post', 'copy',   AuthType.SOURCE,      None, None, 'download'],
+        ['post', 'copy',   AuthType.DESTINATION, None, None, 'upload'],
+        ['post', 'move',   AuthType.SOURCE,      None, None, 'upload'],  # TODO: really?
+        ['post', 'move',   AuthType.DESTINATION, None, None, 'upload'],
+        ['post', 'rename', AuthType.SOURCE,      None, None, 'upload'],  # TODO: really?
+        ['post', 'rename', AuthType.DESTINATION, None, None, 'upload'],
+
+        ['head', None, None, {settings.MFR_ACTION_HEADER: 'render'}, None, 'render'],
+        ['head', None, None, {settings.MFR_ACTION_HEADER: 'export'}, None, 'export'],
+
+        ['get',    None, None, None, None,             'download'],
+        ['put',    None, None, None, None,             'upload'],
+        ['delete', None, None, None, None,             'delete'],
+        ['get',    None, None, None, {'meta': 1},      'metadata'],
+        ['get',    None, None, None, {'revisions': 1}, 'revisions'],
+    ])
+    async def test_action_type(self, method, action, auth_type, headers, query_args, expected):
+
+        handler = OsfAuthHandler()
+
+        request = mock.Mock()
+        request.method = method
+        request.headers = headers if headers is not None else {}
+        request.query_arguments = query_args if query_args is not None else {}
+        request.cookies = {}
+
+        kwargs = {}
+        if action is not None:
+            kwargs['action'] = action
+        if auth_type is not None:
+            kwargs['auth_type'] = auth_type
+
+        handler.build_payload = mock.Mock()
+        handler.make_request = utils.MockCoroutine(
+            return_value={'auth': {}, 'callback_url': 'dummy'}
+        )
+
+        await handler.get('test', 'test', request, **kwargs)
+
+        handler.build_payload.asssert_called_once()
+        args, _ = handler.build_payload.call_args
+        assert args[0]['action'] == expected
+
+    @pytest.mark.asyncio
+    async def test_unhandled_mfr_action(self):
+        handler = OsfAuthHandler()
+        request = mock.Mock()
+        request.method = 'head'
+        request.headers = {settings.MFR_ACTION_HEADER: 'bad-action'}
+        with pytest.raises(UnsupportedActionError):
+            await handler.get('test', 'test', request)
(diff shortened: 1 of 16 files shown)


felliott:
This has been merged into release/yellowstone.  Closing.

PR #379, opened 2019-07-31
[ENG-748] Fix Incorrect Parent Folder For Box Chunked Upload

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/ENG-748

### Main

Chunked upload fails to upload the file to the intended folder. Instead, the file ends up in the root folder. During chunked upload session creation, the wrong folder was provided as the parent in the data payload. It should be `path.parent.identifier` instead of the `self.folder`. Here is the [correct one](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/box/provider.py#L504) with contiguous upload and here is the [wrong one](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/providers/box/provider.py#L596) with chunked upload.

- [x] Unit tests

### By-product

Correctly handle the mutually exclusiveness between file ID and parent folder ID during chunked upload session creation. During the creation, WB should EITHER provide the `path.identifier` (file ID) in the URL if the file already exists OR provide the `path.parent.identifier` (parent folder ID) in the data payload. In addition, providing both will get [a 400 with error `multiple_destinations`](https://developer.box.com/reference#section-error-response-body). Moreover, providing the parent folder ID when the file exists will get [a 409 with error `item_name_in_use`](https://developer.box.com/docs/error-codes#section-409-conflict).  Finally, chunked upload never asks user to confirm file conflicts and just overwrites.

- [x] Unit tests

### Out-of-scope

* [ENG-749](https://openscience.atlassian.net/browse/ENG-749)
  * When session creation returns status other than the expected 201, the error is not delivered to OSF files page immediately. This cause the OSF files page upload to hang either forever or  for a long time until a "unable to reach the provider" error is shown to the user.
* [ENG-750](https://openscience.atlassian.net/browse/ENG-750)
  * Box does not ask for user to confirm file overwrite during both normal and chunked upload.

### Main

Use `path.parent.identifier` instead of the `self.folder`.

### By-product

Only add `folder_id` to payload data when `path.identifier` is `None`.

## Side effects

No

## QA Notes

Please refer to the **QA Notes** section in the [ticket](https://openscience.atlassian.net/browse/ENG-748).

## Deployment Notes

No

### DIFF
--- a/tests/providers/box/test_provider.py
+++ b/tests/providers/box/test_provider.py
@@ -395,51 +395,114 @@ class TestUpload:
 
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
-    async def test_create_upload_session_new_file(self, provider, root_provider_fixtures,
-                                                  file_stream):
-        """Check that the chunked upload session creation makes a request to the correct url when
-        creating a new file.
+    async def test_create_upload_session_new_file(self, provider, file_stream):
+        """Check that the chunked upload session creation is sending the proper data payload to
+        the appropriate URL when creating a new file in the root of the project.
         """
 
         path = WaterButlerPath('/newfile', _ids=(provider.folder, None))
         session_url = provider._build_upload_url('files', 'upload_sessions')
 
-        create_session_metadata = root_provider_fixtures['create_session_metadata']
         aiohttpretty.register_json_uri(
             'POST',
             session_url,
             status=201,
-            body=create_session_metadata
+            body={'dummy': 'data'},
         )
 
-        session_data = await provider._create_chunked_upload_session(path, file_stream)
-
-        assert root_provider_fixtures['create_session_metadata'] == session_data
-        assert aiohttpretty.has_call(method='POST', uri=session_url)
+        await provider._create_chunked_upload_session(path, file_stream)
+        assert aiohttpretty.has_call(
+            method='POST',
+            uri=session_url,
+            data=json.dumps({
+                'folder_id': provider.folder,
+                'file_name': 'newfile',
+                'file_size': 38,
+            }, sort_keys=True),
+        )
 
     @pytest.mark.asyncio
     @pytest.mark.aiohttpretty
-    async def test_create_upload_session_existing_file(self, provider, root_provider_fixtures,
-                                                       file_stream):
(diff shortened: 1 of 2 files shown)


waterbutler/providers/box/provider.py
```diff
@@ -584,19 +584,27 @@ def _build_upload_url(self, *segments, **query):
...
 
         async with self.request(
             'POST',
             self._build_upload_url(*segments),
-            data=json.dumps({
-                'folder_id': self.folder,
-                'file_size': stream.size,
-                'file_name': path.name
-            }),
+            data=json.dumps(data, sort_keys=True),
```
cslzchen: As mentioned in the commit message, the `sort_keys=True` is for testing purpose only. Most of the time, we don't want to change functional code just for testing. However, I can't find any smarter ways to solve this payload comparison issue. The performance hit should be minimal.

tests/providers/box/test_provider.py
```diff
@@ -395,51 +395,116 @@ class TestUpload:
...
+        session_url = 'https://upload.box.com/api/2.0/files/2345/upload_sessions'
+        aiohttpretty.register_json_uri(
+            'POST',
+            session_url,
+            status=201,
+            body={'dummy': 'data'},
         )
 
-        session_data = await provider._create_chunked_upload_session(path, file_stream)
+        _ = await provider._create_chunked_upload_session(path, file_stream)
```
cslzchen: 👍 for removing unused local variable.

One question, why not just `await` without assignment? (A total of 4 occurrences in tests.)

```python
await provider._create_chunked_upload_session(path, file_stream)
```

felliott: No reason, just wasn't thinking.  I will take that out!

felliott:
This has been merged as a hotfix into master, but GH didn't pick up on that when I retargeted the PR to master.  I tried switching to develop and back to master, but it won't let me since the commits are already in master.  Confusing.  Anyway, this has been merged!

felliott:
Gonna try re-opening and reclosing to trigger a travis build.

felliott:
Nope!  That didn't work at all!

PR #380, opened 2019-08-05
[ENG-774] A Naive Solution for Waterbutler API Rate-limiting with Redis

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/ENG-774

### Related Tickets

The parent EPIC ticket: https://openscience.atlassian.net/browse/ENG-451
The previous Analysis / POC ticket: https://openscience.atlassian.net/browse/ENG-468
The next ticket: https://openscience.atlassian.net/browse/ENG-775

## Purpose

Add an naive solution to rate-limit [Waterbutler API V1](https://waterbutler.readthedocs.io/en/latest/api.html#v1-api) using [Redis](https://redis.io/).

Here is the OSF side support for local development: [OSF-PR#9199](https://github.com/CenterForOpenScience/osf.io/pull/9119).

DevOps tasks are out of scope for now.

### Preparation

* Rate-limiting algorithm: fixed Window.
  * Here is an [article]( https://konghq.com/blog/how-to-design-a-scalable-rate-limiting-algorithm/) that explains different rate-limiting algorithms as well as their pros/cons.
  * This same algorithm is used by GitHub, as mentioned in its [API Overview -> Rate Limiting](https://developer.github.com/v3/#rate-limiting).

* Maintaining state: using a Redis server
  * [Redis](https://redis.io) is an ideal choice for rate-limiting. It's fast (in-memory) and simple to use as a dictionary. Here are the commands we use: [INCR](https://redis.io/commands/incr), [EXPIRE](https://redis.io/commands/expire) and [TTL](https://redis.io/commands/ttl). Here is how [a simple rate-limiter](https://redis.io/commands/incr#pattern-rate-limiter-2) is done using Redis.
  * Server-side:  run it using `docker-compose` as configured in [OSF-PR#9199](https://github.com/CenterForOpenScience/osf.io/pull/9119).
  * Client-side: use [redis-py](https://pypi.org/project/redis/). Here is the [API Docs](https://redis-py.readthedocs.io/en/stable/) and here is [an example](https://redislabs.com/lp/python-redis/) on how to use it.
  
### What and How To Rate Limit

* Window size is 1 hour and the limit is 3600 requests
* Rate-limit applies to requests based on token auth, basic auth and no auth
* Rate-limit does not apply to cookie auth requests (which are sent by the OSF API server)
* Redis keys (token, cookie and base64 encoded username/password and IP address) are hashed for better security.

### What's not included in this naive solution

* What will WB do if Redis is down or Redis call returns an exception? 
  * Naive: WB catches it and returns a 503
* Applies stricter limits to no-auth and less strict ones to token and basic auth
  * Naive: same limit for all
* Return `X-WB-RL` headers for every request so benign API users can back-off
  * Naive: only when a request hits the limit
* Remove INFO level log
  * Naive: INFO level log are kept for convenience

## Side effects

Please refer to *Changes*

## Dev / QA Notes

* `assert needQa is False`
* The flag is turned off by default to make Travis happy

## Deployment Notes

* Merge into a feature branch instead of develop or next release 
* This feature should not go to production


felliott:
LGTM!  Added some minor changes / docs / refactors, but nothing earth-shattering.  Shall merge now!

PR #381, opened 2019-08-13
[FR][ENG-444] improved GitHub revision handling

felliott:
## Ticket

https://openscience.atlassian.net/browse/ENG-444

## Purpose

Improve and consistify GH provider revision-identifying query-parameter support.

## Changes

* Unify query param handling: the first commit is a re-application of 9b541f2be708, which was reverted in 5fdc63f2d41e.  Updates the provider to support `ref`, `version`, `sha`, and `revision` as commit sha / branch name -identifying parameters.  Also supports `branch` as a branch name-identifying parameter.
* Allow `branch` to be a commit sha. Since the original implementation was written, it was discovered that some queries **are** passing commit shas in the `branch` parameter. Start supporting this.
* Verify sha length in sha vs. branch identifying heuristic.  GH doesn't support abbreviated SHAs, so improve the test by length-testing the sha string (currently 40 characters, but made configurable for after git newhash work is released).
* Fix revisions endpoint to work with refs.  A bug in the revisions endpoint meant the starting-revision was always being sent to the default branch.

## Side effects

There's a possibility that some urls may start directing to the wrong version.  Some urls in the wild have multiple query parameters with different refs.  The way WB handled this case was undefined before, but we've tried hard to make it DTRT.

## QA Notes

This PR specifically targets operations on non-default branches and old commits, so good places to test for breakage are: 

* Fangorn file listing
* Fangorn branch changer
* Viewing files in MFR
* File revisions
* Links to previous revisions in NodeLogs

No behavior should change, except that some previously-broken links should now work.

## Deployment Notes

* For the future: this PR adds a new setting, `GITHUB_SHA_LENGTHS` to support git Newhash lengths when they arrive.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -332,3 +332,35 @@ OSF-specific parameter used to identify special "view-only" links that are used 
 * **Type**: string
 * **Expected on**: ``GET`` requests for files or folders
 * **Notes**: Only used internally for the Open Science Framework.
+
+
+GitHub Provider Params
+++++++++++++++++++++++
+
+Query parameters specific to the GitHub provider.
+
+
+*commit / branch identification*
+********************************
+
+Not a single parameter, but rather a class of parameters.  WaterButler has used many different parameters to identify the branch or commit a particular file should be found under.  These parameters can be either a commit SHA or a branch name.  These parameters are ``ref``, ``version``, ``branch``, ``sha``, ``revision``.  All will continue to be supported to maintain back-compatibility, but ``ref`` (for SHAs or branch names) and ``branch`` (branch names only) are preferred.
+
+If both a SHA and a branch name are provided in different parameters, the SHA will take precedence.  If multiple parameters are given with different SHAs, then the order of precedence will be: ``ref``, ``version``, ``sha``, ``revision``, ``branch``.  If multiple parameters are given with different branches, the order of precedence is: ``branch``, ``ref``, ``version``, ``sha``, ``revision``.
+
+* **Type**: str
+* **Expected on**: Any GitHub provider request
+* **Interactions**: None
+
+
+fileSha
+*******
+
+Identifies a specific revision of a file via its SHA.
+
+.. warning::
+
+   **PLEASE DON'T USE THIS!**  This was a mistake and should have never been added.  It will hopefully be removed or at the very least demoted in a future version.  File SHAs only identify the contents of a file.  They provide no information about the file name, path, commit, branch, etc.
+
+* **Type**: str
+* **Expected on**: Any GitHub provider request
+* **Interactions**: The ``fileSha`` is always assumed to be a file revision that is an ancestor of the imputed   commit or branch ref.  Providing a ``fileSha`` for a file version that was committed after the imputed ref will result in a 404.
--- a/waterbutler/providers/github/provider.py
+++ b/waterbutler/providers/github/provider.py
@@ -86,8 +86,6 @@ class GitHubProvider(provider.BaseProvider):
             self.default_branch = self._repo['default_branch']
 
         ref, ref_type, ref_from = self._interpret_query_parameters(**kwargs)
-        if isinstance(ref, list):
-            raise exceptions.InvalidParameters('Only one ref or branch may be given.')
         self.metrics.add('branch_ref_from', ref_from)
         self.metrics.add('ref_type', ref_type)
 
@@ -122,8 +120,6 @@ class GitHubProvider(provider.BaseProvider):
             self.default_branch = self._repo['default_branch']
 
         ref, ref_type, ref_from = self._interpret_query_parameters(**kwargs)
-        if isinstance(ref, list):
-            raise exceptions.InvalidParameters('Only one ref or branch may be given.')
         self.metrics.add('branch_ref_from', ref_from)
         self.metrics.add('ref_type', ref_type)
 
@@ -983,7 +979,7 @@ class GitHubProvider(provider.BaseProvider):
 
         return new_head
 
-    def _interpret_query_parameters(self, **kwargs):
+    def _interpret_query_parameters(self, **kwargs) -> Tuple[str, str, str]:
         """This one hurts.
 
         Over the life of WB, the github provider has accepted the following parameters to identify
@@ -1013,8 +1009,12 @@ class GitHubProvider(provider.BaseProvider):
           ``ref``, ``version``, ``sha``, ``revision``. ``branch`` is used by fangorn and so far
           appears to always be a branch name.  The others follow the same priority for commit SHAs.
 
-        * If none of the query parameters are present or contain a valid commit SHA/branch name, WB
-          will fall back to the default branch for the repo.
+        * If multiple values are passed for one param, that param will be ignored and the check will
+          move on to the next params.  If the only params found have multiple values, an error will
+          be thrown to warn the user not to do this.
+
+        * If none of the query parameters are present, WB will fall back to the default branch for
+          the repo.
 
         This method also returns the type of ref it found (``commit_sha`` or ``branch``) and a
         string identifying the source of ref.  The former is used to avoid making a branch-specific
@@ -1031,11 +1031,17 @@ class GitHubProvider(provider.BaseProvider):
                 possible_values[param] = None
 
         inferred_ref, ref_from = None, None
+        multiple_args_seen = False
 
         # look for commit SHA likes ('branch' is least likely to be a sha)
         sha_priority_order = ['ref', 'version', 'sha', 'revision', 'branch']
         for param in sha_priority_order:
             v = possible_values[param]
+
+            if isinstance(v, list):
+                multiple_args_seen = True
+                continue
+
             if v is not None and self._looks_like_sha(v):
                 inferred_ref = v
                 ref_from = 'query_{}'.format(param)
@@ -1048,12 +1054,21 @@ class GitHubProvider(provider.BaseProvider):
         branch_priority_order = ['branch', 'ref', 'version', 'sha', 'revision']
         for param in branch_priority_order:
             v = possible_values[param]
+
+            if isinstance(v, list):
+                multiple_args_seen = True
+                continue
+
             if v is not None:
                 inferred_ref = v
                 ref_from = 'query_{}'.format(param)
                 break
 
         if inferred_ref is None:
+            if multiple_args_seen:
+                raise exceptions.InvalidParameters('Multiple values provided for parameter '
+                                                   '"{}". Only one ref or branch may be '
+                                                   'given.'.format(ref_from))
             inferred_ref = self.default_branch
             ref_from = 'default_branch'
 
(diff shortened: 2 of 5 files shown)


waterbutler/providers/github/provider.py
```diff
@@ -67,70 +67,75 @@ def __init__(self, auth, credentials, settings):
...
-            ref_from = 'query_ref'
-        elif kwargs.get('branch'):
-            branch_ref = kwargs.get('branch')
-            ref_from = 'query_branch'
-        else:
-            branch_ref = self.default_branch
-            ref_from = 'default_branch'
-        if isinstance(branch_ref, list):
+        ref, ref_type, ref_from = self._interpret_query_parameters(**kwargs)
+        if isinstance(ref, list):
```
cslzchen: I think the list check here is redundant now and can be removed. The `ref` in the tuple returned by `self._interpret_query_parameters()` can never be a list. It is either a valid GitHub SHA-like string or a branch name.

waterbutler/providers/github/provider.py
```diff
@@ -67,70 +67,75 @@ def __init__(self, auth, credentials, settings):
...
-            return GitHubPath(path, _ids=[(branch_ref, '')])
+            return GitHubPath(path, _ids=[(ref, '')])
 
-        branch_data = await self._fetch_branch(branch_ref)
+        tree_sha = None
+        if ref_type == 'branch_name':
+            branch_data = await self._fetch_branch(ref)
+            tree_sha = branch_data['commit']['commit']['tree']['sha']
+        else:
+            commit_data = await self._fetch_commit(ref)
```
cslzchen: (For myself) It is much clearer now since we know whether it is a SHA or a branch name. Thus we can make dedicated metadata requests respectively. In addition, 👍 for the naming: using `tree_sha` to generalize the SHA as well as not being confused with `branch_data` and `commit_data`.

waterbutler/providers/github/provider.py
```diff
@@ -303,7 +308,7 @@ def can_intra_copy(self, other, path=None):
     async def revisions(self, path, sha=None, **kwargs):
         resp = await self.make_request(
             'GET',
-            self.build_repo_url('commits', path=path.path, sha=sha or path.file_sha),
+            self.build_repo_url('commits', path=path.path, sha=sha or path.branch_ref),
```
cslzchen: I took a look at the commit message but still wondering why `path.file_sha` is erroneous? By its name, it looks like a the SHA we need when `sha=None`. Or there is a problem when using it to build the repo URL?

docs/api.rst
```diff
@@ -332,3 +332,31 @@ OSF-specific parameter used to identify special "view-only" links that are used
...
+
+Query parameters specific to the GitHub provider.
+
+
+*commit / branch identification*
+********************************
+
+Not a single parameter, but rather a class of parameters.  WaterButler has used many different parameters to identify the branch or commit a particular file should be found under.  These parameters can be either a commit SHA or a branch name.  These parameters are ``ref``, ``version``, ``branch``, ``sha``, ``revision``.  All will continue to be supported to maintain back-compatibility, but ``ref`` (for SHAs or branch names) and ``branch`` (branch names only) are preferred.
+
+If both a SHA and a branch name are provided in different parameters, the SHA will be take precedence.  If multiple parameters are given with different SHAs, then the order of precedence will be: ``ref``, ``version``, ``sha``, ``revision``, ``branch``.  If multiple parameters are given with different branches, the order of precedence is: ``branch``, ``ref``, ``version``, ``sha``, ``revision``.
```
cslzchen: Minor language issue: `the SHA will be take precedence` -> `the SHA will take precedence` or `the SHA will be taken precedence`.

waterbutler/providers/github/settings.py
```diff
@@ -12,3 +12,10 @@
 UPDATE_FILE_MESSAGE = config.get('UPDATE_FILE_MESSAGE', 'File updated on behalf of WaterButler')
 UPLOAD_FILE_MESSAGE = config.get('UPLOAD_FILE_MESSAGE', 'File uploaded on behalf of WaterButler')
 DELETE_FOLDER_MESSAGE = config.get('DELETE_FOLDER_MESSAGE', 'Folder deleted on behalf of WaterButler')
+
+
+# At some point in the near(?) future git will be changing its internal hash function from SHA-1
+# to SHA-256.  sha1-names are 40 hexdigits long and sha256-names are 64 hexdigits long.  At that
+# point, it seems probable that GitHub will update its API to accept both sha types. When that
+# happens, the following config var will need to be updated to include both sizes.
+GITHUB_SHA_LENGTHS = [int(x) for x in config.get('GITHUB_SHA_LENGTHS', '40').split(' ')]
```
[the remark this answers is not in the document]
felliott: Yep! But you might need to quote it (not sure): `GITHUB_SHA_LENGTHS="40 64"`.

And yes, you're right it probably would be good to document how to set that.  We already use that pattern in other settings in both WB and MFR.  The issue is that we need to be able to accept configuration through envvars in a dockerized environment.  A simple list like this can just be a delimited string.  For more complex lists or objects, we'd expect a json-encoded string.

There are a couple of the JIRA tickets I've backburnered related to this.  One is to go through all the `settings.py` file and make sure our config args are env-settable.  I think most of them we've just been converting as we go.  The other is to add some helper methods to [SettingsDict](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/settings.py#L7) to encapsulate some of these common patterns.  Those are more nice-to-haves then must-haves, but maybe we could add them to an LHF ticket list.

I also forgot that we have a [`CONFIGURING`](https://github.com/centerforopenscience/waterbutler#configuring) section in `README.md` that could probably stand an update.

waterbutler/providers/github/provider.py
```diff
@@ -955,3 +982,94 @@ def _prune_subtrees(self, tree):
...
+
+        * If none of the query parameters are present or contain a valid commit SHA/branch name, WB
+          will fall back to the default branch for the repo.
+
+        This method also returns the type of ref it found (``commit_sha`` or ``branch``) and a
+        string identifying the source of ref.  The former is used to avoid making a branch-specific
+        query during validation.  The latter is used for analytics.
+        """
+
+        possible_params = ['ref', 'version', 'sha', 'revision']
```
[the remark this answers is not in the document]
felliott: Exactly.  Though I'm wondering if it would just make sense to have two lists: `commit_sha_params` and `branch_name_params`.  There would be some duplication, but it might be clearer.  Hmmm...

waterbutler/providers/github/provider.py
felliott: Ooh, that's a good point.  I think that's there b/c tornado turns multiple of the same query param into a list.  I don't think my code handles that case.  If not, it should, and yes we can remove that check.

waterbutler/providers/github/provider.py
felliott: The problem with `file_sha` is that the `file_sha` is almost entirely useless for our purposes. The `file_sha` is just the hash of the file contents + a bit of metadata (not including the file path.  `file_sha` doesn't tell us anything about the file name, path, branch, etc.  I'm not sure what I was thinking when I made that part of the GitHubPath object but that was an absolute mistake.

`revisions()` is supposed to return the revision history of a file.  With regular storage providers that is pretty straightforward b/c there is only one "branch" or "commit" which is just the current state of the directory.  But with git-based providers, we need to be able to start our revision history from other branches/commits.

That said, `revisions()` probably shouldn't be letting the `sha` kwarg override the inferred `branch_ref`.  That just gives the `sha` query param primacy it doesn't deserve.

docs/api.rst
felliott: :+1: Good eyes!

waterbutler/providers/github/provider.py
felliott: Ah, right, now I remember this.  Yes, if we pass multiple of the same query param (`?sha=<sha1>&sha=<sha2>`), then tornado represents that as a list.  `_interpret_query_parameters` will pass that list to `_looks_like_sha`, which will throw a `TypeError` when calling `int()` on it.  `_looks_like_sha` catches that error and returns a `False`.  If this is the primary query parameter, then `_interpret_query_parameters` will return a `list`, which we check for and error on.

The purpose of that is to allow the case where there is a valid high-priority qp (like `ref`) and an invalid low-priority ref (like `revision`).  IOW something like `?ref=<sha1>&revision=<sha2>&revision=<sha3>` should return `('<sha1>', 'commit_sha', 'ref')` b/c we do have an unambiguous primary ref.  That said, this does seem like a convoluted way of doing it.  Lemme see if I can improve the logic.

waterbutler/providers/github/provider.py
felliott: Aha!  v1 doesn't pass the `sha` query param as an arg to `revisions()`.  Ergo, it's being ignored and is safe to delete.  Fixing.

waterbutler/providers/github/settings.py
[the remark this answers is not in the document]
felliott: I'm going to punt on a comprehensive fix and just add an example in this file.  We can tackle improving the docs & config code in a later sprint.

waterbutler/providers/github/provider.py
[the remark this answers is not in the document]
cslzchen: I agree. Either an extra comment or separating them into two lists will work. I did spent some time to figure it out.

waterbutler/providers/github/settings.py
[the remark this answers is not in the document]
cslzchen: I just realized that I had `GITHUB_SHA_LENGTHS=40 64` w/o quotes in OSF during my local testing so no need to quote it.

waterbutler/providers/github/settings.py
```diff
@@ -18,4 +18,7 @@
 # to SHA-256.  sha1-names are 40 hexdigits long and sha256-names are 64 hexdigits long.  At that
 # point, it seems probable that GitHub will update its API to accept both sha types. When that
 # happens, the following config var will need to be updated to include both sizes.
+#
+# Example for passing multiple length values via an envvar:
+#   $ GITHUB_SHA_LENGTHS="40 64" invoke server
```
cslzchen: I think on servers the `env` should be OK.

However, as far as [.docker-compose.wb.env](https://github.com/CenterForOpenScience/osf.io/blob/develop/.docker-compose.wb.env) is concerned, having `GITHUB_SHA_LENGTHS="40 64"` breaks Waterbutler with `ValueError: invalid literal for int() with base 10: '"40'` while `GITHUB_SHA_LENGTHS=40 64` works.

To trigger this, you have to load the settings in the main Waterbutler settings instead of GitHub settings. The settings for each individual provider cannot see the `env` set in `.docker-compose.wb.env`.

```python
# In "waterbutler/settings.py"
GITHUB_SHA_LENGTHS = [int(x) for x in config.get('GITHUB_SHA_LENGTHS', '40').split(' ')]

# In "waterbutler/providers/github/settings.py"
GITHUB_SHA_LENGTHS = settings.GITHUB_SHA_LENGTHS
```

PR #386, opened 2019-11-06
[ENG-337] [OATHPIT] Throw ownCloud into the Oath Pit

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/ENG-337

## Purpose

Enable and update the ownCloud provider for `aiohttp` 3.

## Changes

* WB `.make_request()` now supports WebDAV methods
  * `aiohttp.ClientSession` only has functions available for native HTTP methods. For WebDAV (a protocol that extends HTTP) ones, WB lets the `ClientSession` instance call `_request()` directly and then wraps the return object with `aiohttp.client._RequestContextManager`.

</br>

* WB client session now supports customized connector
  * For providers that use a customized connector such as owncloud, the new session is created (if no session is found in the event loop) with the given connector while an existing session simply ignores (and closes) the new connector. Given that the session is per loop for each instance, the existing session if found must already have a connector with qualified customizations.

</br>

* Use kwarg `ssl=` instead of `verify_ssl=` when initializing `aiohttp.TCPConnecto`
  * `verify_ssl=` has been deprecated and no longer takes any effect on setting the property `_ssl` of a `TCPConnector` instance. Use `ssl=` instead. For details see the following links.
    * https://github.com/aio-libs/aiohttp/blob/3.5/aiohttp/connector.py#L733
    * https://github.com/aio-libs/aiohttp/blob/3.5/aiohttp/client_reqrep.py#L152
    * https://github.com/aio-libs/aiohttp/blob/v0.18.4/aiohttp/connector.py#L402

## Side effects

No

## QA Notes

Here is a list of tests that I have done locally, all of which have passed. Extra notes and quirks are added under each list if there is any. In addition, tests are done for both non-root and root.

* Getting metadata for a file and folder
  * This is automatic when you list files and expand folders

* Downloading (20B, 1 MB, 11 MB, 88 MB, 264MB)

* Contiguous Uploading (20B, 1 MB, 11 MB, 88 MB, 264MB one-by-one and together)
  * Chunked upload: N / A

* DAZ (a folder containing five files of 20B, 1 MB, 11 MB, 88 MB, 264MB)

* Delete a file and a folder; delete multiple files and folders

* Folder creation and deletion

* Rename files and folders
  * After renaming a folder or file (successfully), clicking on the renamed folder triggers the following error. Please see the video attachment in the ticket.

click-on-an-renamed-file-or-folder

* Intra move and copy
  * ownCloud supports intra move / copy. My local tests includes moving file and folders between two ownCloud projects with the same ownCloud account.

* Inter move and copy
  * Between ownCloud and OSFStorage
  * Between ownCloud and another 3rd-party (S3 in this case)

* Comments persist with moves

* Revisions: N / A

* Project root is storage root vs. a subfolder
  * Not required but I still tested this since I have to create two projects  (one with root and the other with subfolder) anyway for testing intra copy / move.

* Updating a file

## Deployment Notes

N / A

### DIFF
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -187,25 +187,36 @@ class BaseProvider(metaclass=abc.ABCMeta):
             if value is not None
         }
 
-    def get_or_create_session(self):
+    def get_or_create_session(self, connector=None):
         """
         Obtain an existing session or create a new one for making requests.
 
         Quirks:
+
         Sessions must be carefully managed by WB.  On one hand, we can't just have one session for
         each provider instance since actions such as move and copy are run in background probably
         with a different loop.  On the other hand, we can't have one session for each request since
         sessions are only closed when the provider instance is destroyed.
 
+        For providers that use a customized connector such as owncloud, the new session is created
+        with the given connector; while an existing session simply ignores (and closes) the new
+        connector.  Given that the session is per event loop and instance, the existing session if
+        found must already have a connector with qualified customizations.
+
+        :param connector: a customized connector
         :return: the one session that belongs to the current event loop
         :rtype: :class:`aiohttp.ClientSession`
         """
         loop = asyncio.get_event_loop()
         session = self.loop_session_map.get(loop, None)
         if not session:
-            session = aiohttp.ClientSession()
+            session = aiohttp.ClientSession(connector=connector)
             self.loop_session_map[loop] = session
             self.session_list.append(session)
+        elif connector:
+            # Ignore and close the kwarg connector if an existing session exists.
+            connector._close()
+
         return session
 
     @throttle()
@@ -263,9 +274,10 @@ class BaseProvider(metaclass=abc.ABCMeta):
         byte_range = kwargs.pop('range', None)
         if byte_range:
             kwargs['headers']['Range'] = self._build_range_header(byte_range)
+        connector = kwargs.pop('connector', None)
+        session = self.get_or_create_session(connector=connector)
 
         method = method.upper()
-        session = self.get_or_create_session()
         while retry >= 0:
             # Don't overwrite the callable ``url`` so that signed URLs are refreshed for every retry
             non_callable_url = url() if callable(url) else url
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -10,6 +10,7 @@ from urllib import parse
 
 import furl
 import aiohttp
+from aiohttp.client import _RequestContextManager
 
 from waterbutler.core import streams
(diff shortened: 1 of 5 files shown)


waterbutler/core/provider.py
```diff
@@ -186,25 +187,36 @@ def build_headers(self, **kwargs) -> dict:
...
         loop = asyncio.get_event_loop()
         session = self.loop_session_map.get(loop, None)
         if not session:
-            session = aiohttp.ClientSession()
+            session = aiohttp.ClientSession(connector=connector)
             self.loop_session_map[loop] = session
             self.session_list.append(session)
+        elif connector:
+            # Ignore and close the kwarg connector if an existing session exists.
+            connector._close()
```
cslzchen: For ownCloud, the only extra requirement for connector is that SSL is enforced. Given that

> only one session is created for one event loop per provider instance

, an existing session found for an ownCloud instance must have been created by the ownCloud instance whose connector must also be a TCPConnector with SSL enforce. Thus, it is safe to reuse existing session (and its connector). Finally, the newly created connector should be closed.

waterbutler/settings.py
```diff
@@ -177,3 +177,5 @@ def child(key):
 keen_public_config = keen_config.child('PUBLIC')
 KEEN_PUBLIC_PROJECT_ID = keen_public_config.get_nullable('PROJECT_ID', None)
 KEEN_PUBLIC_WRITE_KEY = keen_public_config.get_nullable('WRITE_KEY', None)
+
+WEBDAV_METHODS = {'PROPFIND', 'MKCOL', 'MOVE', 'COPY'}
```
cslzchen: Only WebDAV methods used by WB / ownCloud are listed. Update this set if new ones are added in the future.

waterbutler/core/provider.py
```diff
@@ -186,25 +187,36 @@ def build_headers(self, **kwargs) -> dict:
...
+
+        :param connector: a customized connector
         :return: the one session that belongs to the current event loop
         :rtype: :class:`aiohttp.ClientSession`
         """
         loop = asyncio.get_event_loop()
         session = self.loop_session_map.get(loop, None)
         if not session:
-            session = aiohttp.ClientSession()
+            session = aiohttp.ClientSession(connector=connector)
```
cslzchen: `.request()` in `aiohttp-0.18` handles both sessions and requests while the new design in WB with `aiohttp-3.5` separates sessions and requests. The `connector=` is no longer a valid argument for `session._request()`. Instead, it is now an argument for initializing the session who makes the request later.

PR #388, opened 2019-11-08
[ENG-338] [OATHPIT] Throw S3 into the Oath Pit

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/ENG-338

## Purpose

Enable and update the S3 provider for `aiohttp` 3.

## Changes

It is unknown why `aiohttp-0.18.x` stores HTTP response header names in all-uppercase (e.g. `'CONTENT-LENGTH'`, `'ETAG'`, etc.). `aiohttp-3.5.x` now uses the more popular title case (e.g. `'Content-Length'`, `'ETAG'`, etc.). This breaks the old implementation for class `S3FileMetadataHeaders` whose calling `self.raw['ETAG']` throws `Key Error`. The class and related tests has been updated.

## Side effects

No

## QA Notes

Here is a list of tests that I have done locally, all of which have passed. Extra notes and quirks are added under each list if there is any. In short, S3 behaves as expected.

* Getting metadata for a file and folder
  * This is automatic when you list files and expand folders

* Downloading (20B, 1 MB, 11 MB, 88 MB, 264MB)

* Uploading (20B, 1 MB, 11 MB, 88 MB, 264MB one-by-one and together)
  * Both contiguous and chunked

* DAZ (a folder containing five files of 20B, 1 MB, 11 MB, 88 MB, 264MB)

* Delete a file and a folder; delete multiple files and folders

* Folder creation and deletion

* Renaming files and folders
  * Similar to ownCloud, after renaming a folder or file (successfully), clicking on the renamed folder triggers the following error.

click-on-an-renamed-file-or-folder

* Intra move and copy
  * Between two components connected to the same account and the same bucket.

* Inter move and copy
  * Between S3 and OSFStorage
  * Between S3 and another 3rd-party (ownCloud in this case)

* Comments persist with moves

* Revisions

* Updating a file

* Project root is storage root vs. a subfolder
  * N / A: s3 only allow bucket root as OSF project root

## Deployment Notes

TBD

### DIFF
--- a/tests/providers/s3/test_metadata.py
+++ b/tests/providers/s3/test_metadata.py
@@ -20,7 +20,7 @@ class TestFileMetadataHeaders:
         assert file_metadata_headers_object.materialized_path == '/test-path'
         assert file_metadata_headers_object.kind == 'file'
         assert file_metadata_headers_object.provider == 's3'
-        assert file_metadata_headers_object.size == 9001
+        assert file_metadata_headers_object.size == '9001'
         assert file_metadata_headers_object.size_as_int == 9001
         assert type(file_metadata_headers_object.size_as_int) == int
         assert file_metadata_headers_object.content_type == 'binary/octet-stream'
--- a/waterbutler/providers/s3/metadata.py
+++ b/waterbutler/providers/s3/metadata.py
@@ -28,15 +28,15 @@ class S3FileMetadataHeaders(S3Metadata, metadata.BaseFileMetadata):
 
     @property
     def size(self):
-        return self.raw['CONTENT-LENGTH']
+        return self.raw['Content-Length']
 
     @property
     def content_type(self):
-        return self.raw['CONTENT-TYPE']
+        return self.raw['Content-Type']
 
     @property
     def modified(self):
-        return self.raw['LAST-MODIFIED']
+        return self.raw['Last-Modified']
 
     @property
     def created_utc(self):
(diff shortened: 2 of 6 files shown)


waterbutler/providers/s3/metadata.py
```diff
@@ -28,30 +28,30 @@ def path(self):
...
+        return self.raw['Etag'].replace('"', '')
 
     @property
     def extra(self):
-        md5 = self.raw['ETAG'].replace('"', '')
+        md5 = self.raw['Etag'].replace('"', '')
         return {
             'md5': md5,
-            'encryption': self.raw.get('X-AMZ-SERVER-SIDE-ENCRYPTION', ''),
+            'encryption': self.raw.get('x-amz-server-side-encryption', ''),
```
cslzchen: Instead of converting all keys to upper-case in the HEAD response header and before being parsed in as `self.raw` to init the metadata, I decided to update the metadata class itself instead. The "title case" is more popular. Please note that HTTP headers themselves are not case-sensitive at all as far as the server / client is concerned.

tests/providers/s3/test_metadata.py
```diff
@@ -20,7 +20,7 @@ def test_file_metadata_headers(self, file_metadata_headers_object, file_header_m
         assert file_metadata_headers_object.materialized_path == '/test-path'
         assert file_metadata_headers_object.kind == 'file'
         assert file_metadata_headers_object.provider == 's3'
-        assert file_metadata_headers_object.size == 9001
+        assert file_metadata_headers_object.size == '9001'
```
cslzchen: The value of `Content-Length` is a string.

PR #395, opened 2020-01-21
[ENG-1393] Fix Sentry events due to `raven-python` -> `sentry-sdk`

cslzchen:
## Ticket

https://openscience.atlassian.net/browse/ENG-1393

## Purpose

Ticket [ENG-481](https://openscience.atlassian.net/browse/ENG-481) with [PR #387](https://github.com/CenterForOpenScience/waterbutler/pull/387) introduced a slight side-effect when `raven-python` was replaced by `sentry-sdk`. Logging messages at error level and above are now sent to Sentry as events by default. We’d love to change this back to it’s original behavior: only exceptions are sent.

## Changes

* Disabled any level of logs being sent as event 
* Bumped the `sentry-sdk` from version `0.13.2` to `0.14.0`.
* Improved tests for `async_retry`
* Improved `import` style on touched files

## Side effects

N / A

## QA Notes

N / A

### Dev QA

Monitor COS Sentry on staging3 (OATHPIT) after merge

## Deployment Notes

N / A

### DIFF
--- a/tests/core/test_utils.py
+++ b/tests/core/test_utils.py
@@ -10,36 +10,58 @@ class TestAsyncRetry:
 
     @pytest.mark.asyncio
     async def test_returns_success(self):
+        """Test the scenario where a function succeeds on first attempt.
+        """
         mock_func = mock.Mock(return_value='Foo')
         retryable = utils.async_retry(5, 0)(mock_func)
+
         x = await retryable()
+
+        # `mock_func` succeeds on first attempt; thus it should be called only once
         assert x == 'Foo'
         assert mock_func.call_count == 1
 
     @pytest.mark.asyncio
-    async def test_retries_until(self):
-        mock_func = mock.Mock(side_effect=[Exception(), 'Foo'])
+    async def test_retries_success(self):
+        """Test a scenario where a function fails first but succeeds after retrying.
+        """
+        mock_func = mock.Mock(side_effect=[Exception(), Exception(), 'Foo'])
         retryable = utils.async_retry(5, 0)(mock_func)
 
         x = await retryable()
 
+        # `mock_func` fails on the first two attempts but succeeds on the third one; thus it should
+        # be called exactly three times.
         assert x == 'Foo'
-        assert mock_func.call_count == 2
+        assert mock_func.call_count == 3
 
     @pytest.mark.asyncio
-    async def test_retries_then_raises(self):
+    async def test_retries_failed(self):
+        """Test a scenario where a function keeps failing / retrying until it reaches retry limit.
+        """
         mock_func = mock.Mock(side_effect=Exception('Foo'))
-        retryable = utils.async_retry(5, 0)(mock_func)
+        retryable = utils.async_retry(8, 0)(mock_func)
 
         with pytest.raises(Exception) as e:
-            coro = await retryable()
+            await retryable()
 
+        # `mock_func` keeps failing until it reaches the maximum retry limit which is 8; thus it
+        # should have been called 1 (initial) + 8 (retries) = 9 times before throwing an exception.
         assert e.type == Exception
         assert e.value.args == ('Foo',)
-        assert mock_func.call_count == 6
+        assert mock_func.call_count == 9
 
     @pytest.mark.asyncio
-    async def test_retries_by_its_self(self):
+    async def test_retries_by_itself(self):
+        """Test ``async_retry`` decorated coroutine itself.
+
+        This test takes care of the case where ``retryable()`` is called w/o ``await``, in which
+        case it returns a future that can run as long as nothing else forces it to yield in the
+        meantime.  The following ``await asyncio.sleep(.1)`` does exactly that.  As long as the
+        number of retries is low, ``retryable()`` can execute them all before the sleep is done.
+        However, if you remove the sleep or bump the number of retries up to a ludicrously high
+        number, the test fails b/c it starts asserting before all the retries have been exhausted.
+        """
         mock_func = mock.Mock(side_effect=Exception())
         retryable = utils.async_retry(8, 0)(mock_func)
 
@@ -47,19 +69,50 @@ class TestAsyncRetry:
 
         await asyncio.sleep(.1)
 
+        # `mock_func` keeps failing until it reaches the maximum retry limit which is 8; thus it
+        # should be called 1 (initial) + 8 (retries) = 9 times
         assert mock_func.call_count == 9
 
+    @pytest.mark.asyncio
+    async def test_all_retry(self):
+        """Test multiple ``async_retry`` decorated coroutines being called at the same time.
+
+        This test is similar to ``test_retries_by_itself()`` where it calls ``retryable()`` w/o
+        ``await``. The only difference is that it tests multiple coroutines with multiple calls to
+        better mimic the scenario in ``waterbutler.core.remote_logging.log_file_action()``.
+        """
+
+        mock_func_a = mock.Mock(side_effect=Exception())
+        mock_func_b = mock.Mock(side_effect=Exception())
+        retryable_a = utils.async_retry(4, 0)(mock_func_a)
+        retryable_b = utils.async_retry(4, 0)(mock_func_b)
+
+        retryable_a()
+        retryable_b()
+        retryable_a()
+        retryable_b()
+
+        await asyncio.sleep(.1)
+
+        # Each call of `mock_func_a` or `mock_func_b` keeps failing until it reaches the maximum
+        # retry limit which is 4; thus for either of the two, the total call count should be the
+        # same: 2 (each has been called twice) * (1 (initial) + 4 (retries)) = 10 times.
+        assert mock_func_a.call_count == 10
+        assert mock_func_b.call_count == 10
+
     def test_docstring_survives(self):
-        async def mytest():
-            '''This is a docstring'''
+
+        async def my_test():
+            """This is a docstring"""
             pass
 
-        retryable = utils.async_retry(8, 0)(mytest)
+        retryable = utils.async_retry(8, 0)(my_test)
 
-        assert retryable.__doc__ == '''This is a docstring'''
+        assert retryable.__doc__ == """This is a docstring"""
 
     @pytest.mark.asyncio
     async def test_kwargs_work(self):
+
         async def mytest(mack, *args, **kwargs):
             mack()
             assert args == ('test', 'Foo')
@@ -73,18 +126,6 @@ class TestAsyncRetry:
         assert await fut
 
         assert merk.call_count == 2
-
-    @pytest.mark.asyncio
-    async def test_all_retry(self):
-        mock_func = mock.Mock(side_effect=Exception())
-        retryable = utils.async_retry(8, 0)(mock_func)
-
-        retryable()
-        retryable()
-
-        await asyncio.sleep(.1)
-
-        assert mock_func.call_count == 18
 
 
 class TestContentDisposition:
--- a/waterbutler/core/utils.py
+++ b/waterbutler/core/utils.py
@@ -12,11 +12,8 @@ from urllib import parse
 import aiohttp
 import sentry_sdk
 from stevedore import driver
-from sentry_sdk.integrations.logging import LoggingIntegration
 
 from waterbutler.core import exceptions
-from waterbutler.settings import config
-from waterbutler.version import __version__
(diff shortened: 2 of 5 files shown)


waterbutler/core/utils.py
```diff
@@ -13,21 +13,15 @@
...
-from waterbutler.version import __version__
+from waterbutler.server import settings as server_settings
 
 logger = logging.getLogger(__name__)
 
 signer = Signer(server_settings.HMAC_SECRET, server_settings.HMAC_ALGORITHM)
 
-sentry_dsn = config.get_nullable('SENTRY_DSN', None)
-if sentry_dsn:
-    sentry_sdk.init(sentry_dsn, release=__version__)
```
cslzchen: Removed since it does nothing except overriding the global Sentry `client` either initialized by `tornado` or `celery`. Tested and confirmed locally that w/ or w/o this (log updated version), sentry behavior remains the same.

### More Notes

The [diff](https://github.com/CenterForOpenScience/waterbutler/commit/ecd9e3a94975892517efb2ef273f5b18af87bbef#diff-403e14f805949c00c29047db245ce114L67-L97) as shown below shows why `client` was needed but no longer now. `raven-sentry`'s client is local and must be defined and initialized before being used / passed to `async_retry`. In contrast, `sentry-sdk`'s client is global.

* Now with `sentry-sdk`

```python
def async_retry(retries=5, backoff=1, exceptions=(Exception, )):
    ...
    with sentry_sdk.configure_scope() as scope:
        scope.set_tag('debug', False)
    ...
```

* Previously with `raven-sentry`

```python
client = Client(sentry_dsn) if sentry_dsn else None
...
def async_retry(retries=5, backoff=1, exceptions=(Exception, ), raven=client):
    ...
    if raven:
        # Only log if a raven client exists
        client.captureException()
    ...
```

tests/core/test_utils.py
```diff
@@ -27,36 +26,26 @@ class TestAsyncRetry:
...
-    async def test_retries_by_its_self(self):
-        mock_func = mock.Mock(side_effect=Exception())
-        retryable = utils.async_retry(8, 0)(mock_func)
-
-        retryable()
-
-        await asyncio.sleep(.1)
-
-        assert mock_func.call_count == 9
-
```
cslzchen: Why I removed this test:

* I don't understand while it used to call coroutine `retryable()` without `await` and without `with pytest.raises() ... as ...`. As a result, the test failed with 1) exception being raised and 2) `asyncio` complaining about "task was destroyed but it is pending".

* After I fixed the issue, this test and `test_all_retry()` test exactly what `test_retries_then_raises()` / `test_retries_failed()` does.

tests/core/test_utils.py
```diff
@@ -74,18 +63,6 @@ def test_docstring_survives(self):
...
-        mock_func = mock.Mock(side_effect=Exception())
-        retryable = utils.async_retry(8, 0)(mock_func)
-
-        retryable()
-        retryable()
-
-        await asyncio.sleep(.1)
-
-        assert mock_func.call_count == 18
-
```
cslzchen: Ditto and removed ...

tests/core/test_utils.py
felliott: This is a tricky one.  The tests *do* test separate things, but could definitely use a comment or docstring to explain the difference.  There are two groups:

`test_returns_success`, `test_retries_until`, and `test_retries_then_raises` are all testing the case where we are `await`-ing the result of `retryable()`.  In this case, we are not yielding at all and can guarantee that all possible retries are done up to we get a successful result.  You can see this by modifying `test_retries_then_raises` to retry an absurd number of times (I tested up to 10,000).  No matter how high you raise it, as long as the `.call_count` assert is `$retries + 1`, the test will pass.

OTOH in `test_retries_by_its_self` we don't `await` `retryable()`.  `retryable()` returns a future, meaning it can run as long as nothing else forces it to yield in the meantime.  But the following `await asyncio.sleep(.1)` does exactly that.  As long as the number of retries is low, `retryable` can execute them all before the sleep is done.  But if you remove the sleep, or if you bump the number of retries up to a ludicrously high number (I used 1000), that test will fail b/c it starts asserting before all of the retries have been exhausted.   You can also see this by playing around with the sleep time.  If I set the sleep time to `.01`, only 14 retries are completed before the `assert`s are run.

This leaves us with a conundrum, since there are a lot of environment dependent magic numbers in the explanation above.  I think the solution is a two-parter:

1. Mention the above in the test methods docstrings.
2. Add comments by each `assert mock_func == <foo_count>` explaining why the count is what is. e.g. for `test_all_retry` it would be something like `# (8 retries+1 final try) + (8 retries + 1 final try)`

We can't really escape the magic numbers, but at least we can explain what they're for.

waterbutler/core/utils.py
felliott: Good catch! I completely overlooked that!

tests/core/test_utils.py
```diff
@@ -27,36 +26,26 @@ class TestAsyncRetry:
...
         with pytest.raises(Exception) as e:
-            coro = await retryable()
+            await retryable()
 
         assert e.type == Exception
         assert e.value.args == ('Foo',)
         assert mock_func.call_count == 6
 
-    @pytest.mark.asyncio
-    async def test_retries_by_its_self(self):
```
felliott: Legacy grammar nitpick: `its_self` should be `itself`

tests/core/test_utils.py
cslzchen: Thanks for the explanation and now I understand. I will add back the test, keep the name and add docstr / comments to explain these tests. The error that I mentioned / encountered may be related to the magic numbers.

cslzchen: I think I figured out the issue `Task was destroyed but it is pending!` that I ran into. Tests was actually passing but I thought it didn't due to many of the following errors polluting the output. 
In addition, I have `assert False` to trigger my extra logs during tests to confirm `async_retry` works as expected.

These following "logging error" happens only when Sentry is enabled. My guess is that Sentry is still trying to send events but the task is destroyed right after tests are done.

```
--- Logging error ---
  File "/usr/local/lib/python3.6/logging/__init__.py", line 996, in emit
    stream.write(msg)
  File "/usr/local/lib/python3.6/site-packages/_pytest/capture.py", line 427, in write
    self.buffer.write(obj)
ValueError: I/O operation on closed file
Call stack:
  File "/usr/local/lib/python3.6/threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "/usr/local/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/threading.py", line 67, in run
    return old_run_func(self, *a, **kw)
  File "/usr/local/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/worker.py", line 127, in _target
    callback()
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/transport.py", line 208, in send_event_wrapper
    self._send_event(event)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/transport.py", line 143, in _send_event
    "Content-Encoding": "gzip",
  File "/usr/local/lib/python3.6/site-packages/urllib3/request.py", line 80, in request
    method, url, fields=fields, headers=headers, **urlopen_kw
  File "/usr/local/lib/python3.6/site-packages/urllib3/request.py", line 171, in request_encode_body
    return self.urlopen(method, url, **extra_kw)
  File "/usr/local/lib/python3.6/site-packages/urllib3/poolmanager.py", line 330, in urlopen
    response = conn.urlopen(method, u.request_uri, **kw)
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 672, in urlopen
    chunked=chunked,
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 376, in _make_request
    self._validate_conn(conn)
  File "/usr/local/lib/python3.6/site-packages/urllib3/connectionpool.py", line 994, in _validate_conn
    conn.connect()
  File "/usr/local/lib/python3.6/site-packages/urllib3/connection.py", line 420, in connect
    _match_hostname(cert, self.assert_hostname or server_hostname)
  File "/usr/local/lib/python3.6/site-packages/urllib3/connection.py", line 430, in _match_hostname
    match_hostname(cert, asserted_hostname)
  File "/usr/local/lib/python3.6/site-packages/idna_ssl.py", line 15, in patched_match_hostname
    hostname = idna.encode(hostname, uts46=True).decode('ascii')
  File "/usr/local/lib/python3.6/site-packages/idna/core.py", line 345, in encode
    s = uts46_remap(s, std3_rules, transitional)
  File "/usr/local/lib/python3.6/site-packages/idna/core.py", line 314, in uts46_remap
    from .uts46data import uts46data
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 665, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 674, in exec_module
  File "<frozen importlib._bootstrap_external>", line 779, in get_code
  File "<frozen importlib._bootstrap_external>", line 487, in _compile_bytecode
  File "/usr/local/lib/python3.6/asyncio/base_events.py", line 1310, in call_exception_handler
    self.default_exception_handler(context)
  File "/usr/local/lib/python3.6/asyncio/base_events.py", line 1285, in default_exception_handler
    logger.error('\n'.join(log_lines), exc_info=exc_info)
  File "/usr/local/lib/python3.6/logging/__init__.py", line 1337, in error
    self._log(ERROR, msg, args, **kwargs)
  File "/usr/local/lib/python3.6/logging/__init__.py", line 1444, in _log
    self.handle(record)
  File "/usr/local/lib/python3.6/logging/__init__.py", line 1454, in handle
    self.callHandlers(record)
  File "/usr/local/lib/python3.6/site-packages/sentry_sdk/integrations/logging.py", line 77, in sentry_patched_callhandlers
    return old_callhandlers(self, record)
Message: 'Task was destroyed but it is pending!\ntask: <Task pending coro=<wrapped() running at /code/waterbutler/core/utils.py:108> wait_for=<Task pending coro=<wrapped() running at /code/waterbutler/core/utils.py:108> wait_for=<Task pending coro=<wrapped() running at /code/waterbutler/core/utils.py:108> wait_for=<Task finished coro=<wrapped() done, defined at
/code/waterbutler/core/utils.py:92> exception=Exception()> cb=[<TaskWakeupMethWrapper object at 0x7efc65622648>()]> cb=[<TaskWakeupMethWrapper object at 0x7efc65622528>()]> cb=[<TaskWakeupMethWrapper object at 0x7efc65622408>()]>'
Arguments: ()
```

tests/core/test_utils.py
```diff
@@ -10,53 +10,105 @@ class TestAsyncRetry:
...
+        assert mock_func.call_count == 3
 
     @pytest.mark.asyncio
-    async def test_retries_then_raises(self):
+    async def test_retries_failed(self):
+        """Test a scenario where a function keeps failing / retrying until it reaches retry limit.
+        """
         mock_func = mock.Mock(side_effect=Exception('Foo'))
-        retryable = utils.async_retry(5, 0)(mock_func)
+        retryable = utils.async_retry(8, 0)(mock_func)
```
cslzchen: Minor tweak: set max retry limit to 8 (was 5), and assertion updated

tests/core/test_utils.py
```diff
@@ -10,56 +10,110 @@ class TestAsyncRetry:
...
+        retryable_a()
+        retryable_b()
+
+        await asyncio.sleep(.1)
+
+        # Each call of `mock_func_a` or `mock_func_b` keeps failing until it reaches the maximum
+        # retry limit which is 4; thus for either of the two, the total call count should be the
+        # same: 2 (each has been called twice) * (1 (initial) + 4 (retries)) = 10 times.
+        assert mock_func_a.call_count == 10
+        assert mock_func_b.call_count == 10
```
cslzchen: Minor tweak: the unit tests to better mimic [waterbutler.core.remote_logging.log_file_action()](https://github.com/CenterForOpenScience/waterbutler/blob/develop/waterbutler/core/remote_logging.py#L213-L224) where two `async_retry` decorated function `log_to_callback()` and `_send_to_keen()` (via `log_to_keen()`)  are called.

PR #403, opened 2023-06-21
[ENG-4579] Allow the base folder to not be a bucket for S3

Johnetordoff:
## OSF side PR

https://github.com/CenterForOpenScience/osf.io/pull/10414/

## Purpose

To allow water butler to accept different paths as the root folder, not just the bucket route.This uses the `settings`  JSON to pass information about the users root folder, infomation sent via the path will not change, but the user will now have a base root folder that will be their `folder_id`, which currently is always their bucket name. 

## Changes

- Extra logic to allow for bucket setting to be used.
- Excludes parent file from results.

## Side effects

N/A

## QA Notes

See QA plans in project notion

## Deployment Notes

To test this I wrote my own docker-compose:
```
version: '3.4'

volumes:
  wb_requirements_vol:
    external: false
  wb_requirements_local_bin_vol:
    external: false
  wb_tmp_vol:
    external: false
  osfstoragecache_vol:
    external: false
  rabbitmq_vol:
    external: false

services:
  ###############
  # WaterButler #
  ###############

  wb_requirements:
    build: .
    command:
      - /bin/bash
      - -c
      - invoke install --develop &&
        (python -m compileall /usr/local/lib/python3.6 || true) &&
        rm -Rf /python3.6/* &&
        cp -Rf -p /usr/local/lib/python3.6 /
    restart: 'no'
    volumes:
      - wb_requirements_vol:/python3.6
      - wb_requirements_local_bin_vol:/usr/local/bin
  wb:
    build: .
    command: invoke server
    restart: unless-stopped
    ports:
      - 7777:7777
    env_file:
      - .docker-compose.wb.env
    volumes:
      - wb_requirements_vol:/usr/local/lib/python3.6
      - wb_requirements_local_bin_vol:/usr/local/bin
      - osfstoragecache_vol:/code/website/osfstoragecache
      - wb_tmp_vol:/tmp
    stdin_open: true
  wb_worker:
    build: .
    command: invoke celery
    restart: unless-stopped
    depends_on:
      - wb
      - rabbitmq
    environment:
      C_FORCE_ROOT: 1
    env_file:
      - .docker-compose.wb.env
    volumes:
      - wb_requirements_vol:/usr/local/lib/python3.6
      - osfstoragecache_vol:/code/website/osfstoragecache
      - wb_tmp_vol:/tmp
    stdin_open: true
  rabbitmq:
    image: rabbitmq:3-management
    ports:
      - 5672:5672
      - 15672:15672
    volumes:
      - rabbitmq_vol:/var/lib/rabbitmq
    stdin_open: true
```
and used that while taking the osf.io side `wb` container down, using my own S3 credentials.

## Ticket

https://openscience.atlassian.net/browse/ENG-4579

### DIFF
--- a/waterbutler/providers/s3/provider.py
+++ b/waterbutler/providers/s3/provider.py
@@ -685,7 +685,11 @@ class S3Provider(provider.BaseProvider):
     async def _metadata_folder(self, path):
         await self._check_region()
 
-        params = {'prefix': path.path, 'delimiter': '/'}
+        if path._orig_path == '/' and self.settings.get('id'):
+            params = {'prefix': self.settings['id'], 'delimiter': '/'}
+        else:
+            params = {'prefix': path.path, 'delimiter': '/'}
+
         resp = await self.make_request(
             'GET',
             functools.partial(self.bucket.generate_url, settings.TEMP_URL_SECS, 'GET', query_parameters=params),
@@ -721,11 +725,15 @@ class S3Provider(provider.BaseProvider):
 
         items = [
             S3FolderMetadata(item)
-            for item in prefixes
+            for item in prefixes if item['Prefix'] != path.path
         ]
 
         for content in contents:
             if content['Key'] == path.path:
+                continue
+
+            base_folder = self.settings.get('id')
+            if base_folder and content['Key'] == base_folder:
                 continue
 
             if content['Key'].endswith('/'):


waterbutler/providers/s3/provider.py
```diff
@@ -721,13 +725,17 @@ async def _metadata_folder(self, path):
 
         items = [
             S3FolderMetadata(item)
-            for item in prefixes
+            for item in prefixes if item['Prefix'] != path.path
```
Johnetordoff, line 728: S3's query system doesn't give all children for a folder, instead matches everything that starts with that path, so we have to remove the parent from the list as obviously it matches it's own path.

waterbutler/providers/s3/provider.py
```diff
@@ -721,13 +725,17 @@ async def _metadata_folder(self, path):
...
         ]
 
         for content in contents:
             if content['Key'] == path.path:
                 continue
 
+            base_folder = self.settings.get('id')
+            if base_folder and content['Key'] == base_folder:
```
Johnetordoff, line 736: Similar to above, but for `Key` which is there term for files object names, as opposed to folders.

cslzchen (approved):
LGTM 🎆 

Cc @felliott

PR #405, opened 2023-07-26
[ENG-4579] S3 Improvement Changes for WB

Johnetordoff:
## Ticket

https://openscience.atlassian.net/browse/ENG-4579

## Purpose

Allow path for s3 addon to be set with a base folder. so only part a of a bucket is selectable

## Changes

-  Change the path given to match the S3 prefix
-  Change the path given correct for path validation


cslzchen (approved):
🎆

PR #406, opened 2023-07-26
[ENG-4624] [S3 Improvements] Project PR - Waterbutler Part

cslzchen:
## Purpose

Enables WB to support both bucket-root and subfolder-root configuration for S3.

Credit @Johnetordoff for all the work 👍 

Project notion: https://www.notion.so/cos/S3-improvements-476a5b07cb7e4f458e9cd4c77cfa03ec
OSF Part: https://github.com/CenterForOpenScience/osf.io/pull/10416

## Changes

* Enables WB to support both bucket-root and subfolder-root configuration for S3
* Updated WB to support `:/` as the new delimiter for S3 bucket
* Added/Updated unit tests

## DevOps Notes

* Deployment requires downtime, must be deployed at the same time as [OSF](https://github.com/CenterForOpenScience/osf.io/pull/10416)
* For details, see our release playbook (TBD)

## Dev Notes

Here are all child PRs that have been merged into this feature branch.

- Unit tests: https://github.com/CenterForOpenScience/waterbutler/pull/408
- Clean-up: https://github.com/CenterForOpenScience/waterbutler/pull/407
- Path fix: https://github.com/CenterForOpenScience/waterbutler/pull/405
- Delimiter: https://github.com/CenterForOpenScience/waterbutler/pull/404 (Use `:/`)
- Prototype Subfolder: https://github.com/CenterForOpenScience/waterbutler/pull/403

## QA Notes

See QA docs in the [project notion page](https://www.notion.so/cos/S3-improvements-476a5b07cb7e4f458e9cd4c77cfa03ec)

## Documentation

Update our developers doc for S3 (if any)

## Side Effects

* WB still uses `boto` and thus won't support new regions added by `boto3`.

## Ticket

https://openscience.atlassian.net/browse/ENG-4624

### DIFF
--- a/waterbutler/providers/s3/provider.py
+++ b/waterbutler/providers/s3/provider.py
@@ -69,7 +69,10 @@ class S3Provider(provider.BaseProvider):
         await self._check_region()
 
         if path == '/':
-            return WaterButlerPath(path)
+            # adjust path using base folder to include the buckets root, so just the `/` translates to just
+            # the `/base_folder/` path
+            base_folder = self.settings.get('id', ':/').split(':/')[1]
+            return WaterButlerPath(f'/{base_folder}')
 
         implicit_folder = path.endswith('/')
 
@@ -685,12 +688,10 @@ class S3Provider(provider.BaseProvider):
     async def _metadata_folder(self, path):
         await self._check_region()
 
-        if path._orig_path == '/' and self.settings.get('id'):
-            base_folder = self.settings['id'].split(':/')[1]
-            params = {'prefix': base_folder, 'delimiter': '/'}
-        else:
-            base_folder = None
-            params = {'prefix': path.path, 'delimiter': '/'}
+        # The user selected base folder, the root of the where that user's node is connected.
+        prefix = self.settings['id'].split(':/')[1] if path == '/' and self.settings.get('id') else path.path
+
+        params = {'prefix': prefix, 'delimiter': '/'}
 
         resp = await self.make_request(
             'GET',
@@ -731,10 +732,7 @@ class S3Provider(provider.BaseProvider):
         ]
 
         for content in contents:
-            if content['Key'] == path.path:
-                continue
-
-            if base_folder and content['Key'] == base_folder:
+            if content['Key'] == params['prefix']:
                 continue
 
             if content['Key'].endswith('/'):
--- a/waterbutler/providers/s3/provider.py
+++ b/waterbutler/providers/s3/provider.py
@@ -686,8 +686,10 @@ class S3Provider(provider.BaseProvider):
         await self._check_region()
 
         if path._orig_path == '/' and self.settings.get('id'):
-            params = {'prefix': self.settings['id'], 'delimiter': '/'}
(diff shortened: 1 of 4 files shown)


waterbutler/providers/s3/provider.py
```diff
@@ -685,7 +688,11 @@ async def _metadata_file(self, path, revision=None):
     async def _metadata_folder(self, path):
         await self._check_region()
 
-        params = {'prefix': path.path, 'delimiter': '/'}
+        # The user selected base folder, the root of the where that user's node is connected.
+        prefix = self.settings['id'].split(':/')[1] if path == '/' and self.settings.get('id') else path.path
```
felliott, line 692: I am pretty sure this is buggy.  The two sides of this `if` statement are not equivalent.  The LHS side is the storage provider prefix.  While WB needs this to send meaningful queries to S3, it is not something that we (directly) expose to the user.  The RHS is the file/folder path that we return to the user in the WB metadata responses.  If the prefix is `/` i.e. not a subfolder, these two things end up being equivalent.  But if the storage provider root folder is a subfolder, e.g. `/foo/`, then the LHS is `/foo/` and the RHS is `/bar/`.

I'm not a 100% convinced I explained this correctly, so maybe an example would be better. On staging, connect a subfolder (for this example, lets call it `blemmo`) of a bucket to a node.  Then fetch the WB metadata listing for the root folder (e.g. `https://files.us.staging.osf.io/v1/resources/mst3k/providers/s3/?meta`). Not only will `blemmo` show up in the file/folder path metadata, but you can get the same response by appending the prefix name to the url (`https://files.us.staging.osf.io/v1/resources/mst3k/providers/s3/blemmo/?meta`).  That's weird.

WB file/folder metadata is defined *relative* to the storage provider prefix.  We treat the prefix as the virtual root. If the prefix is the bucket root folder, then YAY, no extra work for us.  If not, we need to do the path math to make sure our paths are adjusted properly.

felliott (changes requested):
There's some confusion in the path math for metadata. It might need some adjustments.  It'd probably also be good to add some tests for metadata listings when the prefix is a bucket subfolder.

felliott:
I think the thing we need to be looking at for this is the `prepend` kwarg to `WaterButlerPath`.  When building a WBPath object, the first arg is the relative path (`/bar/baz.txt`) and the prepend kwarg is the storage provider root (`/foo/`).  Such a WBPath would represent the file `/foo/bar/baz.txt` on the storage provider, but it's path metadata will be reported as `/bar/baz.txt` since that is the part relative to the project root.  See the owncloud provider for a reference.  owncloud is both path-based and supports subfolder roots and is probably the closest thing to what you want to do here.

The other thing i would recommend is to go ahead and do the `{bucket}:{storage_root}` split in the init method. You can save the results in attributes and refer to them throughout the code.

cslzchen:
fyi, Johnny's response to CR https://github.com/CenterForOpenScience/waterbutler/pull/409 has been merged as https://github.com/CenterForOpenScience/waterbutler/pull/406/commits/7683994054dd906c7c82afb0159a9be650cfde3c

PR #419, opened 2025-03-26
[ENG-7518] Update python code for python3.13 compatibility

mkovalua:
## Ticket

https://openscience.atlassian.net/browse/ENG-7518

## Purpose

Update python code for python3.13 compatibility. Should pass python compile check (python3 -m compileall -q .)

## Changes

Update code to pass check (python3 -m compileall -q .) 
image
,
update code style to be it suitable with 3.13 python version related linters (it looks more strict than in previous versions) and allow to run unit tests.

image

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 aiocontextvars==0.2.2  # recommended for sentry-sdk
-aiohttp==3.11.11
+aiohttp==3.9.4
 boto==2.49.0
 celery==5.4.0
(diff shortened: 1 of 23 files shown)


felliott:
Looks good, but I removed the line-break related changes. Let's keep those until we start applying autoformatting (e.g. ruff) in a future ticket.  Some of those line-breaks are useful for understanding cascading conditions. Post-autoformatting, we may need to restructure the comments we leave on complex conditions.

PR #420, opened 2025-03-31
[ENG-7519] 2.2.3 BE: WB Fix deprecation warnings for py3.13+

mkovalua:
## Ticket

https://openscience.atlassian.net/browse/ENG-7519

## Purpose

Fix any deprecation warnings for upcoming versions of python. Should pass python compile check, but does not need to pass tests.

## Changes

Update code to fix deprecations (use Qodana code inspection tool to find deprecations and possible problems )

image

image

image

image

image

image

image

image

## Side effects

The PR was created from branch feature/ENG-7518 to keep another PR updates

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,8 +2,9 @@
 # E127: continuation line over-indented for visual indent
 # E128: continuation line under-indented for visual indent
 # E731: do not assign a lambda expression, use a def
+# W504: line break after binary operator
 [flake8]
-ignore = E501,E731,E127,E128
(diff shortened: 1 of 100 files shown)


felliott:
Looks good! I made some changes. I removed comments that weren't relevant and added comments and default values that were. I shortened commit messages to make the first line wrap within 72 chars.  I removed the "accessing a protected member" commit. It's not wrong, but it's indicative of APIs that need a broader rethink/redesign.  The changes will be more appropriate then.

PR #428, opened 2025-04-17
[ENG-7847] 2.4.4: BE: Fix broken upload reads in WB for python 3.13

mkovalua:
## Ticket

https://openscience.atlassian.net/browse/ENG-7847

## Purpose

WB cant seem to read file uploads from tornado. Fix that!

## Changes

Change  `async def prepare_stream(self):`   `async def data_received(self, chunk)` write/read stream logic to make it possible to work with python 3.13

https://github.com/user-attachments/assets/f3ddb9e3-4f31-41c1-88b2-e4ba2a697e70

## Side effects

It looks like a low level code changings so maybe there is more explicit way to do it to keep it more clear.


mkovalua:
https://github.com/CenterForOpenScience/waterbutler/pull/428/commits/72b72cca6ba6d6ad3d29617242973657db9eabc0

image

mkovalua:
https://github.com/CenterForOpenScience/waterbutler/pull/428/commits/633beadc5506a16724764e6cd3edfa78114ed9ee

image

PR #431, opened 2025-05-05
ENG-7526 | fixed WB Dockerfile builds successfully

sh-andriy:


felliott (changes requested):
Doesn't this bypass running newrelic? We still want to have the option to run it in a dev environment.

PR #426, opened 2025-04-09
[ENG-7683] integrate aws s3 with python3.13

mkovalua:
## Ticket

https://openscience.atlassian.net/browse/ENG-7683

## Purpose

Make boto possible to interact with aws s3 in python 3.13. 

Created integration of boto as a base but try to find a way how to replace boto calls with aiobotocore (based on boto3 and botocore) it is async that is plus for performance and is much more newer but not confident all the methods that exists in boto may be implemented in aiobotocore

(Just for testing purposes)

## Changes

image

image

image

osf docker-compose.yml

https://drive.google.com/file/d/1QwVeyp83YpsIIXwjDBfOqzxqW6PWjF6k/view?usp=sharing

## Side effects

Need to resolve an issue described in slack

`there is no stream data passed in WB for python 3.13
`

https://centerforopenscience.slack.com/archives/C08C50R8EA2/p1744392546135499

Not confident about usage of intra_copy where is it called to test it out, supposed there but it was not triggered 

image

image

image

PR #433, opened 2025-05-09
[ENG-7946] Add boto support in python 3 13 support big files

mkovalua:
## Ticket

https://openscience.atlassian.net/browse/ENG-7946

## Purpose

Downloads from s3 and owncloud are hanging upredictably. Why?

## Changes

the PR
https://github.com/CenterForOpenScience/waterbutler/pull/426
 is made from the following one PR to keep its changings as well

for such a purpose we avoid aiobotocore and use botocore in seperate thread to be it work as async and dont freeze another workflows it looks like there is a problem with async  aiobotocore stream reading (maybe it is good idea as well to tell about aiobotocore issue into official github account, maybe they fix it somehow in future, because they had already similar issue
https://github.com/aio-libs/aiobotocore/issues/1235 and possible solution did not help for our https://github.com/aio-libs/aiohttp/issues/10169#issuecomment-2547123284 functionality)

1. For now we may download larger files 
https://centerforopenscience.slack.com/archives/C08NK9R1P7C/p1746714237815109

2. Also zip files are downloading properly

https://github.com/user-attachments/assets/a3580bbc-338c-4d1e-b542-a906ed864070

3. about owncloud it looks like it is proper configuration an it works for now

image

image


felliott:
Looks good. We'll explore building urls in a followup ticket and look at speedup/slowdown possibilities as well. I've also updated the branch to use poetry for deps you added.

PR #453, opened 2025-07-24
[ENG-8404] path is needed to be started with '/' 

mkovalua:
on coping from bitbucket to dropbox (suppose it may be common for all providers)

## Ticket

https://openscience.atlassian.net/browse/ENG-8404

## Purpose

Fix coping file from bitbucket to dropbox

`waterbutler.core.exceptions.MetadataError: 400, {"response": "Error in call to API function \"files/get_metadata\": request body: path: '.gitignore' did not match pattern '(/(.|[\\r\\n])*|id:.*)|(rev:[0-9a-f]{9,})|(ns:[0-9]+(/.*)?)'"}`

image

image

## Changes

adding / at the path beginning if it is not set before `self.make_request` call to avoid possible errors 

image

## Side effects

Maybe there is additional use cases where the condition is needed to be added too or do it more generic all over providers

### DIFF
--- a/waterbutler/providers/dropbox/provider.py
+++ b/waterbutler/providers/dropbox/provider.py
@@ -302,7 +302,9 @@ class DropboxProvider(provider.BaseProvider):
         :return: A dictionary of the metadata about the file just uploaded
         """
 
-        path_arg = {"path": path.full_path}
+        path_full_path = path.full_path
+        path_arg = {"path": path_full_path if path_full_path.startswith('/') else f'/{path_full_path}'}
+
         if conflict == 'replace':
             path_arg['mode'] = 'overwrite'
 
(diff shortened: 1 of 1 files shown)


waterbutler/providers/dropbox/provider.py
```diff
@@ -302,7 +305,9 @@ async def _contiguous_upload(self,
         :return: A dictionary of the metadata about the file just uploaded
         """
 
-        path_arg = {"path": path.full_path}
+        path_full_path = path.full_path
```
opaduchak, line 308: nit: rename `path_full_path` to just `full_path`
```suggestion
        full_path = path.full_path
```

waterbutler/providers/dropbox/provider.py
```diff
@@ -302,7 +305,9 @@ async def _contiguous_upload(self,
...
+        path_arg = {"path": path_full_path if path_full_path.startswith('/') else f'/{path_full_path}'}
```
opaduchak, line 309: > `if path and not path.startswith('/') and not path.startswith('rev:'):`

Don't we need to check for path starting with `'rev:'` here

opaduchak (changes requested):
Great work, I have one nit and one question

waterbutler/providers/dropbox/provider.py
mkovalua: thanks, let it be, the additional will not be redundant

PR #456, opened 2025-07-25
[ENG-8457] | Fix broken copy folder to osfstorage issue

sh-andriy:

### DIFF
--- a/waterbutler/core/metadata.py
+++ b/waterbutler/core/metadata.py
@@ -125,7 +125,7 @@ class BaseMetadata(metaclass=abc.ABCMeta):
         module = importlib.import_module(module_name)
         meta_cls = getattr(module, class_name)
 
-        args = cls._rehydrate(payload)
+        args = meta_cls._rehydrate(payload)
         return meta_cls(*args)  # type: ignore
 
     @classmethod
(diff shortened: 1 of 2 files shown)


waterbutler/core/metadata.py
```diff
@@ -418,11 +418,8 @@ def _dehydrate(self) -> dict:
     def rehydrate(cls, payload):
         built_obj = super().rehydrate(payload)
         if payload['_children'] is not None:
```
opaduchak, line 420:
```suggestion
        if payload.get('_children') is not None:
```

opaduchak (changes requested):
Looks good, one suggestion

waterbutler/core/metadata.py
sh-andriy: agree on that

waterbutler/core/metadata.py
```diff
@@ -320,13 +320,13 @@ def dehydrate(self) -> dict:
...
-        module_name = obj.__class__.__module__
-        class_name = obj.__class__.__name__
+        module_name = self.__class__.__module__
+        class_name = self.__class__.__name__
 
         payload: dict[str, object] = {
             "__wb_meta__": True,
             "cls": f"{module_name}.{class_name}",
-            "raw": obj.raw,
+            "raw": self.raw,
```
opaduchak, line 329: I don't like this being shown as a a diff as this is already in the `feature/buff-worms`, probably github issue, but check one more time, please

opaduchak (approved):
LGTM, but please check the diff one more time

PR #457, opened 2025-07-28
[ENG-8413] Fix double encoding spaces in furl

Vlad0n20:
## Ticket

https://openscience.atlassian.net/browse/ENG-8413?atlOrigin=eyJpIjoiZWFmZDI2MTdjYTIyNDcyZmJlZWNlOGQ2YzkzYzk2ZjEiLCJwIjoiaiJ9

### DIFF
--- a/waterbutler/core/provider.py
+++ b/waterbutler/core/provider.py
@@ -55,17 +55,9 @@ def throttle(concurrency=10, interval=1):
 
 def build_url(base, *segments, **query):
     url = furl.furl(base, args=query)
-    # Filters return generators
-    # Cast to list to force "spin" it
     url.path.segments = list(filter(
         lambda segment: segment,
-        map(
-            # Furl requires everything to be quoted or not, no mixtures allowed
-            # prequote everything so %signs don't break everything
-            lambda segment: parse.quote(segment.strip('/')),
-            # Include any segments of the original url, effectively list+list but returns a generator
-            itertools.chain(url.path.segments, segments)
-        )
+        itertools.chain(url.path.segments, [segment.strip('/') for segment in segments])
     ))
     return url.url
 


waterbutler/core/provider.py
```diff
@@ -55,17 +55,9 @@ async def wrapped(*args, **kwargs):
...
     url.path.segments = list(filter(
         lambda segment: segment,
-        map(
-            # Furl requires everything to be quoted or not, no mixtures allowed
-            # prequote everything so %signs don't break everything
-            lambda segment: parse.quote(segment.strip('/')),
-            # Include any segments of the original url, effectively list+list but returns a generator
-            itertools.chain(url.path.segments, segments)
-        )
+        itertools.chain(url.path.segments, [segment.strip('/') for segment in segments])
```
opaduchak, line 60: I'm afraid that this may break many providers, Let's merge it, and in case it breaks things we'll know what to revert

opaduchak (approved):

waterbutler/core/provider.py
cslzchen: I wonder if the segments we provides here has already been quoted so it got double encoded again.

Maybe there is some inconsistence in whether we encode segments ahead of time or in this `build_url`. This may be different for different providers so the change will likely break some.

cslzchen (commented):
As I commented, I am also afraid this may break some providers. I would prefer encoding to happen in one place, e.g. handled here by `build_url`. We just need to make sure the segments we provides here are not already encoded? Any thoughts @opaduchak and @felliott ? Or maybe the upgrade made it hard to do it this way so we have to enforce encoding for segments before calling `build_url`?

cslzchen (approved):

PR #461, opened 2025-08-05
[ENG-8500] Update tornado to 6.5.1

Vlad0n20:


pyproject.toml
```diff
@@ -32,9 +32,10 @@ sentry-sdk = "2.22.0"
 redis = "5.2.1"
 setuptools = "^80.1.0"
 stevedore = "5.4.1"
-tornado = "6.4.2"
+tornado = "^6.5.1"
```
cslzchen, line 35: I am not 100% sure. So a question for @felliott: do we ping `6.5.1` or we use this backward compatible `^` feature. I see some uses it while others don't.

cslzchen: I took a look at the MFR tornado upgrade [commit](https://github.com/CenterForOpenScience/modular-file-renderer/pull/398/commits/e35eb9c8df24f0b469d83bba0763dfeb6e78db7e#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R21), it doesn't use the `^`. So I think we should have some kind of consistency of when to use `^` and when not to.

cslzchen: Update: I changed this to `6.5.1` and used `poetry==2.1.2` to re-generate the lock file.

cslzchen (approved):

PR #474, opened 2026-03-23
ENG-10556 | Add OracleCloud (OCI) Object Storage provider for osfstorage

sh-andriy:


felliott (approved):
Looks good! I've created a `feature/oci-prototype` branch and repointed it at that. Will merge now.

PR #476, opened 2026-04-09
[ENG-10670] | oraclecloud - attempt an OCI native API-only interface

sh-andriy:


felliott (approved):
Awesome, thank you! I've create a new branch, `feature/oci-proto-oci-only`, and will merge into that.

PR #477, opened 2026-04-20
[ENG-10671] | oraclecloud - attempt an s3-compatible only interface

sh-andriy:


felliott (approved):
Thanks, @sh-andriy! This looks really good and makes me think that polling aside, the s3-compat api might be the way to go.

PR #490, opened 2026-07-03
[ENG-11516] Eliminate the request between metadata and download

antkryt:
## Ticket

https://openscience.atlassian.net/browse/ENG-11516

## Purpose

Pass storage meta to provider download method

### DIFF
--- a/waterbutler/providers/osfstorage/provider.py
+++ b/waterbutler/providers/osfstorage/provider.py
@@ -195,7 +195,7 @@ class OSFStorageProvider(provider.BaseProvider):
 
         return url, data, params
 
-    async def download(self, path, version=None, revision=None, mode=None, **kwargs):
+    async def download(self, path, version=None, revision=None, mode=None, metadata=None, **kwargs):
         if not path.identifier:
             raise exceptions.NotFoundError(str(path))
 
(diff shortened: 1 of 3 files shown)


cslzchen (approved):
⭐ 

Looks great and will merge after local verification.

waterbutler/providers/osfstorage/provider.py
```diff
@@ -209,28 +209,37 @@ async def download(self, path, version=None, revision=None, mode=None, **kwargs)
...
-            self.build_url(path.identifier, 'download', version=version, mode=mode),
-            expects=(200, ),
-            params=user_param,
-            throws=exceptions.DownloadError,
-        )
-        data = await resp.json()
+        if version is None and storage and storage.get('data', None):
+            data = storage
+        else:
+            # Capture user_id for analytics if user is logged in
```
cslzchen: Let's add a comment explaining that the first option is to bypass the extra request and only falls back to make that request if `storage` returned is `None`.

In addition, I don't know when this fall back happens, we should add a sentry message to log. Note: don't log any credentials. Just the node and the path.

Another thing I missed earlier is this request affects `osfstorage` download in general instead of just for DAZ. This means, we need to extra careful.

> `# Capture user_id for analytics if user is logged in`

Can you take a look at if our change will lead to missing analytics? What are the analytics this request is capturing? Is it captured on WB or more likely is it sending information to OSF for OSF to capture?

waterbutler/providers/osfstorage/provider.py
```diff
@@ -555,6 +564,8 @@ async def _children_metadata(self, path, limit=None, after=None, **kwargs):
             query['limit'] = limit
             if after is not None:
                 query['after'] = after
+        elif kwargs.get('orm') is not None:
+            query['orm'] = kwargs['orm']
```
cslzchen, line 598: This may be done in a future ticket but we need to decide and document the default behavior. As it is coded right now, the behavior is:

* RAW SQL is used unless `orm` is explicitly specified or `limit` (pagination, which requires `orm`) is used

Another thought I have is, if ORM is equivalent to raw SQL, is there a strong reason to keep both, since it adds maintenance effort to sync both?

cslzchen (changes requested):
Re-review this PR with local testing and raised two questions 🔥

waterbutler/providers/osfstorage/provider.py
antkryt: About metrics, we use this function when authorize request from wb/mfr:
```python
def _enqueue_metrics(file_version, file_node, action, auth, from_mfr=False):
    if not file_version:
        return

    if action == 'render':
        file_signals.file_viewed.send(auth=auth, fileversion=file_version, file_node=file_node)
    elif action == 'download' and not from_mfr:
        file_signals.file_downloaded.send(auth=auth, fileversion=file_version, file_node=file_node)
    return
```
For zip, file_version is always None, so this returns immediately. This means that DAZ did not increment per-file download counts before my changes anyway
osfstorage_download itself also does not update analytics, it only returns {data, settings}

waterbutler/providers/osfstorage/provider.py
```diff
@@ -209,28 +211,51 @@ async def download(self, path, version=None, revision=None, mode=None, **kwargs)
...
+                    })
+                    sentry_sdk.capture_message(
+                        'osfstorage download fell back to OSF /download endpoint',
+                        level='info',
+                    )
+
+            # Capture user_id for analytics if user is logged in
+            user_param = {}
+            if self.auth.get('id', None):
+                user_param = {'user': self.auth['id']}
```
cslzchen: Let's remove this from the request since we confirmed this is not used in the OSF endpoint/view `osfstorage_download`.

waterbutler/providers/osfstorage/provider.py
```diff
@@ -213,9 +215,23 @@ async def download(self, path, version=None, revision=None, mode=None, metadata=
...
+            if metadata is not None:
+                with sentry_sdk.push_scope() as scope:
+                    scope.set_context('osfstorage_data', {
+                        'node': self.nid,
+                        'path': str(path),
+                    })
+                    sentry_sdk.capture_message(
+                        'osfstorage download fell back to OSF /download endpoint',
+                        level='info',
+                    )
```
cslzchen: Move this to the first time we check `metadata` and set `storage`

cslzchen (approved):
👍

waterbutler/providers/osfstorage/provider.py
```diff
@@ -209,28 +211,46 @@ async def download(self, path, version=None, revision=None, mode=None, **kwargs)
...
-        resp = await self.make_signed_request(
-            'GET',
-            self.build_url(path.identifier, 'download', version=version, mode=mode),
-            expects=(200, ),
-            params=user_param,
-            throws=exceptions.DownloadError,
-        )
-        data = await resp.json()
+        storage = None
+        use_embedded_storage = False
```
cslzchen, line 215: 👍

cslzchen (approved):

PR #493, opened 2026-07-14
[ENG-11581] Update WB to use the best option by default

antkryt:
## Ticket

https://openscience.atlassian.net/browse/ENG-11581

## Purpose

Remove orm query param for daz metadata


cslzchen (approved):
Looks good and will merge after I check with QA on current testing cycle 🔥

BE part: https://github.com/CenterForOpenScience/osf.io/pull/11809

PR #494, opened 2026-07-16
[ENG-11736] pass downloaded bytes, IP and completed flag to callback

ihorsokhanexoft:
https://openscience.atlassian.net/browse/ENG-11736

In the payload the following properties appear now:
image

`download_file`, `download_zip` actions are marked as completed only if we get 200 or 302 from a provider

### DIFF
--- a/tests/core/test_remote_logging.py
+++ b/tests/core/test_remote_logging.py
@@ -3,6 +3,77 @@ import pytest
 from waterbutler.core import remote_logging
 
 
+class TestLogToCallback:
+
+    @pytest.mark.asyncio
+    @pytest.mark.parametrize('completed, expected', [(False, False), (True, True)])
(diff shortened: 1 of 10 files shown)


sh-andriy (changes requested):
Looks good overall thanks.
Just one thing that i noticed:
Saw your note - `completed` on `200`/`302` from the provider catches provider-side failures, but not the client cancelling mid-download. WB's already sent `200` once it starts streaming, and `write_stream` swallows the disconnect, so a user who aborts a big zip still lands as `completed=True` with only partial bytes. The client specifically wants to tell finished vs aborted apart, so we'd need the `StreamClosedError` signal out of `write_stream`, not just the response status.

sh-andriy (approved):
lgtm

